PR #4, opened 2021-02-20
Bugfix/foreign key fixes

la4de:
Fixes and improvements as discussed in https://github.com/la4de/strawberry-graphql-django/commit/1a47d26b140549f8951fb74f22052906c610a2a2

### DIFF
--- a/tests/app/models.py
+++ b/tests/app/models.py
@@ -19,7 +19,8 @@ class DataModel(models.Model):
     nullable = models.IntegerField(null=True)
     hasdefault = models.IntegerField(default=1)
     relation = models.ManyToManyField(User)
-    foreign_key = models.ForeignKey(User, null=True, on_delete=models.SET_NULL)
+    mandatory_foreign_key = models.ForeignKey(User, on_delete=models.SET_NULL)
+    optional_foreign_key = models.ForeignKey(User, null=True, on_delete=models.SET_NULL)
 
 class UnknownField(models.Field):
     pass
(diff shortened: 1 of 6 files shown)


AGASS007:
🎉 

definitely the way to go IMO. having `mandatoryForeignKeyId` and `mandatoryForeignKey` is definitely clearer than having one `mandatoryForeignKey` that has two different meanings depending on the context (either an `object` in a query context, or and `ID` in a mutation context).

I had the same interrogations with REST APIs and nested objects in a read context vs setting an id in a write context.

tests/app/models.py
```diff
@@ -19,7 +19,8 @@ class DataModel(models.Model):
     nullable = models.IntegerField(null=True)
     hasdefault = models.IntegerField(default=1)
     relation = models.ManyToManyField(User)
-    foreign_key = models.ForeignKey(User, null=True, on_delete=models.SET_NULL)
+    mandatory_foreign_key = models.ForeignKey(User, on_delete=models.SET_NULL)
```
AGASS007, line 22: not that it matters too much in a test context, but having `on_delete=models.SET_NULL` on a non nullable FK feels weird. Maybe set `models.CASCADE` or `models. PROTECT`

la4de:
> 🎉
> 
> definitely the way to go IMO. having `mandatoryForeignKeyId` and `mandatoryForeignKey` is definitely clearer than having one `mandatoryForeignKey` that has two different meanings depending on the context (either an `object` in a query context, or and `ID` in a mutation context).
> 
> I had the same interrogations with REST APIs and nested objects in a read context vs setting an id in a write context.

I agree. Thanks for your feedback!

la4de:
Changes pushed to main branch

PR #5, opened 2021-02-20
Batch create mutation

star2000:
can do
```graphql
mutation {
  createItems(data: [{name: "item1"}, {name: "item2"}]) {
    id
  }
}
```
Keep the `createItem` at the same time.

### DIFF
--- a/strawberry_django/resolvers.py
+++ b/strawberry_django/resolvers.py
@@ -157,7 +157,7 @@ class ModelQueryMixin:
     def mutation(cls):
         object_name = utils.camel_to_snake(cls.model._meta.object_name)
         class Mutation: pass
-        setattr(Mutation, f'create_{object_name}', cls.create_mutation())
+        setattr(Mutation, f'create_{object_name}s', cls.create_mutation())
         setattr(Mutation, f'update_{object_name}s', cls.update_mutation())
         setattr(Mutation, f'delete_{object_name}s', cls.delete_mutation())
         return strawberry.type(Mutation)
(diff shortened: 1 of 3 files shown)


strawberry_django/resolvers.py
```diff
@@ -48,9 +48,10 @@ def get(self, id):
 
     def create(self, data):
         model = self.get_model()
-        instance = model(**data)
-        instance.save()
-        return instance
+        instances = [model(**i) for i in data]
+        for instance in instances:
+            instance.save()
```
la4de, line 53: I'd prefer bulk creation. It's more efficient. What do you think?

https://docs.djangoproject.com/en/3.1/ref/models/querysets/#bulk-create

la4de (changes requested):
First of all I want to thank you for your contribution.

I'd prefer the name `createItems`.

Please add unit test for batch creation. You can run unit tests by executing following commands under the repo
```
poetry install
poetry run pytest
```
Prerequisite is that you have `poetry` installed and available.

strawberry_django/resolvers.py
star2000: Tried, this method can't return the ID field

star2000:
> First of all I want to thank you for your contribution.
> 
> I'd prefer the name `createItems`.
> 
> Please add unit test for batch creation. You can run unit tests by executing following commands under the repo
> 
> ```
> poetry install
> poetry run pytest
> ```
> 
> Prerequisite is that you have `poetry` installed and available.

complete

star2000:
@la4de

la4de:
Looks good to me. Thank you very much!

PR #9, opened 2021-03-17
Add examples to readme and example repo

joeydebreuk:
Add 
```
    @strawberry.field
    def something(self, info, root) -> List[Resolver.output_type]:
```
as discussed in https://github.com/strawberry-graphql/strawberry-graphql-django/issues/8

And add some useful django auth examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,15 +30,10 @@ class Group(models.Model):
 
 schema.py:
 ```python
+from typing import List
 import strawberry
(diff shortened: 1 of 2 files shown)


la4de:
Changes looks good to me.

btw, I'm working on new API which is more type oriented like existing one. therefore I'm not planning to continue development of old API. please take a look at next branch and feel free to give any comments

PR #15, opened 2021-03-28
Implement django-filters for filter generation

joeydebreuk:
Some things to consider:

- Should we make "django-filters" an optional dependency?
- How do we implement this into the library? I suppose it should be optional, and easy to modify.
We could add global settings eg:
```
STRAWBERRY_DJANGO_GENERATE_FILTERS = "default to false, when set true we raise error when its not installed"
STRAWBERRY_DJANGO_FILTER_FACTORY = "defaults_to_django-filters_factory_fn"
```
And also allow for overwriting it in specific queries.

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -141,7 +141,7 @@ def filter(filterset_class: Type[django_filters.FilterSet], name=None):
     name = name or filterset_class.__name__
     cls = type(name, (), {"__annotations__": {}, "filterset_class": filterset_class})
 
-    for field_name in filterset_class.get_fields():
+    for field_name in filterset_class.get_filters().keys():
         filter_field = filters[field_name]
         field_type = get_filter_field_type(type(filter_field))
         cls.__annotations__[field_name] = field_type
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -6,11 +6,11 @@ from strawberry.arguments import UNSET, is_unset
 
 
 __all__ = [
-    "FilterSetError",
-    "apply_filter",
+    "InvalidFilterError",
+    "apply",
     "filter",
-    "get_filter_field_type",
-    "set_filter_field_type",
+    "get_field_type",
+    "set_field_type",
 ]
 
 
@@ -82,12 +82,12 @@ filter_field_type_map = {
 }
 
 
-class FilterSetError(GraphQLError):
+class InvalidFilterError(GraphQLError):
     pass
 
 
 @assert_django_filters_installed
-def get_filter_field_type(field_type):
+def get_field_type(field_type):
     if type(field_type) != type:
         raise TypeError(
             f"expected 'type', received {type(field_type)}. Maybe you forgot to call type()?"
@@ -99,7 +99,7 @@ def get_filter_field_type(field_type):
 
 
 @assert_django_filters_installed
-def set_filter_field_type(field_type, to_type):
+def set_field_type(field_type, to_type):
     if type(field_type) != type:
         raise TypeError(
             f"expected 'type', received {type(field_type)}. Maybe you forgot to call type()?"
@@ -108,7 +108,7 @@ def set_filter_field_type(field_type, to_type):
 
 
 @assert_django_filters_installed
-def apply_filter(filter_instance, queryset):
+def apply(filter_instance, queryset):
     if is_unset(filter_instance) or not filter_instance:
         return queryset
 
@@ -124,7 +124,7 @@ def apply_filter(filter_instance, queryset):
     )
 
     if not filterset.is_valid():
-        raise FilterSetError(filterset.errors)
+        raise InvalidFilterError(filterset.errors)
 
     return filterset.qs
 
@@ -143,7 +143,7 @@ def filter(filterset_class: Type[django_filters.FilterSet], name=None):
 
     for field_name in filterset_class.get_filters().keys():
         filter_field = filters[field_name]
-        field_type = get_filter_field_type(type(filter_field))
+        field_type = get_field_type(type(filter_field))
         cls.__annotations__[field_name] = field_type
         setattr(cls, field_name, UNSET)
 
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -13,22 +13,26 @@ __all__ = [
     "set_filter_field_type",
 ]
 
+
+class DummyDjangoFilters:
+    def __getattribute__(self, attr):
+        # Dont raise KeyError for not existing attr
+        try:
+            return super(DummyDjangoFilters, self).__getattribute__(attr)
+        except KeyError:
+            return attr
+
+
 try:
     import django_filters
 except ModuleNotFoundError:
-    class DummyDjangoFilters:
-        def __getattribute__(self, attr):
-            return attr
-
     django_filters = DummyDjangoFilters()
 
 
 def assert_django_filters_installed(fn):
     @functools.wraps(fn)
     def wrapper(*args, **kwargs):
-        try:
-            import django_filters
-        except ModuleNotFoundError:
+        if isinstance(django_filters, DummyDjangoFilters):
             raise ModuleNotFoundError(
                 'You need to install django-filter to use "strawberry_django.filter". '
                 "See https://django-filter.readthedocs.io/"
@@ -127,6 +131,12 @@ def apply_filter(filter_instance, queryset):
 
 @assert_django_filters_installed
 def filter(filterset_class: Type[django_filters.FilterSet], name=None):
+    if not isinstance(filterset_class, django_filters.filterset.FilterSetMetaclass):
+        raise TypeError(
+            "strawberry_django.filter expects a class that inherits django_filters.FilterSet, received %s",
+            type(filterset_class),
+        )
+
     filters = filterset_class.get_filters()
     name = name or filterset_class.__name__
     cls = type(name, (), {"__annotations__": {}, "filterset_class": filterset_class})
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,0 +1,140 @@
+import functools
+from typing import Type, Optional, List
+import strawberry
+from graphql import GraphQLError
+from strawberry.arguments import UNSET, is_unset
+
+
+__all__ = [
+    "FilterSetError",
+    "apply_filter",
+    "filter",
+    "get_filter_field_type",
+    "set_filter_field_type",
+]
+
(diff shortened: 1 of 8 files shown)


la4de:
Thank you @joeydebreuk, this looks very good. I agree that filtering should be optional, configurable and flexible. I like the idea of using django-filters library. I'd consider about having it as an optional dependency so that we do not have hard dependencies to anything else than core django.

I'm thinking about the API for filtering. I draft something below. What do you think?
```python
# user defined filter set
class UserFilter(django_filters.FilterSet):
        name = django_filters.CharFilter(lookup_expr="icontains")
        search = django_filters.CharFilter(method="filter_search")

# define filter object (this is optional, see option 1)
user_filter = strawberry_django.filters(UserFilter)

# add filters to generated query field
@strawberry.type
class Query:
     # option 1: filter set as parameter
    users = strawberry_django.queries.list(models.User, filters=UserFilter)

    # option 2: filter object as parameter
    users = strawberry_django.queries.list(models.User, filters=user_filter)

    # option 3: filter object as decorator
    @user_filter
    users = strawberry_django.queries.list(models.User)

    # option 4: own resolver function
    def users(root, info, filters: user_filter.type):
        qs = models.User.objects.all()
        return user_filter.filter(filters, qs)
        # or simply
        return user_filter.filter(filters)
```

Filters generation

```python

# generate all lookups for all fields
generated_filters = strawberry_django.filters(models.Group)

# generate all lookups for selected fields
generated_filters = strawberry_django.filters(
    models.Group,
    fields=['id', 'name'],
)

# generate selected lookups for selected fields
generated_filters = strawberry_django.filters(
    models.Group,
    fields={
        'id': ['in', 'lte', 'gt'],
        'name': ['iexact', 'icontains'],
    },
)

@strawberry_django.type(models.User):
class User:
    # option 1: field parameter
    groups: List[Group] = strawberry_django.field(filters=generated_filters)

    # option 2: filter object as decorator
    @generated_filters
    groups: List[Group] = strawberry_django.field()
    @strawberry_django.filters(models.Group, fields=['id'])
    groups: List[Group] = strawberry_django.field()

    # option 3: field has `filters` registration function
    groups: List[Group] = strawberry_django.field()
    groups.filters(generated_filters)
    
    # option 4: field has `filters` registration function which could be used as decorator
    groups: List[Group] = strawberry_django.field()
    @groups.filters
    class
```

joeydebreuk:
I think we have to consider the pros / cons from creating an API that is coupled with django-filters
```
   # stuff like 
    users = strawberry_django.queries.list(models.User, filters=UserFilter)
    # and 
    generated_filters = strawberry_django.filters(
    models.Group,
    fields={
        'id': ['in', 'lte', 'gt'],
        'name': ['iexact', 'icontains'],
        },
    )
```
This might make it difficult to ever move away from django-filters. Also, IMO declaring FilterSet a filterset is already very straightforward. No need to create shortcuts. Better to have 1 way of doing it. Maybe just stick to:
```
@strawberry_django.filter
class GroupFilter(django_filters.FilterSet):
    class Meta:
        model = Group
        fields = {
            "id": ["in", "lte", "gt"],
            "name": ["iexact", "icontains"],
        }

@strawberry.type
class Query:
    groups = strawberry_django.queries.list(models.Group, filters=GroupFilter)
```

I do think that this makes sense:
```
    # option 3: own resolver function
    def users(root, info, filters: user_filter.type) -> List[User]:
        qs = models.User.objects.all()
        return user_filter.filter(filters, qs)
        # or simply
        return user_filter.filter(filters, models.User)
        # or even this (if model has been defined in meta)
        return user_filter.filter(filters)
```
But maybe even here, we could just always require a qs. `models.User` instead of `models.User.objects.all()`. Or at least for now, to keep it straight forward.

Sidenote about your example, I think this only generates exact lookup:
```
generated_filters = strawberry_django.filters(
    models.Group,
    fields=['id', 'name'],
)
```

la4de:
Yes, Let's keep things simple and start with approach which is straight forward, easy understand, use and implement. Maybe we could just create new `filters.py` under `strawberry_django` package and put code there.

`strawberry_django.filters()` could raise an exception if `django_filters` is not available.

joeydebreuk:
Sounds good, I'll start.

strawberry_django/filters.py
```diff
@@ -0,0 +1,121 @@
...
+]
+
+
+try:
+    import django_filters
+except ModuleNotFoundError as e:
+    raise ModuleNotFoundError(
+        'You need to install django-filters to use "strawberry_django.filters". '
+        "See https://django-filter.readthedocs.io/"
+    )
```
la4de, line 22: This exception breaks whole `strawberry_django` package if `django_filters`is not installed. Would following work?
```python
try:
    import django_filters
except ModuleNotFoundError:
    django_filters = None

def check_django_filters_package():
    if django_filters is None:
        raise ModuleNotFoundError(
            'You need to install django-filters to use "strawberry_django.filters". '
            "See https://django-filter.readthedocs.io/"
        )

def create_filter():
    check_django_filters_package()
    ...
```

tests/filters/test_filter_fields.py
```diff
@@ -0,0 +1,72 @@
...
+
+@pytest.mark.django_db
+def test_filter_map_types():
+    qs = FilterModel.objects.all()
+
+    @strawberry_django.create_filter
+    class Filter(django_filters.FilterSet):
+        class Meta:
+            model = FilterModel
+            exclude = ()
```
la4de, line 34: Nice that `django_filters` can autogenerate filters.

I'm still thinking that it would be convenient to have shortcut to autogenerate filters especially at the beginning of project when you start prototyping new API. Need to think about that.

tests/filters/test_filter_query.py
```diff
@@ -0,0 +1,92 @@
...
+    assert result.errors == None
+
+    # Execute with name filter
+    variable_values = {"filters": {"name": "user1"}}
+    result = schema.execute_sync(query, variable_values=variable_values)
+    assert result.data
+    assert result.data.get("userIds", None) == [1]
+    assert result.errors == None
+
+    # Execute with search filter
```
la4de, line 81: could we split these into smaller test cases so that one test would test one thing? pytest fixtures would help here

la4de (commented):
@joeydebreuk, thanks and good job!

I started to review the pull request and in general the functionality looks good. I'd still like to continue discussion about the API. I need to have a break now but I'll continue soon.

strawberry_django/filters.py
```diff
@@ -0,0 +1,121 @@
...
+        queryset=queryset,
+    )
+
+    if not filterset.is_valid():
+        raise FilterSetError(filterset.errors)
+
+    return filterset.qs
+
+
+def create_filter(filterset_class: Type[django_filters.FilterSet], name=None):
```
joeydebreuk, line 110: Change to `filter` because  `django_strawberry.filter` is more similar to `django_strawberry.type`

strawberry_django/filters.py
joeydebreuk: Yeah good point

tests/filters/test_filter_fields.py
joeydebreuk: Yeah, we can get some setting from django.config.settings. But thats the next step I would say.

la4de:
I drafted some ideas for filtering API in #20. We can do that later once the feature is ready.

strawberry_django/filters.py
la4de: Nice. I'm getting very useful error when trying to use strawberry_django.filter without having django-filters installed!

Do you have ideas how we could unit test that?

tests/filters/test_filter_fields.py
la4de: Let's do that later.

tests/filters/test_filter_query.py
la4de: Thanks for this test improvevent

strawberry_django/filters.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def assert_django_filters_installed(fn):
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
+        try:
+            import django_filters
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                'You need to install django-filters to use "strawberry_django.filters". '
```
la4de, line 33: I just noticed that the name of this package in pypi is `django-filter`, not `django-filters`. Should we use `django-filter` name instead?

strawberry_django/filters.py
```diff
@@ -0,0 +1,140 @@
...
+    import django_filters
+except ModuleNotFoundError:
+    class DummyDjangoFilters:
+        def __getattribute__(self, attr):
+            return attr
+
+    django_filters = DummyDjangoFilters()
+
+
+def assert_django_filters_installed(fn):
```
la4de, line 26: Should we add unit tests for this too?

strawberry_django/filters.py
la4de: I opened another discussion for unit testing. This has been resolved. Thanks!

strawberry_django/filters.py
joeydebreuk: Yeah good idea

strawberry_django/filters.py
joeydebreuk: I wouldnt know how...

la4de (changes requested):
We need to think about how to enable support for lookups.

```python
@strawberry_django.filter
class Filter(FilterSet):
    class Meta:
        model = Model
        fields = { 'username': ['iexact'] }
```

Currently I'm getting following error when trying to configure lookups.
```
File "strawberry-graphql-django/filters/strawberry_django/filters.py", line 135, in filter
    filter_field = filters[field_name]
KeyError: 'username'
```

strawberry_django/filters.py
la4de: done! thanks

strawberry_django/filters.py
```diff
@@ -0,0 +1,140 @@
...
+        raise FilterSetError(filterset.errors)
+
+    return filterset.qs
+
+
+@assert_django_filters_installed
+def filter(filterset_class: Type[django_filters.FilterSet], name=None):
+    filters = filterset_class.get_filters()
+    name = name or filterset_class.__name__
+    cls = type(name, (), {"__annotations__": {}, "filterset_class": filterset_class})
```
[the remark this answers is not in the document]
joeydebreuk: The way I understand this convention, is that class private variables should be used within class only, which is not the case here.

PR #19, opened 2021-04-06
Deprecate is_update parameter

la4de:
We are planning to deprecate `is_update` parameter and introduce new `partial` parameter of `strawberry_django.input`. `is_update` will be removed in v0.2 version.

### DIFF
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -1,13 +1,15 @@
 import strawberry
 import typing
 from .types import get_model_fields, update_fields
+from .utils import deprecated
 
 _type = type
 
-def type(model, *, fields=None, types=None, is_update=False, **kwargs):
+def type(model, *, fields=None, types=None, **kwargs):
     def wrapper(cls):
         is_input = kwargs.get('is_input', False)
-        model_fields = get_model_fields(cls, model, fields, types, is_input, is_update)
+        partial = kwargs.pop('partial', False)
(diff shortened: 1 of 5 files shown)


strawberry_django/type.py
```diff
@@ -1,13 +1,15 @@
...
+from .utils import deprecated
 
 _type = type
 
-def type(model, *, fields=None, types=None, is_update=False, **kwargs):
+def type(model, *, fields=None, types=None, **kwargs):
     def wrapper(cls):
         is_input = kwargs.get('is_input', False)
-        model_fields = get_model_fields(cls, model, fields, types, is_input, is_update)
+        is_partial = kwargs.pop('is_partial', False)
```
patrick91, line 11:
```suggestion
        is_partial = kwargs.pop('partial', False)
```

what about just `partial`? :)

la4de: that's even better :). why do we have is_input, instead if input? I just followed the same format herem. is that to avoid collision with internals?

la4de: Renamed

patrick91: good question! yes to prevent clashing with internals, but also `is_input` is not really exposed to end users (and I think it might go away in future) :)

patrick91 (approved):

la4de:
I'll close this now and push commits manually to main branch. Thanks for the feedback :)

Issue #20, opened 2021-04-08
Proposal for class-oriented API


la4de:
Would it make sense to define all fields always in class? I had also few other improvement ideas for authentication and filters. We have functional implementation in `feature/class-api` branch which is still under development.

First draft of doc is available, please take a look and leave your comments. https://github.com/strawberry-graphql/strawberry-graphql-django/blob/feature/class-api/docs/index.md

Changelog:
* 9.4.2021 - initial draft
* 10.4.2021 - API adjustments
* 4.5.2021 - Add couple of new examples and update existing ones
* 12.5.2021 - Add link to docs and update filtering api

types
```python
import strawberry_django
from strawberry_django import auto

@strawberry_django.type(models.User)
class User:
    id: auto
    name: auto
    car: List['Car'] = strawberry_django.field(field_name='car_set')

@strawberry_django.type(models.Car)
class Car:
    id: auto
    model: auto
    owner: User

@strawberry_django.input(models.User)
class UserInput:
    name: auto

# type inheritance
@strawberry_django.input(models.User, partial=True)
class UserPartialInput(UserInput):
    pass
```

queries and mutations
```python
from strawberry_django import mutations

@strawberry.type
class Query:
    user: User = strawberry_django.field()
    users: List[User] = strawberry_django.field(filters=UserFilter)
    cars: List[Car] = strawberry_django.field(filters=CarFilter, pagination=True)

@strawberry.type
class Mutation:
    create_user: User = mutations.create(UserInput)
    update_user: User = mutations.update(UserPartialInput)
    update_users: List[User] = mutations.update(UserPartialInput, filters=UserFilter, many=True)
    delete_users: List[User] = mutations.delete(filters=UserFilter, many=True)
```

authentication
```python
from strawberry_django import auth

@strawberry.type
class Query:
    me: User = auth.current_user()

@strawberry.type
class Mutation:
    login: User = auth.login()
    logout = auth.logout()
```

filters
```python
@strawberry_django.filters.filter(models.Car, lookups=True)
class CarFilter:
    id: auto
    model : auto
    search: str

    # question: how should we define the filter function for special fields?
    def filter_search(self, queryset):
        return queryset.filters(name__contains=self.search)
```

patrick91:
I don't dislike this, my only concert is that might be a bit more confusing than a `fields` argument (as you'd do on django rest framework and django forms)

the other thing, should consider a way of automatically exposing all the fields? 🤔 

[...]

PR #30, opened 2021-05-18
Class Oriented API

la4de:
This pull request adds new class oriented API. See more information about the new API from:
* Ticket #20
* https://github.com/strawberry-graphql/strawberry-graphql-django/blob/feature/class-api/README.md
* https://github.com/strawberry-graphql/strawberry-graphql-django/blob/feature/class-api/docs/index.md


README.md
```diff
@@ -4,247 +4,108 @@
 [![PyPI](https://img.shields.io/pypi/v/strawberry-graphql-django)](https://pypi.org/project/strawberry-graphql-django/)
 [![PyPI - Downloads](https://img.shields.io/pypi/dm/strawberry-graphql-django)](https://pypi.org/project/strawberry-graphql-django/)
 
-This library provides helpers to generate types, mutations and resolvers for Django models.
+This package provides simple and powerful tools to generate GraphQL types, quries, mutations and resolvers from Django models.
```
joeydebreuk, line 7: queries

README.md
```diff
@@ -4,247 +4,108 @@
...
 
-Installing strawberry-graphql-django packet from the python package repository.
+Installing `strawberry-graphql-django` packet from the python package repository.
```
joeydebreuk, line 9: packet should be package

docs/guides/quick-start.md
```diff
@@ -0,0 +1,147 @@
...
+    color {
+      name
+    }
+  }
+}
+# fruits: [
+#   { name: "strawberry", color: { name: "red" } },
+#   { name: "raspberry", color: { name: "red" } }
+# ]
+```
```
joeydebreuk, line 147: Would be useful to add mutation here? Or you are leaving that out because API is not definitive yet?

strawberry_django/auth/mutations.py
```diff
@@ -0,0 +1,26 @@
...
+def login():
+    @strawberry.mutation
+    @django_resolver
+    def login(info, username: str, password: str):
+        request = info.context.request
+        user = auth.authenticate(request, username=username, password=password)
+        if user is not None:
+            auth.login(request, user)
+            return user
+        auth.logout(request)
```
joeydebreuk, line 14: Why do you wrap the login function in a login function?

Shouldnt this login fn have a return type?

strawberry_django/auth/mutations.py
```diff
@@ -0,0 +1,26 @@
...
+        return request.user
+    return login
+
+def logout():
```
joeydebreuk, line 18: I don't see how string return type makes sense here, although I do not have much better suggestions, but maybe a bool, and `return True`?

strawberry_django/fields/field.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class StrawberryDjangoField(
+        StrawberryDjangoFieldOrdering,
+        StrawberryDjangoFieldFilters,
+        StrawberryDjangoPagination,
+        StrawberryDjangoFieldBase,
+        StrawberryField):
+
+    def __init__(self, django_name=None, graphql_name=None, python_name=None, **kwargs):
```
joeydebreuk, line 33: Comments explaining the kwargs would be useful, also for the class properties.

strawberry_django/mutations/mutations.py
```diff
@@ -0,0 +1,27 @@
...
+        args = (input_type,) + args
+        return mutations_legacy.update(*args, types=types)
+    return DjangoUpdateMutation(input_type, filters=filters)
+
+def delete(*args, types=None, filters=UNSET):
+    if args or types:
+        return mutations_legacy.delete(*args, types=types)
+    return DjangoDeleteMutation(input_type=None, filters=filters)
+
+mutations.create = create
```
joeydebreuk, line 25: This confuses me, and also my IDE. Isn't here a better way?

strawberry_django/fields/field.py
joeydebreuk: And docstring to explain the purpose of this class

strawberry_django/fields/field.py
```diff
@@ -0,0 +1,105 @@
...
+            django_name=getattr(field, 'django_name', field.name),
+            graphql_name=getattr(field, 'graphql_name', None),
+            python_name=field.name,
+            type_=field.type,
+        )
+        new_field.is_auto = getattr(field, 'is_auto', False)
+        new_field.origin_django_type = getattr(field, 'origin_django_type', None)
+        return new_field
+
+    def get_result(self, source, info, kwargs):
```
joeydebreuk, line 60: Shouldnt this be `**kwargs`? If not, find a different name.

strawberry_django/fields/types.py
```diff
@@ -0,0 +1,159 @@
+import datetime, decimal, uuid
+import django
+import strawberry
+from django.db.models import fields
+from django.db.models.fields.reverse_related import ForeignObjectRel, ManyToOneRel
+from strawberry.arguments import UNSET
+from typing import List, Optional
+from .. import filters
+
+class auto:
```
joeydebreuk, line 10: I like lowercase because it fits nicely with `str`, etc. But we could also consider using `AUTO` to be consistent with `UNSET`, just a thought.

strawberry_django/fields/types.py
```diff
@@ -0,0 +1,159 @@
...
+    set: Optional[List[strawberry.ID]] = UNSET
+
+@strawberry.input
+class ManyToManyInput:
+    add: Optional[List[strawberry.ID]] = UNSET
+    remove: Optional[List[strawberry.ID]] = UNSET
+    set: Optional[List[strawberry.ID]] = UNSET
+
+field_type_map = {
+    fields.AutoField: strawberry.ID,
```
joeydebreuk, line 50: I think its important we can extend this. Maybe something like:
```py
field_type_map = {}

def register_field_type...

@strawberry.input
@register_field_type
class SomeClass
```
But we could add this later I guess

strawberry_django/filters.py
```diff
@@ -1,147 +1,165 @@
...
+        return filter_method(queryset)
+
+    filter_kwargs, filter_methods = build_filter_kwargs(filters)
+    queryset = queryset.filter(**filter_kwargs)
+    for filter_method in filter_methods:
+        queryset = filter_method(queryset=queryset)
+    return queryset
+
+
+class StrawberryDjangoFieldFilters:
```
joeydebreuk, line 135: Maybe we should have a base type for all these mixins

strawberry_django/mutations/fields.py
```diff
@@ -1,21 +1,159 @@
...
+
+class DjangoUpdateMutation(
+        DjangoMutationBase,
+        StrawberryDjangoFieldFilters,
+        StrawberryDjangoFieldBase,
+        StrawberryField):
+
+    @transaction.atomic
+    def resolver(self, info, source, data, **kwargs):
+        print('UPDATE', data)
```
joeydebreuk, line 72: You forgot a print here :laughing:

strawberry_django/auth/queries.py
```diff
@@ -0,0 +1,7 @@
+import strawberry_django
+
+def current_user():
+    @strawberry_django.field
```
joeydebreuk, line 4: Shouldn't this use `django_resolver`?

strawberry_django/type.py
```diff
@@ -1,45 +1,147 @@
...
 _type = type
 
-def type(model, *, fields=None, types=None, **kwargs):
+def get_type_attr(type_, field_name):
+    attr = getattr(type_, field_name, UNSET)
+    if utils.is_unset(attr):
+        attr = getattr(type_, '__dataclass_fields__', {}).get(field_name, UNSET)
+    return attr
+
+def get_field(django_type, field_name, field_annotation=None):
```
joeydebreuk, line 23: Explaining the blocks in this function would be helpful

joeydebreuk (commented):
Nice work! I do think it would be good to use [black](https://github.com/psf/black)

README.md
2malh: Typo: queries

la4de: done

README.md
la4de: done

strawberry_django/mutations/fields.py
la4de: thanks. this definitely should not be there :)

strawberry_django/type.py
la4de: I agree. This is one of the most magical function. I added few comments here.

strawberry_django/auth/queries.py
la4de: The idea is that `current_user` function returns new field which resolves current user.  Following code does the same thing. Would it be better?

```python
def resolve_current_user(info):
    return info.context.request.user

def current_user():
    return strawberry_django.field(resolver=resolve_current_user)
```

strawberry_django/fields/field.py
la4de: This is the name strawberry core uses. I wanted to use same naming here for consistency. I already proposed to use `* *kwargs` syntax for that.

See https://github.com/strawberry-graphql/strawberry/issues/920#issuecomment-842568355

strawberry_django/auth/mutations.py
la4de: I have the same feeling. Boolean is better. Changed

strawberry_django/auth/mutations.py
la4de: We want to return mutation field, not function. Would following code be better?

```python
@django_resolver
def resolve_login(info, username: str, password: str):
    ...
    return request.user

@django_resolver
def resolve_logout(info) -> bool:
    ...
    return True

def login():
    return strawberry.mutation(resolver=resolve_login)

def logout():
    return strawberry.mutation(resolver=resolve_logout)
```

Return type of login should come from app, in that way developer can define own type for user like this

```python
@strawberry.type
class Mutation:
    login: MyUserType: strawberry_django.auth.login()
```

strawberry_django/fields/field.py
la4de: Documentation added

strawberry_django/filters.py
la4de: We have `StrawberryDjangoFieldBase` class in https://github.com/strawberry-graphql/strawberry-graphql-django/blob/feature/class-api/strawberry_django/fields/field.py#L12-L18.

Probably these should be inherited from that because they are using functions from there.

strawberry_django/fields/types.py
la4de: Users can already override all these types by not using `auto` type. Only fields annotated with `auto` are using these maps.

One may want to override also scalar types like this `register_field_type(fields.AutoField, UUID)`. Would it makes sense to add this into settings?

strawberry_django/fields/types.py
la4de: Good question. I like lower case type too. We have also discussed if `auto` is good name or if we should use some other name instead. One idea was to use `infer` instead of `auto`.

docs/guides/quick-start.md
la4de: > Nice work!

Thanks :) and thanks for your feedback. Really appreciate it.


>I do think it would be good to use [black](https://github.com/psf/black)

That's probably good idea. I would leave that out from this massive pull request.

la4de:
> Nice work!

Thanks :) and thanks for your feedback. Really appreciate it.


>I do think it would be good to use [black](https://github.com/psf/black)

That's probably good idea. I would leave that out from this massive pull request.

docs/guides/quick-start.md
la4de: There is separate page for mutations but it's good idea to say something about mutations in quick start as well. Documentation is one big thing we really need to focus on soon.

strawberry_django/fields/types.py
joeydebreuk: I wouldn't want to clog my settings with this. Especially because you might want to create custom scalars here also.

strawberry_django/auth/queries.py
joeydebreuk: Cant you do `current_user = strawberry_django.field(resolver=resolve_current_user)` and use that in schema?

strawberry_django/auth/mutations.py
joeydebreuk: Wouldn't a decorator work here?
```py
@strawberry.mutation
@django_resolver
def login(info, username: str, password: str):
    ...
    return request.user
    
# in schema:
      login = login
```

joeydebreuk:
Responded to your comments, nothing mayor. Can release IMO.

strawberry_django/auth/mutations.py
la4de: Yes that would work but then the syntax is a bit different compared to field. Strawberry field and mutation are both functions.

```python
@strawberry.type
class Mutation:
    login: User = auth.login
    # vs
    login: User = auth.login()
    # vs
    login: User = strawberry.mutation(resolver=login_resolver)
```

Function would allow us to add new parameters in the future if we need to make login mutation configurable. See following example.

```python
@strawberry.type
class Mutation:
    login: User = auth.login(only_superusers=True)
```

Do you agree?

strawberry_django/fields/types.py
la4de: That's good point

strawberry_django/auth/queries.py
la4de: Many characters to write if you care :)

I thought that it would be nice to provide simple API

```python
@strawberry.type
class Query:
     me: User = auth.current_user()

@strawberry.type
class Mutation:
    login: User = auth.login()
    logout: auth.logout()
```

Instead of more verbose syntax

```python
@strawberry.type
class Query:
     me: User = strawberry_django.field(resolver=auth.resolve_current_user)

@strawberry.type
class Mutation:
    login: User = strawberry.mutation(resolver=auth.resolve_login)
    logout: strawberry.mutation(resolver=auth.resolve_logout)
```

la4de:
v0.2.0 published.

Let's continue discussion about type registration / overriding. API in general and documentation.

Thanks for your contribution and feedback.

strawberry_django/auth/mutations.py
joeydebreuk: To be honest, I don't really care much for this login function, especially if we have to cover a lot of different cases. I'd prefer to just add it in the documentation in a "how to", then everyone can alter it to their use case. Its extremely simple if you need to cover 1 case. It will be a mess if it needs to cover all cases.

strawberry_django/auth/queries.py
joeydebreuk: I ment
```py
current_user = strawberry_django.field(resolver=resolve_current_user)

@strawberry.type
class Query:
     me: User = current_user
```
Or better:
```py
@strawberry_django.field
def current_user(info):
    return info.context.request.user

@strawberry.type
class Query:
     me: User = current_user
```

PR #34, opened 2021-05-26
Proposal: Function based filters

la4de:
This draft pull request adds support for function based filters. Let's take an example.

```python
def filter_name(queryset, name: str):
    return queryset.filter(name__contains=name)

@strawberry.type
class Query:
    fruits: List[Fruit] = strawberry_django.field(filters=filter_name)
```

Function `filter_name` is called every time someone queries `fruits` field of Query type. Code above would generate following schema.

```graphql
type Query {
  fruits(name: String): [Fruit!]!
}
```

It is possible to attach filter directly into type. Following code would generate exactly same schema.

```python
@strawberry_django.type(models.Fruit, filters=filter_name)
class Fruit:
    id: auto
    name: auto

@strawberry.type
class Query:
    fruits: List[Fruit] = strawberry_django.field()
```

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,3 +1,4 @@
+import inspect
 import strawberry
 from typing import Generic, List, Optional, TypeVar, Union
 from strawberry.field import StrawberryField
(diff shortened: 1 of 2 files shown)


bellini666:
This has become stale and probably won't be updated in the future.

Issue #40, opened 2021-06-18
Unwanted argument in model Types for O2O and ForeignKeys


g-as:
For models with a ForeignKey (or a O2O), a `pk` arg appears on the generated type, which, to my knowledge doesn't do anything.

Here is the schema used in tests:
```
type BerryFruit {
  id: ID!
  name: String!
  nameUpper: String!
  nameLower: String!
}

type Group {
  id: ID!
  name: String!
  users: [User!]
}

type Query {
  user(pk: ID): User!
  users: [User!]!
  group(pk: ID): Group!
  groups: [Group!]!
  berries: [BerryFruit!]!
}

type User {
  id: ID!
  name: String!
  group(pk: ID): Group
}
```
The `type User` has a `group(pk: ID): Group`. `group: Group` should be enough.

g-as:
Solved by #41

PR #41, opened 2021-06-18
Removes pk argument on model relations

g-as:
Test schema becomes:

```
type BerryFruit {
  id: ID!
  name: String!
  nameUpper: String!
  nameLower: String!
}

type Group {
  id: ID!
  name: String!
  users: [User!]
}

type Query {
  user(pk: ID): User!
  users: [User!]!
  group(pk: ID): Group!
  groups: [Group!]!
  berries: [BerryFruit!]!
}

type User {
  id: ID!
  name: String!
  group: Group
}
```

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -143,9 +143,10 @@ class StrawberryDjangoFieldFilters:
         if not self.base_resolver:
             filters = self.get_filters()
             if self.django_model and not self.is_list:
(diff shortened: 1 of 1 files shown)


g-as:
Solves #40

la4de:
I agree with you that it does not make sense to have pk filtering with foreign key or one to one relation.

Thanks for your contribution.

Issue #42, opened 2021-07-06
Strawberry 0.68.1 internals change


g-as:
https://github.com/strawberry-graphql/strawberry/pull/1044 introduced some backward incompatible changes, messing with this lib.

No easy fix comes to mind so far.

In the meantime, maybe pinning `strawberry <= 0.68` and releasing a new minor version can mitigate.

patrick91:
@g-as can you post the error you get? I'll try to fix it :)

la4de:
> @g-as can you post the error you get? I'll try to fix it :)

I updated strawberry to the latest, run unit tests and got following error.

```
ERROR    strawberry.execution:schema.py:96 get_result() got an unexpected keyword argument 'args'
Traceback (most recent call last):
  File "/Users/lahintsa/Library/Caches/pypoetry/virtualenvs/strawberry-graphql-django-3j3kIc2H-py3.9/lib/python3.9/site-packages/graphql/execution/execute.py", line 617, in resolve_field
    result = resolve_fn(source, info, **args)
  File "/Users/lahintsa/Library/Caches/pypoetry/virtualenvs/strawberry-graphql-django-3j3kIc2H-py3.9/lib/python3.9/site-packages/strawberry/middleware.py", line 29, in resolve
    result = next_(root, info, **kwargs)
  File "/Users/lahintsa/Library/Caches/pypoetry/virtualenvs/strawberry-graphql-django-3j3kIc2H-py3.9/lib/python3.9/site-packages/strawberry/schema/schema_converter.py", line 376, in _resolver
    result = field.get_result(_source, args=args, kwargs=kwargs)
TypeError: get_result() got an unexpected keyword argument 'args'
```

We are extending `StrawberryDjangoField` from `StrawberryField` and implementing resolver to `get_result` method. That's done similar way what is proposed in https://github.com/strawberry-graphql/strawberry/issues/920. `StrawberryDjangoField` does not use `base_resolver` which seems to cause the problem. It is expected in strawberry-graphql/strawberry#1044  that `base_resolver` is used.

This is easy to reproduce by cloning `strawberry-graphql-django` repo, installing dependencies, updating `strawberry-graphql` and running unit tests.

I'm feeling that the workaround is to provide dummy `base_resolver`...

g-as:
> This is easy to reproduce by cloning strawberry-graphql-django repo, installing dependencies, updating strawberry-graphql and running unit tests.

[...]

PR #43, opened 2021-07-06
temporarily pinned strawberry upper bound to 0.68

g-as:
See #42


la4de:
Fixed in `strawberry-graphql` v0.68.2 and `strawberry-graphql-django` v0.2.3

PR #47, opened 2021-07-21
Use Strawberry earlier than 0.69

patrick91:
Strawberry 0.69 has quite a few changes that are not compatible with
strawberry-django, we'll send another PR to fix them, but for now we 
should pin the version


la4de:
I spent few hours looking at changes in strawberry-graphql. Changes are so big that I could not fix django integration. We would need more time to fix them all. Temporal solution is to use old strawberry package.

PR #53, opened 2021-08-03
Add permissions to django mutations

wellzenon:
Django mutations couldn't handle permissions, but adding the keyword argument `permission_classes` to each django mutation allow the permissions to work as it already inherits the functionality from the StrawberryField class, it just needed to be wired


la4de:
@wellzenon, this is great addition. Could you please add a test case for this too?

I apologize for the delay in my response.

la4de:
Tests added

PR #45, opened 2021-07-17
Feature: Register mutation

NeoLight1010:
This PR adds an `auth.register()` mutation as discussed on #37. In contrast to the usual `create()` mutation, this one validates the password using Django's `validate_password()`.

### Example

```python
@strawberry.type
class Mutation:
    register: User = auth.register(UserInput)
```

### Notes
I forgot my editor formats the code on save, so there are some extra newlines in the code. Please tell me if that's an issue.

PR #62, opened 2021-10-15
Fix a bug related to creating users with unhashed passwords

illia-v:
This fixes the security flaw of #45 where `auth.register` saves unhashed passwords.


la4de:
Thanks!

PR #63, opened 2021-10-15
Allow specifying a maximum pagination limit and default offset and limit

illia-v:
This proposes adding a `pagination_config` argument to fields and types for specifying a maximum pagination limit and default offset and limit.

### DIFF
--- a/strawberry_django/pagination.py
+++ b/strawberry_django/pagination.py
@@ -1,4 +1,5 @@
-from typing import List
+import dataclasses
+from typing import List, Optional, Union
 
 import strawberry
 from strawberry.arguments import UNSET, StrawberryArgument, is_unset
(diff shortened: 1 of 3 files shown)


la4de:
Hi @illia-v,

Thanks for your contribution. I like this improvement but I'm still thinking about the API. Do we want to introduce new argument or could we reuse existing argument?

Here are few options for the API.
1. reuse exisiting `pagination` argument
1. introduce new `pagination_config` argument
1. introduce new field for pagination

```python
pagination_config = PaginationConfig(default_offset=1, default_limit=1)

@strawberry.type
class Query:
    fruits_pagination_option1: List[Fruit] = strawberry_django.field(pagination=pagination_config)
    fruits_pagination_option2: List[Fruit] = strawberry_django.field(pagination_config=pagination_config)
    fruits_pagination_option3: List[Fruit] = strawberry_django.pagination_field(default_offset=1, default_limit=1)    
```

In the future we may want to support other complex pagination methods too. For example pagination from Relay project. See the spec https://relay.dev/graphql/connections.htm. I'm not sure how the API would look like. The biggest challenge is that Relay pagination changes the output type. One option would be to implement new separate fields for paginations (option 3).

What do you and others think about that?

illia-v:
@la4de thanks for taking a look at this. I like the third option the most.

la4de:
Thanks for your feedback, @illia-v.

Should we start working on option 3, introduce `pagination_field` and add the `default_offset` and `default_limit` parameters, you implemented, under that new field?

star2000:
In my case, only the max limit per page is useful.
```py
import strawberry_django.pagination


max_per_page_limit = 50


def apply_pagination(pagination, queryset):
    if is_unset(pagination) or pagination is None:
        return queryset[:max_per_page_limit]
    if pagination.limit > max_per_page_limit:
        raise Exception('More than the maximum number of per page limit')
    start = pagination.offset
    stop = start + pagination.limit
    return queryset[start:stop]


strawberry_django.pagination.apply = apply_pagination

```

PR #72, opened 2021-11-25
updated the way event loop is detected in 'is_async'

g-as:
See https://github.com/django/django/pull/14345.


la4de:
Looks good to me. Thank you!

PR #76, opened 2021-12-10
Fix incompatibility with Strawberry >= 0.92.0 related to interfaces

illia-v:
Strawberry [0.92.0](https://github.com/strawberry-graphql/strawberry/releases/tag/0.92.0) stopped allowing non-Strawberry types to be returned from fields when a type implements an interface.

This started causing errors like this one when `strawberry.django.type` is used to create a type implementing an interface:
```
graphql.error.graphql_error.GraphQLError: Expected value of type 'BerryFruit' but got: <Fruit instance>.
```

The proposed solution is setting `is_type_of` before Strawberry has a chance to do this and allow model instances to be returned too.


la4de:
The change looks good. Thanks.

I'm still thinking about testing. Would it make sense to add dedicated test suite for interface? What do you think?

illia-v:
> I'm still thinking about testing. Would it make sense to add dedicated test suite for interface? What do you think?

I added a separate test for this case. A dedicated suite may be useful, but only querying types implementing interfaces caused the error.

PR #78, opened 2021-12-16
Updated docs

ccsv:
Changes:
- changed `strawberry_django` to `strawberry.django` 
- added file name in python notes
- added `schema = strawberry.Schema( mutation=Mutation)` to mutations
- added note about custom field types from other libraries (e.g: if I use auto the RichTextField from ckeditor library it will error)


la4de:
Good job and thanks for your contribution!

PR #90, opened 2022-02-17
Fixed issue with generating order args

jaydensmith:
In this example, the `supplier__name` order arg was generated first, and the subsequent `name` arg was also prefixed with `supplier__`.

```graphql
query {
  inventoryItems(order: {
    supplier: {
      name: ASC
    },
    name: ASC
  }) {
    name
    supplier {
      name
    }
  }
}


la4de:
good catch and thanks for your contribution

PR #91, opened 2022-02-18
feat: add default typing for Django GEOS fields

hiporox:
Django support GIS fields, but it doesn't look like Strawberry does yet. This will add default typing for each of the Django Geometry fields so that a custom type doesn't need to be defined by users for each one


la4de:
Implementation looks good to me. Could add test cases for new types?

la4de:
I left my previous comment too early. Tests failed.

This adds new dependency but by default we do not want to depend on gdal. I propose to change the implementation so that new types are added only if gdal library is available.

How about doing something like this?
```python
try:
    from django.contrib.gis.db import models as geos_fields
    field_type_map.update({
        geos_fields.PointField: Point,
        ...
    })
except django.core.exceptions.ImproperlyConfigured:
    pass # gdal library not available
```

strawberry_django/fields/types.py
```diff
@@ -60,6 +61,13 @@ class ManyToManyInput:
     set: Optional[List[strawberry.ID]] = UNSET
 
 
+Point = strawberry.scalar(
+    NewType("Point", (Optional(float), Optional(float), Optional(float))),
```
bellini666, line 65: I don't think this is correct.

1) The typing in this case should be `Tuple[Optional(float), Optional(float), Optional(float)]`

2) The arguments are not actually optional, trying to pass `None` to any of those is an error

3) The third argument is not mandatory. You can create a point with 2 or 3 floats (you just can't use 1 or more than 3)

The best case in here would be to use `Union[Tuple[float, float], Tuple[float, float, float]]`, but `NewType` doesn't accept it. The correct way would probably be `Tuple[float, ...]`

bellini666 (changes requested):

hiporox:
I have address both of your comments @bellini666 @la4de 

I am ready going to go write some tests, but I am realizing that with the change that @la4de suggested, none of these will build without adding `gdal` to the CI. How should I best deal with that?

hiporox:
@bellini666 Thinking about your comment, would it be better to replace instances of  `Point` with `Union[Tuple[float, float], Tuple[float, float, float]]`? Just do away with using NewType all together

strawberry_django/fields/types.py
```diff
@@ -102,6 +103,31 @@ class ManyToManyInput:
...
+            geos_fields.LineStringField: List[Point],
+            geos_fields.PolygonField: List[Point],
+            geos_fields.MultiPointField: List[Point],
+            geos_fields.MultiLineStringField: List[List[Point]],
+            geos_fields.MultiPolygonField: List[List[Point]],
+        }
+    )
+except django.core.exceptions.ImproperlyConfigured:
+    # in gdal is not available
+    # skip
```
bellini666, line 127: I think you mean `if`, not `in`. Also, one line for this comment should be enough, like `# If gdal is not available, skip.`

bellini666 (changes requested):

bellini666:
> I am ready going to go write some tests, but I am realizing that with the change that @la4de suggested, none of these will build without adding `gdal` to the CI. How should I best deal with that?

You can install it in the cicd script. Add those lines there, before `poetry install`:

```yaml
- name: Install gdal to test geofields
  run: sudo apt-get install gdal-bin
```

> @bellini666 Thinking about your comment, would it be better to replace instances of Point with Union[Tuple[float, float], Tuple[float, float, float]]? Just do away with using NewType all together

Hrm, if that works fine without any downsides, I think it is better.

hiporox:
> > I am ready going to go write some tests, but I am realizing that with the change that @la4de suggested, none of these will build without adding `gdal` to the CI. How should I best deal with that?
> 
> You can install it in the cicd script. Add those lines there, before `poetry install`:
> 
> ```yaml
> - name: Install gdal to test geofields
>   run: sudo apt-get install gdal-bin
> ```
> 
> > @bellini666 Thinking about your comment, would it be better to replace instances of Point with Union[Tuple[float, float], Tuple[float, float, float]]? Just do away with using NewType all together
> 
> Hrm, if that works fine without any downsides, I think it is better.

Great, thank you for the help. I am going to need some time to figure out how to enable SQLite GIS support and get it to work with the test suite, but once I do, I will get this PR completed

hiporox:
@la4de Can I ask you to run tests on this? I want to check that Spatialite will work properly in CI since it was a pain for me to get it set up locally

hiporox:
@la4de Found a way for it to work without needed Spatialite (since we don't have to write the GeosFields to the DB). It should be ready to test and review

strawberry_django/fields/types.py
```diff
@@ -1,10 +1,11 @@
 import datetime
 import decimal
 import uuid
-from typing import List, Optional
+from typing import List, Optional, Tuple, Union
 
 import django
 import strawberry
+from django.contrib.gis.db import models as geos_fields
```
la4de, line 8: Please remove this import. geos_fields is reimported in try block below. pre-commit check fails due that

.gitignore
```diff
@@ -1,4 +1,145 @@
-*.pyc
-*.egg-info
-__pycache__
-/dist
+# Byte-compiled / optimized / DLL files
```
la4de, line 1: I propose to remove `.gitignore` changes from this pull request. Please create new pull request if you think `.gitignore` has to be updated.

la4de (changes requested):

.gitignore
hiporox: Sounds good, will do

bellini666:
Hey @hiporox , thanks for updating this PR so far and sorry for taking some time to review this again.

I think the PR is in a good shape to be merged! I just would like to see a test of some django geos values being parsed/serialized by the scalar in a schema call to make sure that they work properly and will not break in the future.

bellini666 (changes requested):
(check my previous comment)

kwongtn:
@bellini666 would like to bump this PR and see if anything else requires to be amended? 😁

bellini666:
Hi @kwongtn . The only thing missing from this PR are some tests, that I requested in a previous commit. Other than that it is ok to be merged

hiporox:
Closing as it is replaced by #213

PR #93, opened 2022-02-24
Caching Extensions using Django Cache

hiporox:
Implementations of Strawberry's `ParserCache` and `ValidationCache`, but using Django's cache instead of the `functools.lru_cache`

This is blocked by upgrading `strawberry-graphql` to `>=0.98.0`, since that will get us access to `graphql-core@3.2.0`, which fixes this issue https://github.com/graphql-python/graphql-core/issues/112 which stops unpickling `DocumentNode`


bellini666:
Hey @hiporox .

I can see how the validation extension being cached using django's cache would be useful, but I wonder in which scenario the parsing would would be? It seems to me that parsing should be faster than requesting an external resource for it (unless it is a very complicated query?)

hiporox:
@bellini666 I just copied over the functionality of the two types of caches that base strawberry already supports. It is very possible that the parsing cache will just be slower. I have not tested the speed of it in comparison to a regular query lookup. I am very happy to drop that chunk of the code and just leave the validation cache

bellini666:
> @bellini666 I just copied over the functionality of the two types of caches that base strawberry already supports. It is very possible that the parsing cache will just be slower. I have not tested the speed of it in comparison to a regular query lookup. I am very happy to drop that chunk of the code and just leave the validation cache

Yeah, I think it is better to only have the validation cache here :)

hiporox:
@bellini666 Removed the parser cache and fixed the tests. How does it look now?

bellini666 (changes requested):
Just a few typing issues and it is good to merge! :)

strawberry_django/extensions/django_cache_base.py
```diff
@@ -0,0 +1,52 @@
...
+class DjangoCacheBase(Extension):
+    """
+    Base for a Cache that uses Django built in cache instead of an in memory cache
+
+    Arguements:
+
+    `cache_name: str`
+        Name of the Django Cache to use, defaults to 'default'
+
+    `timeout: Optional[int | float]`
```
bellini666, line 23: int is also a valid float, no need to use the union here

```suggestion
    `timeout: Optional[float]`
```

strawberry_django/extensions/django_cache_base.py
```diff
@@ -0,0 +1,52 @@
...
+        A function to use to generate the cache keys
+        Defaults to the same key generator as functools.lru_cache
+        WARNING! The default function does NOT work with memcached
+        and will generate warnings
+    """
+
+    def __init__(
+        self,
+        cache_name: str = "default",
+        timeout: Optional[Union[int, float]] = None,
```
bellini666, line 37:
```suggestion
        timeout: Optional[float] = None,
```

bellini666 (changes requested):
Just 2 minor details missing

bellini666:
@hiporox the PR is fine, just need to make the tests pass and I'll merge it!

hiporox:
@bellini666 Tests pass now, not sure what is up with precommit

bellini666:
Thanks for the contribution @hiporox :)

PR #98, opened 2022-02-25
docs: filled in some missing info in the docs

hiporox:


docs/references/django_model_types.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Fields
+
+By default, no fields are implemented on the new type. For details on adding fields,
+see the [Fields](fields.md) documentation.
+
+## Queryset setup
+
+By default, a Strawberry Django type will pull data from the default manager for it's Django Model.
+You can change this behavior by implementing a `get_queryset` method on the type.
```
bellini666, line 21: Don't know if "change this behavior" is the correct phrase here. The reason one would override this would be to do some extra filtering, do some annotation or something like that.

Maybe rewrite this to something like "You can implement a `get_queryset` to your type to do some extra processing to the queryset, like further filtering it."

docs/references/django_model_types.md
```diff
@@ -0,0 +1,49 @@
...
+class Fruit:
+    ...
+```
+
+## Fields
+
+By default, no fields are implemented on the new type. For details on adding fields,
+see the [Fields](fields.md) documentation.
+
+## Queryset setup
```
bellini666, line 18: Django defines this as `QuerySet` instead of `Queryset`

docs/references/django_model_types.md
```diff
@@ -0,0 +1,49 @@
...
+
+```python
+@strawberry.django.type(models.Fruit)
+class Berry:
+
+    def get_queryset(self, queryset, info):
+        return queryset.filter(name__contains="berry")
+```
+
+The `get_queryset` method is given a Queryset to filter and
```
bellini666, line 31: `QuerySet`

docs/references/django_model_types.md
```diff
@@ -0,0 +1,49 @@
...
+a Strawberry `Info` object, containing details about the request.
+
+You can use the `info` parameter to limit access to items.
```
bellini666, line 34: Maybe "You can use that `info` parameter to, for example, limit access to results based on the current user in the request:"

docs/references/django_model_types.md
```diff
@@ -0,0 +1,49 @@
...
+
+    def get_queryset(self, queryset, info):
+        if info.context['request'].user.is_staff:
+            return queryset.filter(
+                name__contains="berry",
+            )
+        return queryset.filter(
+            name__contains="berry",
+            is_top_secret=False,
+        )
```
bellini666, line 48: Maybe rewrite this for clarity in the example:

```python
def get_queryset(self, queryset, info):
    if not info.context.request.user.is_staff:
        # Limit access to top secret if the user is not a staff member
        queryset = queryset.filter(
            is_top_secret=False,
        )

    return queryset.filter(
        name__contains="berry",
    )
```

docs/references/filters.md
```diff
@@ -62,9 +62,20 @@ class ColorFilter:
...
-## Overriding default filtering method
+## Custom filters and overriding default filtering method
 
-TODO
+You can define custom filter methods and override default filter methods by defining your own resolver.
+
+```python
+@strawberry.django.filters.filter(models.Fruit)
+class FruitFilter:
+    is_banana: bool
```
bellini666, line 72: A better example would be to type this as `bool | None`, to show that we can have non-mandatory filter attributes.

(or define a mandatory and a non-mandatory one)

docs/references/resolvers.md
```diff
@@ -45,3 +45,46 @@ class Color:
...
+It is important that if you override resolvers, you will lose access to all of the things that come by default
+with the library (e.g. `Pagination`, `Filter`). For Types that are attached to your base `Query`, it
+is recommended to define a custom `get_queryset` on the Type
+(see [Django Model Types](django_model_types.md) for details).
+
+For example, if we wanted a look up for berries and one for non-berry fruits.
+
+```python
+
+#types.py
```
bellini666, line 60: `# types.py`

docs/references/resolvers.md
```diff
@@ -45,3 +45,46 @@ class Color:
             return list(self.fruits.objects.filter(...))
         return query()
 ```
+
+## Issues with Resolvers
+
+It is important that if you override resolvers, you will lose access to all of the things that come by default
```
bellini666, line 51: Maybe rewrite as `It is important to note that ...`

docs/references/resolvers.md
```diff
@@ -45,3 +45,46 @@ class Color:
...
+with the library (e.g. `Pagination`, `Filter`). For Types that are attached to your base `Query`, it
+is recommended to define a custom `get_queryset` on the Type
```
bellini666, line 53: I wouldn't say that it is recommended. I think that it is important to explain that for root fields (the ones defined on `Query`), `get_queryset` can be used.

But I would add a warning there, because it might confuse people into thinking that it would always be used like that, and it would affect things like retrieving related fields.

bellini666 (changes requested):

docs/references/django_model_types.md
hiporox: @bellini666 Will `info.context` has a property called `request`? I thought it was a dictionary that you would need to use square bracket lookups for

docs/references/django_model_types.md
```diff
@@ -0,0 +1,49 @@
...
+
+@strawberry.django.type(models.Fruit)
+class Fruit:
+    ...
+```
+
+## Fields
+
+By default, no fields are implemented on the new type. For details on adding fields,
+see the [Fields](fields.md) documentation.
```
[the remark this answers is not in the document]
hiporox: What do you mean by this?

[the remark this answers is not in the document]
bellini666: I mean that there is already a documentation page for the types [here](https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/docs/references/types.md).

This `django_model_type` is basically a duplicated page of it. IMO the info here should be used to improve that page, or else we end up with duplicated documentation content.

docs/references/django_model_types.md
bellini666: @hiporox yes it does! You can actually access it both ways, by either doing `info.context.request` and `info.context["request"]`, but the attribute access has better static typing support in the current code (unless the one using it types the context themselves using a TypeDict).

docs/references/django_model_types.md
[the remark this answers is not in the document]
hiporox: Ah, okay. I have moved all of the info from this file into `type` already, so I think we can close this comment chain

hiporox:
@bellini666 I think that addresses all of your comments. How does it seem now?

docs/references/filters.md
```diff
@@ -62,9 +62,22 @@ class ColorFilter:
...
 
-TODO
+You can define custom filter methods and override default filter methods by defining your own resolver.
+
+```python
+@strawberry.django.filters.filter(models.Fruit)
+class FruitFilter:
+    is_banana: bool | None
+
+    def is_banana(self, queryset):
```
bellini666, line 74: Isn't this `filter_is_banana`?

docs/references/resolvers.md
```diff
@@ -45,3 +45,45 @@ class Color:
             return list(self.fruits.objects.filter(...))
         return query()
 ```
+
+## Issues with Resolvers
+
+It is important to note that if you override resolvers, you will lose access to all of the things that come by default
+with the library (e.g. `Pagination`, `Filter`). On your root `Query`, you can just use a custom `get_queryset` to achieve
+similar results, while keeping pagination and filtering intact.
```
bellini666, line 53: What do you think?

```suggestion
similar results, while keeping pagination and filtering intact, but note that it will affect all root queries for that type.
```

bellini666 (changes requested):
Much better :)

Just a few comments and we can merge this

docs/references/filters.md
hiporox: Yes it is, good point

bellini666:
@hiporox this PR is fine to merge, just need to fix issues with pre-commit.

hiporox:
@bellini666 Fixed the issue that pre-commit had, but it is now error in black for some reason. Not sure what is going on there, but I believe it should be good

bellini666:
> @bellini666 Fixed the issue that pre-commit had, but it is now error in black for some reason. Not sure what is going on there, but I believe it should be good

Yeah, there seems to be an issue with pre-commit. I'm going to take a look at it soon.

But thanks for the contribution, looks good now :)

PR #100, opened 2022-03-03
feat: add Enum support to filtering

hiporox:
Often in Django you will add `ChoiceField` to a model. With Django >= 3.0, you can now build enums that work for choices (https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types). Using this, we can use the same Choices to build Strawberry Enums to use for filters.

For example, you can now:
```python
class ColorChoices(TextChoices):
    blue = "BLUE"
    red = "RED"

class Bags(models.Model):
    color = models.ChoiceField(TextChoices)

@strawberry.enum
class ColorEnum(ColorChoices):
    ...

@strawberry_django.filters.filter(Bags)
class BagFilter:
    color: ColorEnum
```

This will allow the GraphQL to expose the Enum options and enable filtering using that Enum.

Prior to this, a user would have to define the enum filter method explicitly using:
```python
@strawberry_django.filters.filter(Bags)
class BagFilter:
    color: ColorEnum

    def filter_color(self, queryset):
         return queryset.filter(color=self.color.value)
```

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import Generic, List, Optional, TypeVar
 
 import strawberry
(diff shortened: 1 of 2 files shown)


bellini666 (approved):
Didn't notice that enums weren't working for filtering...

LGTM, thanks for the contribution! :)

PR #101, opened 2022-03-07
Use auto from strawberry instead of define our own

bellini666:
This makes it possible to use "strawberry.auto" directly instead of
having to use own own auto.

We are still exporting auto in our `__init__` file for backwards
compatibility, but we should remove that as well in the future.

### DIFF
--- a/strawberry_django/fields/types.py
+++ b/strawberry_django/fields/types.py
@@ -1,16 +1,14 @@
 import datetime
 import decimal
 import uuid
-from typing import Any, List, Optional, get_args, get_origin
+from typing import List, Optional
 
 import django
 import strawberry
 from django.db.models import fields
 from django.db.models.fields.reverse_related import ForeignObjectRel, ManyToOneRel
-from strawberry.annotation import StrawberryAnnotation
 from strawberry.arguments import UNSET
-from strawberry.auto import StrawberryAuto, auto
-from typing_extensions import Annotated
+from strawberry.auto import StrawberryAuto
 
 from .. import filters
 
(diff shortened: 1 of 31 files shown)


strawberry_django/fields/types.py
```diff
@@ -165,14 +163,26 @@ def get_model_field(model, field_name):
         raise e
 
 
+# FIXME: maybe this should be defined at strawberry/auto.py?
```
patrick91, line 166: +1

patrick91 (approved):
Nice one! maybe we can also add a deprecation warning when someone uses auto from this package :)

bellini666:
> Nice one! maybe we can also add a deprecation warning when someone uses auto from this package :)

Hey @patrick91 . Sorry for the delay, it has been a complicated week for me :(

I thought the same, but since I totally removed `auto` from this package, when someone tries to import it from `fields/types.py` it will actually be importing the one from strawberry. And there's no way to add a warning on that specific import also.

But this can be done as: Keep the current `auto` as well and support both, but raise a deprecation warning if it is used. Then we proceed to totally remove it in a future release.

What do you think?

strawberry_django/fields/types.py
bellini666: Will open a PR on strawberry :)

bellini666:
@patrick91 totally forgot about this PR. Am starting to take a look at the open PRs and saw mine in the list :)

Just added the suggested deprecation to `auto`/`is_auto` from this package. Do you want to take another look at it?

patrick91:
Looks good to me 😊 feel free to merge 😊

PR #103, opened 2022-03-15
Provide a Django TestCase sub-class to facilitate testing of Strawberry APIs in Django

joshourisman:
Inspired by graphene-python's Django integration, provides a sub-class of Django's TestCase that includes a helper method for testing your GraphQL API with real Django requests. This enables things like authentication, which otherwise are not possible, because there is no Django HTTPRequest provided if you execute queries against your schema manually.

### DIFF
--- a/strawberry_django/tests.py
+++ b/strawberry_django/tests.py
@@ -1,0 +1,52 @@
+import json
+from typing import Optional
+
+from django.test import Client, TestCase
+
+
+class StrawberryTestCase(TestCase):
+    """
+    Based on: https://www.sam.today/blog/testing-graphql-with-graphene-django/ and https://docs.graphene-python.org/projects/django/en/latest/testing/
+    """
+
+    def setUp(self):
+        self._client = Client()
+
+    def query(
+        self,
+        query: str,
+        op_name: Optional[str] = None,
+        variables: Optional[dict] = None,
+    ):
+        """
+        Args:
+            query (string)   - GraphQL query to run
+            op_name (string) - If the query is a mutation or named query, you must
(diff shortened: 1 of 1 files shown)


strawberry_django/tests.py
```diff
@@ -0,0 +1,52 @@
...
+from django.test import Client, TestCase
+
+
+class StrawberryTestCase(TestCase):
+    """
+    Based on: https://www.sam.today/blog/testing-graphql-with-graphene-django/ and https://docs.graphene-python.org/projects/django/en/latest/testing/
+    """
+
+    def setUp(self):
+        self._client = Client()
```
bellini666, line 13: Could we use strawberry's test client in here? https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/django/test/client.py#L6

strawberry_django/tests.py
```diff
@@ -0,0 +1,52 @@
...
+        if op_name:
+            body["operation_name"] = op_name
+        if variables:
+            body["variables"] = variables
+
+        resp = self._client.post(
+            "/graphql", json.dumps(body), content_type="application/json"
+        )
+        jresp = json.loads(resp.content.decode())
+        return jresp
```
bellini666, line 44: Then it could be used here and return the response typed as in there

bellini666 (changes requested):
Sorry for the late review.

bellini666:
v0.10.0 introduced a test client that can be used to solve the issues this PR was trying to solve: https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/strawberry_django/test/client.py

And since this has been stale for a while, I'm going to close it

PR #108, opened 2022-04-15
Fix ordering with custom filters

hiporox:
The current MRO applies the ordering before any filters are applied. This is an issue if your filter uses any annotations to do its filtering, as `Queryset`s lose all of their ordering when an annotation is applied. To solve for this, we can change the MRO of `get_queryset` on `StrawberryDjangoField` to apply the filters first, and the ordering after, there by removing the issue


bellini666:
Thanks for the contribution and sorry for the late review

PR #109, opened 2022-04-16
Fix filtering cannot use relational reflection fields

star2000:
Closed https://github.com/strawberry-graphql/strawberry-graphql-django/issues/84

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,4 +1,6 @@
 from enum import Enum
+from functools import lru_cache
+from itertools import chain
 from typing import Generic, List, Optional, TypeVar
 
 import strawberry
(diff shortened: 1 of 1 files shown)


strawberry_django/filters.py
```diff
@@ -72,6 +74,19 @@ def filter_deprecated(model, *, name=None, lookups=False):
...
+def get_field_names_from_opts(opts):
+    '''
+    copy from django.db.models.sql.query.get_field_names_from_opts
+    because get_field_names_from_opts not in django.db.models.sql.query.__all__
+    and no cache optimization
+    '''
+    return set(chain.from_iterable(
+        (f.name, f.attname) if f.concrete else (f.name,)
+        for f in opts.get_fields()
+    ))
```
bellini666, line 87: It's not a problem that the function is not in `__all__`. That only matters when you do `from <some_place> import *`

Just import it from there and use it directly :)

bellini666 (changes requested):
Hi, thanks for the contribution and sorry for the late review. Just a small change and this should be good to merge

strawberry_django/filters.py
star2000: I think functions not in __all__ are meant for internal use only and may be renamed or removed later

bellini666 (approved):

bellini666:
Thanks for the contribution :)

PR #115, opened 2022-05-17
bugfix: ignore filters argument if it is an arbitary argument

devkral:
## Description

The current behavior is that every filters argument is treated as filter. But filters can also be an arbitrary argument which should not trigger the filter logic.
This MR prevents this by applying checks if filters is really a filter 

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


bellini666:
LGTM! Thanks for the contribution :)

PR #116, opened 2022-05-17
docs: document how to use a custom filter logic

devkral:
## Description

Document how to overwrite default filter logic.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

## Checklist

### DIFF
--- a/docs/references/filters.md
+++ b/docs/references/filters.md
@@ -79,10 +79,11 @@ class FruitFilter:
         return queryset.exclude(name='banana')
 ```
 
-## Custom filter logic
+## Overriding the filter method
 
 For overriding the default filter logic you can provide the filter method.
-Other filter methods are not executed anymore (default as well as custom).
+Note that no default filtering will be done anymore, which means your custom
+method is responsible for handling all the all filter related operations.
 
 ```python
 @strawberry.django.filters.filter(models.Fruit)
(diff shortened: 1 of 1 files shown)


docs/references/filters.md
```diff
@@ -79,6 +79,24 @@ class FruitFilter:
         return queryset.exclude(name='banana')
 ```
 
+## Custom filter logic
```
bellini666, line 82:
```suggestion
## Overriding the filter method
```

docs/references/filters.md
```diff
@@ -79,6 +79,24 @@ class FruitFilter:
...
+
+For overriding the default filter logic you can provide the filter method.
+Other filter methods are not executed anymore (default as well as custom).
```
bellini666, line 85: I would be more clear here of what exactly does it mean to override the `filter` method. Not everyone should be doing it IMO.

Don't know if the suggestion here is enough (feel free to suggest something different)

```suggestion
Note that no default filtering will be done anymore, which means your custom
method is responsible for handling all the all filter related operations.
```

bellini666 (changes requested):
Hi, thanks for the contribution and sorry for the late review.

bellini666 (approved):

PR #122, opened 2022-06-06
Change order of inheritance for `StrawberryDjangoField`

hiporox:
The current MRO of `StrawberryDjangoField` has it applying pagination before the `StrawberryDjangoFieldBase.get_queryset` resolves. Since pagination does array slicing, if you have a custom `get_queryset` defined on the type, it will raise the following error:
```
AssertionError: Cannot filter a query once a slice has been taken.
```

By changing the MRO to have pagination come after field base, we will avoid this issue, since pagination will be applied last

### DIFF
--- a/strawberry_django/fields/field.py
+++ b/strawberry_django/fields/field.py
@@ -24,10 +24,10 @@ class StrawberryDjangoFieldBase:
 
 
 class StrawberryDjangoField(
-    StrawberryDjangoFieldFilters,
-    StrawberryDjangoFieldOrdering,
-    StrawberryDjangoFieldBase,
     StrawberryDjangoPagination,
+    StrawberryDjangoFieldOrdering,
+    StrawberryDjangoFieldFilters,
+    StrawberryDjangoFieldBase,
     StrawberryField,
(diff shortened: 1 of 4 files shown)


bellini666:
@hiporox although the MRO usually matters, in this case the fix can be even easier.

Do you know if changing this (in [this line](https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/strawberry_django/pagination.py#L49)):

```python
    def get_queryset(self, queryset, info, pagination=UNSET, **kwargs):
        queryset = apply(pagination, queryset)
        return super().get_queryset(queryset, info, **kwargs)
```

To this:

```python
    def get_queryset(self, queryset, info, pagination=UNSET, **kwargs):
        queryset = super().get_queryset(queryset, info, **kwargs)
        return apply(pagination, queryset)
```

solves the issue?

hiporox:
@bellini666 That does work, and seems like a much better solution. Let me do that instead and update the PR

bellini666 (approved):
LGTM! 👍🏼

PR #126, opened 2022-06-16
feat: allow Enums to work with FilterLookup

hiporox:
Update `build_filter_kwargs` to handle when an `Enum` is provided to the `FilterLookup`. This is similar to #100, but adds support for if the `Enum` is nested


bellini666 (approved):
LGTM! Thanks :)

PR #131, opened 2022-06-27
fix: resolve `ManyToManyRel` and `ManyToOneRel` as non-null lists

### LINKED ISSUE #128: Dispute on nullability
I think there is a problem regarding generating nullable types but I'm not sure. 

Here is my situation. So I have three models, `A`, `B`, `C` listed below. `A` is a foreign key of `B`, and is in m2m rel with `C` 

```python
from __future__ import annotations

from typing import List

import strawberry
import strawberry_django
from django.core.management import BaseCommand
from django.db import models
from strawberry import auto
from strawberry.printer import print_schema


class A(models.Model):
    pass


class B(models.Model):
    a = models.ForeignKey(A, on_delete=models.PROTECT, related_name="bs", null=False)
    s = models.TextField(null=False)
    sn = models.TextField(null=True)


class C(models.Model):
    a = models.ManyToManyField(A, related_name="cs", null=False)


@strawberry_django.type(A)
class AType:
    bs: List[BType]
    cs: List[CType]


@strawberry_django.type(B)
class BType:
    a: AType
    s: auto
    sn: auto


@strawberry_django.type(C)
class CType:
    a: List[AType]


@strawberry.type
class Query:
    @strawberry.field
    def get_a(self, info) -> List[AType]:
        return A.objects.all()

```

When I declare my type for `A`, I explicitly want the reverse reference `bs` to be the type `[BType!]!`, and `cs` to be the type `[CType!]!`. But if I print the schema, the type resolution gives me `[BType!]` and `[CType!]` instead. 

```graphql
type AType {
  bs: [BType!]
  cs: [CType!]
}

type BType {
  a: AType!
  s: String!
  sn: String
}

type CType {
  a: [AType!]!
}

type Query {
  getA: [AType!]!
}
```

But curiously, the type of `a` in `CType` is correct. 

When I look into the code, the nullability of `bs` and `cs` should be assign [here](https://github.com/strawberry-graphql/strawberry-graphql-django/blob/ab665de24efae7bf7fd31183eecfd931a30963f8/strawberry_django/type.py#L78), which is determined by the function [`is_optioinal`](https://github.com/strawberry-graphql/strawberry-graphql-django/blob/9641f70517141e5001cb573adbf23fac059fdade/strawberry_django/fields/types.py#L168). When a field is not an input nor partial, the code checks if the field can be null (the `null=True` kw arg in the field decleration I think ), and assigns nullability to the annotation. 

I found it problematic since if a field is `ManyToManyRel` or `ManyToOneRel`, there doesn't seem to exist a way to modify the nullability of the rel. I'm not very familiar with Django, and to my best knowledge I can't remember the value carried in these two kinds of rels can ever be `null`/`None`. 

I hope this makes sense :)

FlickerSoul:
## Description

This PR is intended to fix #128. Reverse-related fields like `ManyToManyRel` and `ManyToOneRel` are now resolved as non-null lists, instead of nullable lists, since a query of such relations from the database will always be a list. 

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #128 

## Checklist

### DIFF
--- a/strawberry_django/fields/types.py
+++ b/strawberry_django/fields/types.py
@@ -180,6 +180,10 @@ def is_optional(model_field, is_input, partial):
         has_default = model_field.default is not fields.NOT_PROVIDED
         if model_field.blank or has_default:
             return True
-    if model_field.null:
-        return True
+    if not isinstance(
+        model_field,
+        (fields.reverse_related.ManyToManyRel, fields.reverse_related.ManyToOneRel),
+    ) or isinstance(model_field, fields.reverse_related.OneToOneRel):
+        # OneToOneRel is the subclass of ManyToOneRel, so additional check is needed
+        return model_field.null
     return False
--- a/tests/auth/test_types.py
+++ b/tests/auth/test_types.py
@@ -1,5 +1,5 @@
 from django.contrib import auth
-from strawberry.type import StrawberryList, StrawberryOptional
+from strawberry.type import StrawberryList
 
 import strawberry_django
 from strawberry_django import DjangoModelType, auto, fields
@@ -27,5 +27,5 @@ def test_group_type():
 
     assert [(f.name, f.type or f.child.type) for f in fields(Type)] == [
         ("name", str),
-        ("users", StrawberryOptional(StrawberryList(DjangoModelType))),
+        ("users", StrawberryList(DjangoModelType)),
     ]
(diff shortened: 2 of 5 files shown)


FlickerSoul:
I'm not sure where the CONTRIBUTING document is though. If there is anything I'm missing, please let me know :)

bellini666 (approved):
LGTM

PR #135, opened 2022-07-01
feature: allow overriding field class

benhowes:
## Description
For a project I am working on there are a couple of ways in which I need to extend the behavior of the `StrawberryDjangoField` to accommodate:
1. A different pagination style
2. The ability to null some fields when permission is denied, rather than raising errors. This is for schema backwards compatibility
3. Some altered behavior on the filters

As such, I need to be able to have fields be created with my custom field class rather than the base class.

This PR introduces a `field_cls` argument which can be passed to `@strawberry_django.type` and set to an alternative implementation.

N.B. I am aware some documentation updates may be needed, but I wanted to get some feedback on how acceptable the maintainers find this approach before continuing to work on it.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -1,5 +1,5 @@
 import dataclasses
-from typing import Any, Optional
+from typing import Any, Optional, TypeVar
 
 import django
 import strawberry
@@ -7,7 +7,7 @@ from strawberry import UNSET
 from strawberry.annotation import StrawberryAnnotation
 
 from . import utils
-from .fields.field import StrawberryDjangoField
+from .fields.field import StrawberryDjangoField, StrawberryDjangoFieldBase
 from .fields.types import (
     get_model_field,
     is_optional,
@@ -17,6 +17,10 @@ from .fields.types import (
 
 
 _type = type
+
+StrawberryDjangoFieldType = TypeVar(
+    "StrawberryDjangoFieldType", bound=StrawberryDjangoFieldBase
+)
 
 
 def get_type_attr(type_, field_name):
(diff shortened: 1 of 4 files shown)


tests/utils.py
```diff
@@ -47,3 +47,13 @@ def wrapper(cls):
...
+
+
+def get_field(cls, field_name):
+    matches = [
+        field for field in cls._type_definition.fields if field.name == field_name
+    ]
+    try:
+        return matches[0]
+    except IndexError:
+        raise ValueError(f"{cls} does cont contain a field named {field_name}")
```
bellini666, line 59: Is this required? I feel like we could use [TypeDefinition.get_field](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/types/types.py#L108) from strawberry directly

strawberry_django/type.py
```diff
@@ -32,9 +36,9 @@ def get_field(django_type, field_name, field_annotation=None):
     attr = get_type_attr(django_type.origin, field_name)
 
     if utils.is_field(attr):
-        field = StrawberryDjangoField.from_field(attr, django_type)
+        field = django_type.field_cls.from_field(attr, django_type)
```
bellini666, line 39: I didn't though of all possibilities, but wouldn't this break for someone not implementing his own field since if by default raises `NotImplementedError`?

bellini666 (changes requested):

tests/utils.py
benhowes: TIL, thank you!

bellini666 (approved):

nrbnlulu:
Please wait before you write any changes to the docs, because there is a PR for a gh-pages docs.
#140

PR #144, opened 2022-07-09
feat: link JSONField to strawberry.scalars.JSON

### LINKED ISSUE #143: Field Type Parser for `JSONField`
## Feature Request Type

- [ ] Core functionality
- [ ] Alteration (enhancement/optimization) of existing feature(s)
- [x] New behavior

## Description
I'm not sure if I should put this as a new behavior or as the core functionality. When I'm trying to resolve `JSONField` in my model type, an error is raised saying `NotImplementedError: GraphQL type for model field 'backend.A.json' has not been implemented`. 

The mimimal reproducable code is like the following 
```python
import strawberry
import strawberry_django
from django.core.management import BaseCommand
from django.db import models
from strawberry import auto
from strawberry.printer import print_schema


class A(models.Model):
    json = models.JSONField()


@strawberry_django.type(A)
class AType:
    json: auto


@strawberry.type
class Query:
    pass


schema = strawberry.Schema(query=Query)
```

My proposal is that the value in the `JSONField` can be resolved as a string, which I think is also the default behavior in graphene. 

The following code should be able to do the lifting. It works on `django 3.2.x` with `PostgreSQL`. I'm not sure about the compatibility but I can do some testing to see if this works. 

```python
def _serialize_for_json_type(obj: object | str) -> str:
    try:
        if isinstance(obj, str):
            json.loads(obj)
            return obj
        else:
            return json.dumps(obj)
    except Exception as e:
        raise TypeError(f"{obj} is not a valid type for json")


def _parse_for_json_type(obj: str | object) -> object:
    try:
        if isinstance(obj, str):
            return json.loads(obj)
        else:
            json.dumps(obj)
            return obj
    except Exception as e:
        raise TypeError(f"{obj} is not a valid type for json")


JSONType = strawberry.scalar(
    NewType("JSONType", str | object),
    description="A JSON object or string",
    serialize=_serialize_for_json_type,
    parse_value=_parse_for_json_type,
)
```

FlickerSoul:
## Description

This PR is intended to fix #143. I saw in the discord channel that there already exists a `JSON` type so I'm using that one instead of inventing one as did in #143. 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #143 

## Checklist

### DIFF
--- a/strawberry_django/fields/types.py
+++ b/strawberry_django/fields/types.py
@@ -9,6 +9,7 @@ from django.db.models import fields
 from django.db.models.fields.reverse_related import ForeignObjectRel, ManyToOneRel
 from strawberry import UNSET
 from strawberry.auto import StrawberryAuto
+from strawberry.scalars import JSON
 
 from .. import filters
 
@@ -93,7 +94,7 @@ field_type_map = {
 if django.VERSION >= (3, 1):
     field_type_map.update(
         {
-            fields.json.JSONField: NotImplemented,
+            fields.json.JSONField: JSON,
             fields.PositiveBigIntegerField: int,
         }
     )
--- a/tests/fields/test_types.py
+++ b/tests/fields/test_types.py
@@ -10,6 +10,7 @@ import strawberry
 from django.db import models
 from strawberry import auto
 from strawberry.enum import EnumDefinition, EnumValue
+from strawberry.scalars import JSON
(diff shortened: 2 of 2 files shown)


bellini666 (approved):
Thank you!

PR #160, opened 2022-08-02
feat: Use Django textual metadata in GraphQL 

noelleleigh:
## Description

Automatically add Django model docstring and field `help_text` to the Strawberry type description and field descriptions, respectively.

These descriptions can be overridden by passing the `description` kwarg to `strawberry.django.type()` or `strawberry.django.field()`.

Added tests:

- `tests/test_types.py::test_field_metadata_preserved`
- `tests/test_types.py::test_field_metadata_overridden`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,10 @@ class Fruit(models.Model):
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20,
-           help_text="field description")
+    name = models.CharField(
+      max_length=20,
+      help_text="field description",
+    )
 ```
 
 ```python
--- a/README.md
+++ b/README.md
@@ -28,12 +28,14 @@ Full documentation is available under [docs](https://strawberry-graphql.github.i
 from django.db import models
 
 class Fruit(models.Model):
+    """A tasty treat"""
     name = models.CharField(max_length=20)
     color = models.ForeignKey('Color', blank=True, null=True,
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20)
+    name = models.CharField(max_length=20,
+           help_text="field description")
 ```
 
 ```python
@@ -72,6 +74,9 @@ schema = strawberry.Schema(query=Query)
 Code above generates following schema.
 
 ```schema
+"""
+A tasty treat
+"""
 type Fruit {
   id: ID!
   name: String!
@@ -80,6 +85,9 @@ type Fruit {
 
 type Color {
   id: ID!
+  """
+  field description
+  """
   name: String!
   fruits: [Fruit!]
 }
--- a/README.md
+++ b/README.md
@@ -35,8 +35,8 @@ class Fruit(models.Model):
 
 class Color(models.Model):
     name = models.CharField(
-      max_length=20,
-      help_text="field description",
+        max_length=20,
+        help_text="field description",
     )
 ```
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -32,8 +32,8 @@ class Fruit(models.Model):
 
 class Color(models.Model):
     name = models.CharField(
-      max_length=20,
-      help_text="field description",
+        max_length=20,
+        help_text="field description",
     )
 ```
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -31,8 +31,10 @@ class Fruit(models.Model):
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20,
-           help_text="field description")
+    name = models.CharField(
+      max_length=20,
+      help_text="field description",
+    )
 ```
 
 ```python
--- a/docs/index.md
+++ b/docs/index.md
@@ -76,6 +76,14 @@ class Query:
 schema = strawberry.Schema(query=Query)
 ```
 
+```python
+# settings.py
+STRAWBERRY_DJANGO = {
+    "FIELD_DESCRIPTION_FROM_HELP_TEXT": True,
+    "TYPE_DESCRIPTION_FROM_MODEL_DOCSTRING": True,
+}
+```
+
 Code above generates following schema.
 
 ```schema
--- a/docs/index.md
+++ b/docs/index.md
@@ -25,12 +25,14 @@ Strawberry integration with Django.
 from django.db import models
 
 class Fruit(models.Model):
+    """A tasty treat"""
     name = models.CharField(max_length=20)
     color = models.ForeignKey('Color', blank=True, null=True,
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20)
+    name = models.CharField(max_length=20,
+           help_text="field description")
 ```
 
 ```python
@@ -69,6 +71,9 @@ schema = strawberry.Schema(query=Query)
 Code above generates following schema.
 
 ```schema
+"""
+A tasty treat
+"""
 type Fruit {
   id: ID!
   name: String!
@@ -77,6 +82,9 @@ type Fruit {
 
 type Color {
   id: ID!
+  """
+  field description
+  """
   name: String!
   fruits: [Fruit!]
 }
--- a/strawberry_django/settings.py
+++ b/strawberry_django/settings.py
@@ -1,0 +1,41 @@
+"""
+Code for interacting with Django settings.
+"""
+from typing import TypedDict
+
+from django.conf import settings
+
+
+class StrawberryDjangoSettings(TypedDict):
+    """
+    Dictionary defining the shape `settings.STRAWBERRY_DJANGO` should have.
+
+    All settings are optional and have defaults as described in their docstrings and
+    defined in `DEFAULT_DJANGO_SETTINGS`.
+    """
+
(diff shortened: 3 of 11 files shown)


strawberry_django/type.py
```diff
@@ -178,7 +184,10 @@ def process_type(
     if not hasattr(cls, "is_type_of"):
         cls.is_type_of = lambda obj, _info: isinstance(obj, (cls, model))
 
-    strawberry.type(cls, **kwargs)
+    # Get type description from either kwargs, or the model's docstring
+    description = kwargs.pop("description", cleandoc(model.__doc__))
```
bellini666, line 188: Just to make sure we don't add an empty string here

```suggestion
    description = kwargs.pop("description", cleandoc(model.__doc__) if model.__doc__ else None)
```

README.md
```diff
@@ -28,12 +28,14 @@ Full documentation is available under [docs](https://strawberry-graphql.github.i
...
 class Fruit(models.Model):
+    """A tasty treat"""
     name = models.CharField(max_length=20)
     color = models.ForeignKey('Color', blank=True, null=True,
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20)
+    name = models.CharField(max_length=20,
+           help_text="field description")
```
bellini666, line 38:
```suggestion
    name = models.CharField(
        max_length=20,
        help_text="field description",
    )
```

docs/index.md
```diff
@@ -25,12 +25,14 @@ Strawberry integration with Django.
...
 class Fruit(models.Model):
+    """A tasty treat"""
     name = models.CharField(max_length=20)
     color = models.ForeignKey('Color', blank=True, null=True,
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20)
+    name = models.CharField(max_length=20,
+           help_text="field description")
```
bellini666, line 35:
```suggestion
    name = models.CharField(
        max_length=20,
        help_text="field description",
    )
```

bellini666 (changes requested):
Some suggestions.

Either way, what do you think of this @patrick91 ?

strawberry_django/type.py
noelleleigh: Here's my implementation: https://github.com/strawberry-graphql/strawberry-graphql-django/pull/160/commits/d52cfc7200aa7ed911f416299aa0b3a0f2535c89

README.md
```diff
@@ -28,12 +28,16 @@ Full documentation is available under [docs](https://strawberry-graphql.github.i
...
+    """A tasty treat"""
     name = models.CharField(max_length=20)
     color = models.ForeignKey('Color', blank=True, null=True,
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20)
+    name = models.CharField(
+      max_length=20,
+      help_text="field description",
```
bellini666, line 39: Maybe this was my suggestion, but it ended up with 2 spaces which is not usual for python =P

```suggestion
        max_length=20,
        help_text="field description",
```

docs/index.md
```diff
@@ -25,12 +25,16 @@ Strawberry integration with Django.
...
+    """A tasty treat"""
     name = models.CharField(max_length=20)
     color = models.ForeignKey('Color', blank=True, null=True,
             related_name='fruits', on_delete=models.CASCADE)
 
 class Color(models.Model):
-    name = models.CharField(max_length=20)
+    name = models.CharField(
+      max_length=20,
+      help_text="field description",
```
bellini666, line 36: Ditto

```suggestion
        max_length=20,
        help_text="field description",
```

bellini666 (changes requested):

README.md
noelleleigh: Ah yep, you're right, I don't know why I missed that inconsistency!

bellini666 (approved):
This looks fine by me! :)

I'll just ask a second opinion from @patrick91

patrick91 (commented):
This looks great to me! I'd only make this configurable (which is what we will be doing in strawberry core too)

bellini666:
@patrick91 any suggestions on where to add configurations for @noelleleigh to implement? Or maybe we should do that first in strawberry so that we can use it here?

patrick91:
> @patrick91 any suggestions on where to add configurations for @noelleleigh to implement? Or maybe we should do that first in strawberry so that we can use it here?

in Strawberry core we have a config parameter on the schema, we could extend that maybe?

or use the common django way of defining settings for Strawberry django?

in core we use config on the schema so we can have multiple schemas with different configurations

noelleleigh:
> in Strawberry core we have a config parameter on the schema, we could extend that maybe?
> 
> or use the common django way of defining settings for Strawberry django?
> 
> in core we use config on the schema so we can have multiple schemas with different configurations

Could you point me to the Strawberry core code for this? Also I don't have a preference for how to configure this, so give me your desired strategy and I'll implement it.

patrick91:
> > in Strawberry core we have a config parameter on the schema, we could extend that maybe?
> > or use the common django way of defining settings for Strawberry django?
> > in core we use config on the schema so we can have multiple schemas with different configurations
> 
> Could you point me to the Strawberry core code for this? Also I don't have a preference for how to configure this, so give me your desired strategy and I'll implement it.

This is where we set the schema config: https://github.com/strawberry-graphql/strawberry/blob/9b3be99a765ae2fdb95c9802507e5600351c7006/strawberry/schema/schema.py#L68

And it is used in the schema converter, it might be more difficult to use in the context of strawberry django though 🤔 


There's an in-flight PR for this feature in core: https://github.com/strawberry-graphql/strawberry/pull/1645 I'm mentioning it because I don't know if it would conflict with what you're doing, I'd love to merge the PR once it is slimmed down a bit :)

patrick91:
> > This is where we set the schema config: [strawberry-graphql/strawberry@`9b3be99`/strawberry/schema/schema.py#L68](https://github.com/strawberry-graphql/strawberry/blob/9b3be99a765ae2fdb95c9802507e5600351c7006/strawberry/schema/schema.py#L68)
> > And it is used in the schema converter, it might be more difficult to use in the context of strawberry django though 🤔
> 
> I've poked around a bit and it doesn't appear that any existing code in strawberry-graphql-django has either the schema or its config in scope. It's not clear to me how my code in the `process_type` function could even reference the schema config at all. Am I missing something?

No, you're exactly right, this is where the complexity is going to be, I don't have a good suggestion at the moment for handling this. Unfortunately having the configuration at the schema level made some implementations quite complex and it will make this one complex too.

We could try to figure out something for this, maybe like we did for the name conversion part: https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/schema/name_converter.py

or we could decide that for django we can use a "global" configuration, like other framework do. What do you think? /cc @bellini666

noelleleigh:
A Django setting would also be a simpler option

noelleleigh:
> > A Django setting would also be a simpler option
> 
> yeah, let's go with that, we can update it later if needed 😊

Should it be a single setting that controls both type and field description inclusion? Or two settings, one for each?

```python
STRAWBERRY_DJANGO_INHERIT_DESCRIPTIONS = True
```

vs.

```python
STRAWBERRY_DJANGO_INHERIT_MODEL_DESCRIPTIONS = True
STRAWBERRY_DJANGO_INHERIT_FIELD_DESCRIPTIONS = True
```

patrick91:
> > > A Django setting would also be a simpler option
> > 
> > 
> > yeah, let's go with that, we can update it later if needed 😊
> 
> Should it be a single setting that controls both type and field description inclusion? Or two settings, one for each?
> 
> ```python
> STRAWBERRY_DJANGO_INHERIT_DESCRIPTIONS = True
> ```
> 
> vs.
> 
> ```python
> STRAWBERRY_DJANGO_INHERIT_MODEL_DESCRIPTIONS = True
> STRAWBERRY_DJANGO_INHERIT_FIELD_DESCRIPTIONS = True
> ```
> 
> And what should their default value(s) be?

I'd do this personally:

```python
STRAWBERRY_DJANGO = {
    "FIELD_DESCRIPTION_FROM_HELP_TEXT": False,
    "TYPE_DESCRIPTION_FROM_MODEL_DOCSTRING": False,
}
```

and disabled by default for backwards compatibility and to also prevent potential information leaking 😊

bellini666:
> or we could decide that for django we can use a "global" configuration, like other framework do. What do you think? /cc @bellini666

I like the idea! :)

We still need to think how to do that configuration for strawberry in the future though, but for django I think this is perfect.

noelleleigh:
Alright, I've added code for new Django settings, added tests, and added documentation.

noelleleigh:
How do those settings changes look to y'all?

strawberry_django/settings.py
```diff
@@ -0,0 +1,41 @@
+"""
+Code for interacting with Django settings.
+"""
+from typing import TypedDict
+
+from django.conf import settings
+
+
+class StrawberryDjangoSettings(TypedDict):
```
patrick91, line 9: Nice!

patrick91 (approved):
Looks pretty good to me! I'll leave it to @bellini666 to review and merge 😊

bellini666 (approved):
It's fine by me too! :)

PR #161, opened 2022-08-02
fix: Raise NotImplementedError on unknown Django fields

noelleleigh:
## Description

Ensures that any Django model field that is not recognized by strawberry_django will cause `NotImplementedError` to be raised with a helpful message

This improves the behavior for models using [GeoDjango Spatial Fields](https://docs.djangoproject.com/en/3.2/ref/contrib/gis/model-api/#spatial-field-types), which previously would raise a cryptic `KeyError`.

Added tests:

- `tests/fields/test_types.py::test_notimplemented`

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


bellini666 (approved):
Awesome, thank you! :)

PR #162, opened 2022-08-02
Add many type hints

noelleleigh:
## Description

In the spirit of Strawberry's love of [type hints](https://docs.python.org/3/library/typing.html), I added a bunch of annotations and corrected existing ones after spending time in the code for #160. These changes were made with the intent of improving the feedback given by Pylance while exploring the "types" and "fields" code areas, and to provide implicit documentation of how the functions operate.

The details can be found in the commit messages, which I tried to make as atomic as possible for clarity.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -2,6 +2,7 @@ import dataclasses
 from typing import Any, Dict, Optional, Type
 
 import django
+import django.db.models
 import strawberry
 from strawberry import UNSET
 from strawberry.annotation import StrawberryAnnotation
@@ -119,7 +120,7 @@ def get_fields(django_type: "StrawberryDjangoType"):
 @dataclasses.dataclass
 class StrawberryDjangoType:
     origin: Any
-    model: Any
+    model: Type[django.db.models.Model]
     is_input: bool
     is_partial: bool
     is_filter: bool
@@ -131,7 +132,7 @@ class StrawberryDjangoType:
 
 def process_type(
     cls,
-    model,
+    model: Type[django.db.models.Model],
     *,
     filters=UNSET,
     pagination=UNSET,
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -1,5 +1,5 @@
 import dataclasses
-from typing import Any, Optional, Type
+from typing import Any, Dict, Optional, Type
 
 import django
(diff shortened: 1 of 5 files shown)


strawberry_django/type.py
```diff
@@ -119,19 +120,19 @@ def get_fields(django_type):
 @dataclasses.dataclass
 class StrawberryDjangoType:
     origin: Any
-    model: Any
+    model: Type[django.db.models.Model]
```
bellini666, line 123: In `strawberry-django-plus`, which I'm planning on merging here, I made this a generic type to be able to specialize this: https://github.com/blb-ventures/strawberry-django-plus/blob/main/strawberry_django_plus/type.py#L338

What do you think about doing the same here?

bellini666 (approved):
Thank you very much for this.

This is one step closer for resolving this issue: https://github.com/strawberry-graphql/strawberry-graphql-django/issues/136

I've made a small suggestion, but it is not necessary to merge this.

strawberry_django/type.py
noelleleigh: Help me understand, for this use case, what's the difference between  `Type[django.db.models.Model]` and `Type[_M]`? They're both tied to the `Model` class, right?

bellini666: They are both bound to model, the case there is that I transformed `StrawberryDjangoType` into a generic class which can be specialized. e.g. `foo = StrawberryDjangoType[FooBarModel]` would make accessing `foo.model` to be `FooBarModel` instead of simply `Model`

bellini666: https://docs.python.org/3/library/typing.html#typing.Generic

noelleleigh: Okay, I'll make that change.

noelleleigh: See https://github.com/strawberry-graphql/strawberry-graphql-django/pull/162/commits/855d92d48af4f58a695657eda1c1c1775cd08e86

bellini666 (approved):

strawberry_django/type.py
bellini666: Thank you! :)

PR #183, opened 2022-09-28
Fix mutations and filtering for when using strawberry-graphql >=0.132.1

jkimbo:
## Description

The [v0.132.1](https://github.com/strawberry-graphql/strawberry/releases/tag/0.132.1) release of strawberry-graphql broke mutations and filtering because it introduced the concept of "basic fields" which skip calling resolvers in certain circumstances. Strawberry-Django doesn't use `base_resolver` for it's custom fields but it does define custom resolvers without implementing the `is_basic_field` property and so the custom resolvers were not longer being called. This PR makes sure that the custom Strawberry Django fields are never treated as "basic" by Strawberry.

It also fixes some tests that were instantiated without keyword arguments. Keyword arguments were required from [v0.128.0](https://github.com/strawberry-graphql/strawberry/releases/tag/0.132.1) of Strawberry.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* fixes https://github.com/strawberry-graphql/strawberry-graphql-django/issues/182

## Checklist

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -57,7 +57,7 @@ unicode_backport = ["unicodedata2"]
 name = "click"
 version = "8.1.3"
 description = "Composable command line interface toolkit"
-category = "main"
+category = "dev"
 optional = false
 python-versions = ">=3.7"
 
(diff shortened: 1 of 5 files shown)


g-as:
That's great, thanks!

illia-v (approved):
@jkimbo thanks!

This fixes tests in a project that strawberry-graphql 0.132.1 could break.

bellini666:
Hey @jkimbo . PR looks great! :)
Just need to resolve the conflicts in poetry.lock.

Also, since this will fix the issue from v0.132.1, maybe we should alson unpin the restriction in here?

jkimbo:
> Also, since this will fix the issue from v0.132.1, maybe we should alson unpin the restriction in here?

Good point, I've removed the restriction now.

bellini666 (approved):

PR #185, opened 2022-10-01
Broaden is_strawberry_django_field to support custom field classes

benhowes:
## Description
When:
1. Using a custom field class in a type
2. It's processed by `type.get_field` is called on an instance of a custom field
3. The custom field class's `from_field` wants to replicate similar functionality to `StrawberryDjangoField.from_field`

It will be useful to have the `is_strawberry_django_field` use the `StrawberryDjangoFieldBase` class as it's means of detecting a strawberry-django field class since that will also be used in a custom class.

I don't believe this change will affect any existing usecase

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry_django/utils.py
+++ b/strawberry_django/utils.py
@@ -37,9 +37,9 @@ def is_strawberry_field(obj):
 
 
 def is_strawberry_django_field(obj):
-    from strawberry_django.fields.field import StrawberryDjangoField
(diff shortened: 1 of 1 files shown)


bellini666 (approved):
LGTM :)

PR #191, opened 2022-10-10
Allow passing info in filters

kwongtn:
## Description

This patch allows passing the info object to filters, catering for use cases like filtering with respect to the logged in user (which, its relative cookie will be passed via the input object).

At the same time, it addresses existing codebase and not breaking them by performing analysis on the filter function provided. If the `info` parameter exists, then it is passed. Else not.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes as per discussion here: https://discord.com/channels/689806334337482765/1025667620407619614 
* Personally decided against rewriting a 'resolvers' support code as not only I think it will potentially break more codebase, it will also increase complexity

## Checklist

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,3 +1,4 @@
+import inspect
 from enum import Enum
 from typing import Generic, List, Optional, TypeVar
 
@@ -118,7 +119,7 @@ def build_filter_kwargs(filters):
     return filter_kwargs, filter_methods
 
 
-def apply(filters, queryset, pk=UNSET):
+def apply(filters, queryset, info=UNSET, pk=UNSET):
     if pk is not UNSET:
         queryset = queryset.filter(pk=pk)
 
@@ -137,7 +138,16 @@ def apply(filters, queryset, pk=UNSET):
     filter_kwargs, filter_methods = build_filter_kwargs(filters)
     queryset = queryset.filter(**filter_kwargs)
     for filter_method in filter_methods:
-        queryset = filter_method(queryset=queryset)
+        argspec = inspect.getfullargspec(filter_method)
+
+        if "info" in getattr(argspec, "args", []) or "info" in getattr(
+            argspec, "kwargs", []
+        ):
+            queryset = filter_method(queryset=queryset, info=info)
+
+        else:
+            queryset = filter_method(queryset=queryset)
+
     return queryset
 
 
@@ -169,4 +179,4 @@ class StrawberryDjangoFieldFilters:
 
     def get_queryset(self, queryset, info, pk=UNSET, filters=UNSET, **kwargs):
         queryset = super().get_queryset(queryset, info, **kwargs)
-        return apply(filters, queryset, pk)
+        return apply(filters, queryset, info, pk)
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,5 +1,7 @@
+import functools
 import inspect
 from enum import Enum
+from types import FunctionType
 from typing import Generic, List, Optional, TypeVar
 
 import strawberry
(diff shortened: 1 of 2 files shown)


strawberry_django/filters.py
```diff
@@ -137,7 +138,16 @@ def apply(filters, queryset, pk=UNSET):
...
-        queryset = filter_method(queryset=queryset)
+        argspec = inspect.getfullargspec(filter_method)
+
+        if "info" in getattr(argspec, "args", []) or "info" in getattr(
+            argspec, "kwargs", []
+        ):
+            queryset = filter_method(queryset=queryset, info=info)
+
+        else:
+            queryset = filter_method(queryset=queryset)
```
bellini666, line 149: Just worried regarding the performance here.

Maybe add a utility decorated with `@functools.cache` to retrieve the filter method?

tests/filters/test_filters.py
```diff
@@ -172,6 +172,34 @@ def fruits(filters: FruitFilter) -> List[Fruit]:
...
+    from strawberry.types.info import Info
+
+    @strawberry_django.filters.filter(models.Fruit, lookups=True)
+    class FruitFilterWithInfo:
+        id: auto
+        name: auto
+        custom_field: bool
+
+        def filter_custom_field(self, queryset, info: Info):
+            # Test here is to prove that info can be passed properly
```
bellini666, line 185: Just to make sure it comes as `Info` properly and not with some other value

```suggestion
            # Test here is to prove that info can be passed properly
            assert isinstance(info, Info)
```

bellini666 (changes requested):

strawberry_django/filters.py
kwongtn: Added accordingly -- not sure if this is what you meant tho 😅
152cf68ac27ffb46bf54d489aa7c5b299ea8a382

strawberry_django/filters.py
```diff
@@ -118,7 +121,19 @@ def build_filter_kwargs(filters):
...
+@functools.cache()
+def function_allow_passing_info(filter_method: FunctionType) -> bool:
+    argspec = inspect.getfullargspec(filter_method)
+
+    if "info" in getattr(argspec, "args", []) or "info" in getattr(
+        argspec, "kwargs", []
+    ):
+        return True
+    else:
+        return False
```
bellini666, line 135: :)

```suggestion
    return "info" in getattr(argspec, "args", []) or "info" in getattr(
        argspec, "kwargs", []
    )
```

strawberry_django/filters.py
bellini666: Exactly that 👍🏼

bellini666 (changes requested):
Just a small improvement and this should be good to merge

strawberry_django/filters.py
kwongtn: Holy shite never thought of that -- thanks! 😁

strawberry_django/filters.py
```diff
@@ -119,6 +121,18 @@ def build_filter_kwargs(filters):
     return filter_kwargs, filter_methods
 
 
+@functools.cache()
```
blueyed, line 126: I have not finished investigating, but am seeing instances running out of memory due to upgrading strawberry-graphql-django from 0.5.4 to 0.7, with this change being added in 0.6..

Since this is basically an unbound `lru_cache`, I suggest adding some `max_size` here at least - but have no idea currently why it was added in the first place etc..

Will try with 0.6 to harden / narrow down the changes that might cause the leaking / increasing memory usage.

bellini666: Do you have too much filter methods in your schema?

I have nothing against adding a `max_size` here. Maybe a `max_size` of ~200 is enough?

blueyed: There shouldn't be too many, but it looks more like e.g. lambdas being passed in there or something similar.
Maybe due to https://github.com/blb-ventures/strawberry-django-plus?

Can investigate more next week.. (it is also an issue with 0.6, making this change more likely to cause it)

bellini666: Just investigated here, it was due to bound methods being different every time (for obvious reasons). Just fixed it here and added a max_size to avoid any issue alike in the future: https://github.com/strawberry-graphql/strawberry-graphql-django/commit/dd09948f35849ae442ab75fc4fa5c171b8cdfbb1

PR #193, opened 2022-10-11
Implementing filter, order and pagination in `StrawberryDjangoField` super classes

ManiacMaxo:
## Description

Changed base classes for `StrawberryDjangoField` to allow overloading filter, order and pagination of queryset without too much hassle. If someone were to want to change how filtering works and supply a custom filter type - they'd have to subclass `StrawberryDjangoField` and change the `get_queryset` method which already calls `StrawberryDjangoFieldFilters.get_queryset` which implements the filtering in a unrelated function that cannot be overrwritten - with this change they'd just have to overload `_apply_filters`.

Also added type hints to the methods.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -169,7 +169,7 @@ class StrawberryDjangoFieldFilters:
             return type_._django_type.filters
         return None
 
-    def _apply_filters(
+    def apply_filters(
         self, queryset: QuerySet, filters: Type = UNSET, pk=UNSET
     ) -> QuerySet:
         return apply(filters, queryset, pk)
@@ -178,4 +178,4 @@ class StrawberryDjangoFieldFilters:
         self, queryset: QuerySet, info: Info, pk=UNSET, filters: Type = UNSET, **kwargs
     ):
         queryset = super().get_queryset(queryset, info, **kwargs)
-        return self._apply_filters(queryset, filters, pk)
+        return self.apply_filters(queryset, filters, pk)
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,10 +1,12 @@
 from enum import Enum
-from typing import Generic, List, Optional, TypeVar
+from typing import Generic, List, Optional, Type, TypeVar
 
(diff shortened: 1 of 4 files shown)


strawberry_django/filters.py
```diff
@@ -167,6 +169,13 @@ def get_filters(self):
             return type_._django_type.filters
         return None
 
-    def get_queryset(self, queryset, info, pk=UNSET, filters=UNSET, **kwargs):
-        queryset = super().get_queryset(queryset, info, **kwargs)
+    def _apply_filters(
```
bellini666, line 172: Since this is here for overriding, I would suggest setting it as a public method to avoid someone in the future thinking it can be removed without breaking anything

strawberry_django/ordering.py
```diff
@@ -77,6 +80,11 @@ def get_order(self):
             return type_._django_type.order
         return None
 
-    def get_queryset(self, queryset, info, order=UNSET, **kwargs):
-        queryset = super().get_queryset(queryset, info, **kwargs)
+    def _apply_order(self, queryset: QuerySet, order) -> QuerySet:
```
bellini666, line 83: Ditto

strawberry_django/pagination.py
```diff
@@ -46,6 +48,11 @@ def get_pagination(self):
             return type_._django_type.pagination
         return None
 
-    def get_queryset(self, queryset, info, pagination=UNSET, **kwargs):
+    def _apply_pagination(self, queryset: QuerySet, pagination):
```
bellini666, line 51: Ditto

bellini666 (changes requested):
PR seems fine overall! :) Just a minor suggestion and it should be good to merge

strawberry_django/filters.py
ManiacMaxo: I was thinking the same but wasn't sure

bellini666 (approved):

PR #206, opened 2022-11-04
Change how we set the default annotation

### LINKED ISSUE #205: Field annotation auto breaks when upgrading to Strawberry 0.139.0
## Describe the Bug
After upgrading to Strawberry 0.139.0 any `strawberry.django.type` classes that use the `auto` type will fail with error. See the recent release: https://github.com/strawberry-graphql/strawberry/releases/tag/0.139.0

**Example**

```
strawberry-graphql==0.139.0
strawberry-graphql-django==0.7.1
```

```python
@strawberry.django.type(models.Service)
class Service:
    id: auto
```

```
TypeError: Service fields cannot be resolved. Unexpected type 'auto'
```

## System Information

- Operating system: macOS Monterey version `12.6`
- Strawberry version (if applicable): `0.139.0`
- `strawberry-graphql` version (if applicable): `0.7.1`

## Additional Context

The Strawberry 0.139.0 release just went out Nov 4 @ 11am ET. It is very new.

patrick91:
This fixes #205 

@bellini666 do you think the changes make sense?

### DIFF
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -180,8 +180,8 @@ def process_type(
             annotation = field.type_annotation.annotation
         elif field.base_resolver and  field.base_resolver.type_annotation:
             annotation = field.base_resolver.type_annotation.annotation
-        else:
-            annotation = field.type
+
+        # TODO: should we raise an error if annotation is None here?
 
         cls_annotations[field.name] = annotation
         setattr(cls, field.name, field)
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -174,13 +174,15 @@ def process_type(
     # update annotations and fields
     cls.__annotations__ = cls_annotations = {}
     for field in fields:
-        annotation = (
-            field.type
-            if field.type_annotation is None
-            else field.type_annotation.annotation
-        )
-        if annotation is None:
-            annotation = StrawberryAnnotation(strawberry.auto)
+        annotation = None
+
+        if field.type_annotation and field.type_annotation.annotation:
+            annotation = field.type_annotation.annotation
+        elif field.base_resolver and  field.base_resolver.type_annotation:
+            annotation = field.base_resolver.type_annotation.annotation
+        else:
+            annotation = field.type
+
         cls_annotations[field.name] = annotation
         setattr(cls, field.name, field)
 
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -178,7 +178,7 @@ def process_type(
 
         if field.type_annotation and field.type_annotation.annotation:
             annotation = field.type_annotation.annotation
-        elif field.base_resolver and  field.base_resolver.type_annotation:
+        elif field.base_resolver and field.base_resolver.type_annotation:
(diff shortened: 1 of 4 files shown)


strawberry_django/type.py
```diff
@@ -174,13 +174,15 @@ def process_type(
...
-            else field.type_annotation.annotation
-        )
-        if annotation is None:
-            annotation = StrawberryAnnotation(strawberry.auto)
+        annotation = None
+
+        if field.type_annotation and field.type_annotation.annotation:
+            annotation = field.type_annotation.annotation
+        elif field.base_resolver and field.base_resolver.type_annotation:
+            annotation = field.base_resolver.type_annotation.annotation
```
patrick91, line 182: here we try to get the annotation from the base resolver if we didn't get one before

patrick91:
tests are failing with the older version of strawberry, shall we bump it? or try to find a fix that works with older versions too?

strawberry_django/type.py
jkimbo: @patrick91 Why aren't you using `field.type`?

patrick91: field.type tries to use .resolve and it fails because it is too early: 
CleanShot 2022-11-04 at 18 35 19@2x

strawberry_django/type.py
```diff
@@ -174,13 +174,15 @@ def process_type(
...
+        else:
+            annotation = field.type
```
patrick91, line 184: I think we can remove this too

strawberry_django/type.py
jkimbo: Hmm ok. I think we need a `resolved_type` method (or `unresolved_type`) which encapsulates this logic otherwise we're just spreading it out. It will break if the `graphql_type` PR lands for example.

patrick91: yes totally :D I think we also need find a way to consolidate the logic for getting the field and types with the one we use for the pydantic extension

bellini666: I agree with @jkimbo that we really need a `resolve_type` method =P

bellini666:
@patrick91 

> @bellini666 do you think the changes make sense?

Taking a quick look at it, it does seem to make sense.

> tests are failing with the older version of strawberry, shall we bump it? or try to find a fix that works with older versions too?

I think it makes sense to bump the minimum strawberry version. IMHO there's no reason for anyone to not upgrade to the latest strawberry version other than bugs like this one, since there's no breaking change there which would make the user need to "decide" one way or another of doing things

patrick91:
@bellini666 updated! shall I merge? :)

bellini666:
@patrick91 just merged for you! :)

Will do a new release as well. Should this be 0.7.2 or 0.8? It is a bugfix release, but it bumps the minimum strawberry version, so I'm not sure =P

patrick91:
> @patrick91 just merged for you! :)
> 
> 
> 
> Will do a new release as well. Should this be 0.7.2 or 0.8? It is a bugfix release, but it bumps the minimum strawberry version, so I'm not sure =P

let's go with 0.8 :)

PR #214, opened 2022-11-15
make pk argument required when querying single object

stygmate:
It's just a one line correction to make pk argument required when querying a single django object

edit: and test fot required pk argument

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -182,7 +182,7 @@ class StrawberryDjangoFieldFilters:
             filters = self.get_filters()
             if self.django_model and not self.is_list:
                 if self.is_relation is False:
-                    arguments.append(argument("pk", strawberry.ID))
+                    arguments.append(argument("pk", strawberry.ID, is_optional=False))
             elif filters and filters is not UNSET:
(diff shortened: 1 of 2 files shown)


patrick91:
@stygmate looks good to me, can you update the tests? :)

stygmate:
> @stygmate looks good to me, can you update the tests? :)

Hi the test are not failing because of my modifications. Looks like a error in creation of docker image. Tests are ok my side. But if you want I can add a test to verify that the field is actually required.

patrick91:
> > @stygmate looks good to me, can you update the tests? :)
> 
> Hi the test are not failing because of my modifications. Looks like a error in creation of docker image. Tests are ok my side. But if you want I can add a test to verify that the field is actually required.

yes, let's add that! so we don't get any regression 😊

stygmate:
@patrick91 the ugly pre-commit.ci was annoying me ;) but seems ok now !

patrick91 (approved):
Looks great! I'll leave the final review to @bellini666 😊

bellini666 (approved):

bellini666:
LGTM 👍🏼

PR #237, opened 2023-02-26
chore: migrate from flake8/isort to ruff

bellini666:


strawberry_django/filters.py
```diff
@@ -1,19 +1,26 @@
...
 
 import strawberry
-from django.db.models import QuerySet
 from django.db.models.sql.query import get_field_names_from_opts
 from strawberry import UNSET
-from strawberry.arguments import StrawberryArgument
-from strawberry.types import Info
+
+# FIXME: Import from typing_extensions to support | notation on python older than 3.10
+from typing_extensions import TypeVar
```
bellini666, line 13: That's interesting. It broke with both using `TypeVar` from `typing` and from `typing_extensions` as well (and only for python 3.7/3.8/3.9)

The error happens when strawberry calls `_eval_type`. Maybe there's another approach to handle those when using future annotations?

patrick91: what error do you get?

bellini666: The errors are in the tests, for python 3.8 and 3.9. 3.10 and 3.11 are fine

I thought that importing from `typing_extensions` would solve the issue, but it seems that `_eval_type` can't resolve `Foo | Bar`.

I'm working on some fixes for future references, I'll try to address that as well, specially because in python 3.12 it will be the default behaviour.

bellini666: This PR should fix the issue, and also add full support for future annotations on strawberry: https://github.com/strawberry-graphql/strawberry/pull/2592

bellini666: @patrick91 everything fixed now :)

bellini666:
@patrick91 this PR should be good to me merged

patrick91 (approved):

patrick91 (approved):

PR #239, opened 2023-02-28
feature: backporting django-debug-toolbar from strawberry-django-plus

frleb:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* Fixes https://github.com/strawberry-graphql/strawberry-graphql-django/issues/138

## Checklist


pyproject.toml
```diff
@@ -29,6 +29,7 @@ classifiers = [
 python = ">=3.7,<4.0"
 Django = ">=3.2"
 strawberry-graphql = ">=0.154.0"
+django-debug-toolbar = { version = ">=3.4", optional = true }
```
bellini666, line 32: Probably also need to add a `tool.poetry.extras` section

bellini666 (changes requested):
Awesome! Thank you for doing this

Just made a small code suggestion. Also, we need to make sure that the `strawberry_django` app can be added to `INSTALLED_APPS`, otherwise it won't be able to find the templates. Probably just need to define an `apps.py` file

pyproject.toml
frleb: done !

bellini666 (approved):
LGTM 👍🏼 

Just a question, did you test that this is indeed working? For example by running the example project?

frleb:
> LGTM 👍🏼
> 
> Just a question, did you test that this is indeed working? For example by running the example project?

Good point, I retested by cloning my branch again. I had to force the addition of the debug_toolbar in the poetry of the demo project

blueyed:
It should also copy over/adjust the documentation then: https://blb-ventures.github.io/strawberry-django-plus/debug-toolbar/ (https://github.com/blb-ventures/strawberry-django-plus/blob/90fa11f4c5254a1fba0db60a9a4362568b11cbcc/docs/debug-toolbar.md).

bellini666 (approved):
LGTM 👍🏼

PR #246, opened 2023-04-02
refactor: do not insert `pk` arguments inside non root fields

### LINKED ISSUE #215: GraphQLError with property field in v0.8.2
With v0.8.2 (due to https://github.com/strawberry-graphql/strawberry-graphql-django/commit/d2ad144f51a3fd66228655ee2ae9b2718b78e377, added via https://github.com/strawberry-graphql/strawberry-graphql-django/pull/214) I am seeing an error with a field that is a property on the model:

> GraphQLError("Field 'fooProp' argument 'pk' of type 'ID!' is required, but it was not provided.")

```python
@django_type(models.Bar, filters=BarFilter, pagination=True)
class Bar:
    foo_prop: Foo | None = field(
        only=["foo__status"],
    )
```

/cc @benjaminderei @stygmate

### LINKED ISSUE #235: OneToOne should work like `ForeignKey`
I was doing a quick test with a friend and we had a model similar to this:

```python
class User(models.Model):
    ...

class Profile(models.Model):
    user = OneToOne(User)
```

and this was generating a query like this:

```graphql

query {
    user(id: 1) {
       profile(id: "ABC") { ... }
    }
}
```

while it should generate a query like this:

```graphql

query {
    user(id: 1) {
       profile { ... }
    }
}
```

without the id on the profile field

### LINKED ISSUE #245: One-to-One Models results in errors
## Describe the Bug

I try to use on-to-one models in strawberry. However, I face this issue: Its might related to this issue mentioned here https://github.com/strawberry-graphql/strawberry-graphql-django/issues/235, but I create a new issue in case the issue is different from that.

models.py
```python
from django.contrib.auth.models import User
class User(models.Model):
    user = models.OneToOneField(User, on_delete=models.CASCADE)
    birthday = models.DateField()
    phone_number = PhoneNumberField(blank=True)
```

schema.py
```python
@strawberry.django.type(DJangoUser)
class DJangoUser:
    username: auto
    first_name: auto
    last_name: auto
    email: auto

@strawberry.django.type(models.User)
class User:
    id: strawberry.ID
    birthday: date
    phone_number: str
    django_user: DJangoUser
```

Error:
```cmd
{
	"data": null,
	"errors": [
		{
			"message": "Field 'djangoUser' argument 'pk' of type 'ID!' is required, but it was not provided.",
			"locations": [
				{
					"line": 4,
					"column": 3
				}
			]
		}
	]
}
```

The details of the code can be look at here https://github.com/Skyquek/Trip-Bill-Management-System/pull/6.

bellini666:
Instead of checking if the field is a relation or not, just check if
it is defined in a root type (e.g. a `Query`). If it is not, do not
insert the pk argument for it.

This fixes a lot of issues like:

1) Defining a field with a name different from the relation
2) Defining a field for a function or property inside the model
3) etc

Fix #215
Fix #235
Fix #245

### DIFF
--- a/tests/filters/test_filters.py
+++ b/tests/filters/test_filters.py
@@ -250,6 +250,7 @@ def test_enum(query, fruits):
     ]
 
 
+@pytest.mark.django_db(transaction=True)
 def test_pk_inserted_for_root_field_only():
     @strawberry_django.type(models.Group)
     class GroupType(models.Group):
@@ -268,7 +269,6 @@ def test_pk_inserted_for_root_field_only():
 
     schema = strawberry.Schema(query=Query)
 
-    __import__("pprint").pprint(str(schema))
     assert (
         textwrap.dedent(str(schema))
         == textwrap.dedent(
@@ -290,3 +290,35 @@ def test_pk_inserted_for_root_field_only():
     """
         ).strip()
     )
+
+    group = models.Group.objects.create(name="Some Group")
+    user = models.User.objects.create(name="Some User", group=group)
+
+    res = schema.execute_sync(
+        """\
+      query GetUser ($pk: ID!) {
+        user(pk: $pk) {
+          name
+          group {
+            name
+          }
+          getGroup {
+            name
+          }
+          groupProp {
+            name
+          }
+        }
+      }
+    """,
+        variable_values={"pk": user.pk},
+    )
+    assert res.errors is None
+    assert res.data == {
+        "user": {
+            "name": "Some User",
+            "group": {"name": "Some Group"},
+            "getGroup": {"name": "Some Group"},
+            "groupProp": {"name": "Some Group"},
+        }
+    }
--- a/tests/filters/test_filters.py
+++ b/tests/filters/test_filters.py
@@ -1,3 +1,4 @@
+import textwrap
 from enum import Enum
 from typing import List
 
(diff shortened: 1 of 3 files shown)


tests/filters/test_filters.py
```diff
@@ -247,3 +248,45 @@ def test_enum(query, fruits):
...
+        get_group: GroupType
+        group_prop: GroupType
+
+    @strawberry.type
+    class Query:
+        user: UserType = strawberry_django.field()
+
+    schema = strawberry.Schema(query=Query)
+
+    __import__("pprint").pprint(str(schema))
```
patrick91, line 271: does this need to stay here? :)

tests/filters/test_filters.py
```diff
@@ -247,3 +248,45 @@ def test_enum(query, fruits):
...
+      }
+
+      type UserType {
+        name: String!
+        group: GroupType
+        getGroup: GroupType!
+        groupProp: GroupType!
+      }
+    """
+        ).strip()
```
patrick91, line 291: should we also test the output when running a query?

tests/filters/test_filters.py
bellini666: loool, I used it to print the schema so that I could copy and paste and forgot it there

bellini666: Done

tests/filters/test_filters.py
bellini666: Done! :)

patrick91 (approved):

g-as:
There's a branch that has been removed here:
```python
            if self.django_model and not self.is_list:
                if self.is_relation is False:
                    arguments.append(argument("pk", strawberry.ID, is_optional=False))
            elif filters and filters is not UNSET:
                arguments.append(argument("filters", filters))
```
It becomes more apparent if we flatten the conditions:
```python
            if self.django_model and not self.is_list and self.is_relation is False:
                arguments.append(argument("pk", strawberry.ID, is_optional=False))
            elif self.django_model and not self.is_list:
                pass
            elif filters and filters is not UNSET:
                arguments.append(argument("filters", filters))
```
The side effect of that is that `filters` is appended on foreign keys.

For exemple, in the newly added test, if you add:
```python
    @strawberry_django.filters.filter(models.Group)
    class GroupFilter:
        name: str

    @strawberry_django.type(models.Group, filters=GroupFilter)
    class GroupType:
        name: strawberry.auto
```
you end up with:
```
type UserType {
  name: String!
  group(filters: GroupFilter): GroupType
  getGroup(filters: GroupFilter): GroupType!
  groupProp(filters: GroupFilter): GroupType!
}
```

bellini666:
@g-as thanks for reporting this. I didn't notice I removed one branch by mistake... I prefer flatten conditions for this reason, it is a lot better to visualize the branches than nested ifs =P

I released a fix for that :)

PR #250, opened 2023-04-19
Add kwargs to the documentation about get_queryset

cdroege:
If you have a type that uses filters or the pagination feature, the `get_queryset` method gets called with more parameters and the previously documented example would fail with an error message.

For example, if you add the `filters` parameter to a type, there will be a `filters` parameters in `kwargs`.


patrick91 (approved):

patrick91:
thanks @cdroege!

Issue #258, opened 2023-06-10
relay: problems with _id and _id_attr


devkral:
The novel relay approach causes the  _id and _id_attr to appear in strawberry-graphql-django

The problem is the get_fields function which doesn't filter fields as dataclass.fields do.

It would be maybe better to use the same approach as in strawberry as it caused already some problems with Private

bellini666:
I'm currently adjusting that for strawberry-django-plus (not only to fix that but also to use the new relay infrastructure instead of its own implementation) and will update this to use the same solution :)

devkral:
finished the PR ;). Now it should work

bellini666:
Closing this as fixed on strawberry core

PR #260, opened 2023-06-13
add strawberry.relay tests, fix compatibility with relay, fix other issues

devkral:
Fix the compatibility with strawberry.relay

<del>The part which keeps the schema validation crashing is:</del>

<del>the graphql name of the `_id` field is ignored. Instead ´_id´ is transformed to `Id` even with ``auto_camel_case` off, This causes the interface validation to crash.</del>

<del>Seemingly strawberry-graphql-django just ignores graphql names and need some updates </del>

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #258

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+-   Fix compatibility with relay, ClassVars and method fields
+-   Fix bug where django_name is incorrectly set
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,3 +2,4 @@ Release type: patch
 
 -   Fix compatibility with relay, ClassVars and method fields
 -   Fix bug where django_name is incorrectly set
+-   Add tests for relay resolve_id/NodeID
--- a/strawberry_django/utils.py
+++ b/strawberry_django/utils.py
@@ -2,7 +2,7 @@ import asyncio
 import dataclasses
 import sys
 import warnings
-from typing import Dict
+from typing import ClassVar, Dict
 
 from django.db import models
 from strawberry.annotation import StrawberryAnnotation
@@ -85,15 +85,29 @@ def is_similar_django_type(a, b):
     return True
 
 
+# dirty workaround, but similar to the one of dataclasses
+# it would be better to use dataclasses.fields
+def _is_not_classvar(annotation, namespace):
+    if isinstance(annotation, str):
+        if annotation.startswith("ClassVar["):
+            annotation = namespace["ClassVar"]
+        elif annotation.startswith("typing.ClassVar["):
+            annotation = namespace["typing"].ClassVar
+    return annotation is not ClassVar
+
+
 def get_annotations(cls):
     annotations: Dict[str, StrawberryAnnotation] = {}
-    namespace = sys.modules[cls.__module__].__dict__
     for c in reversed(cls.__mro__):
+        namespace = sys.modules[c.__module__].__dict__
         if "__annotations__" in c.__dict__:
             annotations.update(
                 {
                     k: StrawberryAnnotation(v, namespace=namespace)
-                    for k, v in c.__annotations__.items()
+                    for k, v in filter(
+                        lambda item: _is_not_classvar(item[1], namespace),
+                        c.__annotations__.items(),
+                    )
                 },
             )
     return annotations
--- a/strawberry_django/utils.py
+++ b/strawberry_django/utils.py
@@ -87,13 +87,13 @@ def is_similar_django_type(a, b):
 
 # dirty workaround, but similar to the one of dataclasses
 # it would be better to use dataclasses.fields
-def _is_not_classvar(annotation, namespace):
+def _is_classvar(annotation, namespace):
     if isinstance(annotation, str):
         if annotation.startswith("ClassVar["):
             annotation = namespace["ClassVar"]
         elif annotation.startswith("typing.ClassVar["):
             annotation = namespace["typing"].ClassVar
-    return annotation is not ClassVar
+    return annotation is ClassVar
 
 
 def get_annotations(cls):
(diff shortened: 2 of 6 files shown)


devkral:
And the classvar stuff is a little hacky but seemingly there is no better solution

strawberry_django/utils.py
```diff
@@ -85,15 +85,29 @@ def is_similar_django_type(a, b):
     return True
 
 
+# dirty workaround, but similar to the one of dataclasses
+# it would be better to use dataclasses.fields
+def _is_not_classvar(annotation, namespace):
```
bellini666, line 90: I usually don't like negative checks because it can generate code like `if not _is_not_classvar(...)` to check of it is.

I think it is better to define this as `_is_classvar` and check for `not _is_classvar` below

strawberry_django/utils.py
```diff
@@ -85,15 +85,29 @@ def is_similar_django_type(a, b):
...
+    if isinstance(annotation, str):
+        if annotation.startswith("ClassVar["):
+            annotation = namespace["ClassVar"]
+        elif annotation.startswith("typing.ClassVar["):
+            annotation = namespace["typing"].ClassVar
```
bellini666, line 95: Although this solution works in general cases, it has some issues:

1) `ClassVar` can be imported from other modules like `typing_extensions`, the `typing` module can be aliased to something else, or the user can have a "compat" module from typing to import it from there

2) The `ClassVar` can be inside a `if TYPE_CHECKING`, specially when using future annotations, and thus would not be on namespace

Having said that, this lib already has a lot of those issues, specially with future annotations, based on the fact that it tries to parse annotations when creating the class, instead of delaying to when the type is requested from the field.

I refactored the `type` funtion on strawberry django plus in this PR (https://github.com/blb-ventures/strawberry-django-plus/pull/235) to solve those issues that I mentioned, and I'll soon (probably this weekend) port that refactoring to this lib as well, meaning that this annotations code will get thrown away.

So, we can either merge this even with those issues based on the fact that the facky solution will have a "short life", or we can avoid it based on the fact that a better overall solution will be made really soon.

What do you think about that? Also, any suggestions @patrick91 ?

tests/test_relay.py
```diff
@@ -0,0 +1,57 @@
...
+    class UserType(relay.Node):
+        id: relay.NodeID[int]
+        name: auto
+
+    @strawberry.type
+    class Query:
+        user: UserType
+
+    schema = strawberry.Schema(query=Query, config=StrawberryConfig(False))
+    str(schema)
```
bellini666, line 21: Left over

tests/test_relay.py
```diff
@@ -0,0 +1,57 @@
...
+    assert UserType._id_attr == "id"
```
bellini666, line 22: Maybe call `UserType.resolve_id_attr()` instead of accessing the private attr?

tests/test_relay.py
```diff
@@ -0,0 +1,57 @@
...
+        def resolve_id(cls, root):
+            return root.id
+
+    @strawberry.type
+    class Query:
+        user: UserType
+
+    schema = strawberry.Schema(query=Query, config=StrawberryConfig(False))
+    str(schema)
+    assert UserType._id_attr == "id"
```
bellini666, line 57: Ditto

bellini666 (changes requested):

tests/test_relay.py
devkral: nope ;).
It is for testing/executing print_schema.

tests/test_relay.py
devkral: after executing print_schema, the attribute should be set

tests/test_relay.py
```diff
@@ -0,0 +1,57 @@
...
+
+        @classmethod
+        def resolve_id(cls, root):
+            return root.id
+
+    @strawberry.type
+    class Query:
+        user: UserType
+
+    assert UserType._id_attr is None
```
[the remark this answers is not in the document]
devkral: I would like to do that but there is a bug in strawberry, see my PR

tests/test_relay.py
devkral: same reason

tests/test_relay.py
devkral: updated

strawberry_django/utils.py
```diff
@@ -85,15 +85,29 @@ def is_similar_django_type(a, b):
...
+        namespace = sys.modules[c.__module__].__dict__
         if "__annotations__" in c.__dict__:
             annotations.update(
                 {
                     k: StrawberryAnnotation(v, namespace=namespace)
-                    for k, v in c.__annotations__.items()
+                    for k, v in filter(
+                        lambda item: _is_not_classvar(item[1], namespace),
+                        c.__annotations__.items(),
+                    )
```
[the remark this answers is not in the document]
bellini666: Like this:

```python
for k, v in cls.__annotations__.items()
if not _is_classvar(v, namespace)
```

tests/test_relay.py
bellini666: Oh, I see. So probably you want to compare it to the expected value. If the schema was generated it will print, the main thing to test here is that it prints what we expect

Check `tests/filters/test_filters.py` on function `test_pk_inserted_for_root_field_only` for an example

strawberry_django/utils.py
patrick91: let's do the solution you think it is best @bellini666 😊

and thanks @devkral for all your PRs and work! <3

bellini666: > let's do the solution you think it is best @bellini666 blush
> 
> and thanks @devkral for all your PRs and work! <3

Ok, lets merge this then! :)

strawberry_django/utils.py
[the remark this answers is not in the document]
devkral: cool, didn't know this syntax ;)

RELEASE.md
```diff
@@ -0,0 +1,5 @@
+Release type: patch
```
bellini666, line 1: Oh, there's no autopub integration in here T_T

But I'll not reject the PR because of that. I can remove it next when doing the release

bellini666 (approved):

Issue #139, opened 2022-07-03
Merge strawberry-django-plus into this repo


bellini666:
There are a lot of features on [strawberry-django-plus](https://github.com/blb-ventures/strawberry-django-plus) that not only are usually requested by the community, they really would benefit this project.

I would like to make this a meta issue for discussing which features should be ported.

Here is a list of features that I think should be ported, linked to their respective issues:

- [ ] Better typing for exposed api (#36)
- [ ] Improve attribute resolving by avoiding sync_to_async overhead when the value is already cached (#137)
- [ ] Query optimizer that automatically adds `select_related`/`prefetch_related`/`only` optimizations to querysets (#75)
- [ ] Support for model `property`(#134)
- [ ] Better mutations handling, with support for nested inputs (#133 and #67)

Those here needs further discussion to know if they are interesting for this project or not, since they are integrations with 3rd party libs:

- [ ] Built-in support for django-debug-toolbar (#138)
- [ ] Built-in support for using django's `TextChoices` as enums. This is implemented in `strawberry-django-plus` with the help of [django-choices-field](https://github.com/bellini666/django-choices-field) (#65 and #66)

Those would be better ported with support on `strawberry` first:

- [ ] Relay-style pagination support (https://github.com/strawberry-graphql/strawberry/issues/1573)
- [ ] Relay-style input support that automatically converts function args to an input (https://github.com/strawberry-graphql/strawberry/issues/1573)
- [ ] Permissioned resolvers that uses the django authentication system implemented as directives (https://github.com/strawberry-graphql/strawberry/issues/1694)

oleksandr-luxmedien:
I would love to see that to happened!

bellini666:
This is finally resolved by https://github.com/strawberry-graphql/strawberry-graphql-django/releases/tag/v0.10.0 :)

PR #265, opened 2023-06-20
refactor!: overall revamp of the type/field code and typing improvements

### LINKED ISSUE #136: Better typing for exposed api (e.g. strawberry.django.type, strawberry.django.field, etc)
The current state of this project api is lacking typing in a lot of places.

Even though some of those use `*args`/`**kwargs` to pass not implemented parameters up to `strawberry`, static type checkers like `pyright` doesn't understand that syntax, which means that everything should be explicitly typed, with cautelous consideration in places where overloads and `__dataclass_transform__` should be sed.

This work is already done on [strabwery-django-plus](https://github.com/blb-ventures/strawberry-django-plus) and it can be ported from there. The cases that come to my mind right now are:

- `strawberry.django.type`
- `strawberry.django.input`
- `strawberry.django.partial`
- `strawberry.django.field`
- `strawberry.django.mutation`
- `strawberry.django.ordering`
- `strawberry.django.filter`

There might be other places as well

### LINKED ISSUE #137: Improve attribute resolving by avoiding `sync_to_async` overhead when the value is already cached
[strawberry-django-plus](https://github.com/blb-ventures/strawberry-django-plus) has a custom field implementation which tries to retrieve model attributes from its cache, then it fallsback to calling `sync_to_async(getattr)(model, attr)`.

This is useful because usually the attributes will be already cached and it is safe to retrieve it directly, without having the `sync_to_async` overhead.

This and other field improvements can be ported here.

bellini666:
This is the initial merge of a lot of improvements from
[strawberry-django-plus](https://github.com/blb-ventures/strawberry-django-plus):

* Strawberry django fields will not resolve their types lazily, just
  like strawberry core fields does. This is a major improvement for
  future annotations and other typing corner case issues.

* Django fields will now try to resolve their values from the django's
  object cache first, then fallback to the "safe resolver". This
  should remove some of the overhead `sync_to_async` causes when
  we don't actually need it.

* Since this is a major refactor, all unused and deprecated code was
  removed.

* Everything in this lib was properly typed. This means that
  `strawberry_django.type(...)`/`strawberry_django.field(...)`/etc
  will now properly validate its arguments and offer proper suggestions
  for the user.

* The code is 100% free of pyright issues (except the tests). There's
  now a new CI test that will run pyright on the codebase.

This should set the ground for the remaining work that needs
to be done for #139.

Fix #136
Fix #137


.github/workflows/main.yml
```diff
@@ -1,63 +1,98 @@
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
       - uses: pre-commit/action@v3.0.0
         with:
           extra_args: -a
```
patrick91, line 28: we don't need this anymore I think 😊 we use pre-commit.ci

.github/workflows/main.yml
```diff
@@ -1,63 +1,98 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+          cache: poetry
+      - name: Install dependencies
+        run: |
+          poetry install -E debug-toolbar
+      - name: Check for pyright errors
+        uses: jakebailey/pyright-action@v1
```
patrick91, line 49: nice!

patrick91 (approved):
I had a very quick look and it looks good to me 😊

.github/workflows/main.yml
bellini666: Oh, that's nice to know! Will remove this then

PR #267, opened 2023-06-25
feat: relay integration

bellini666:
Relay integration ported from strawberry-django-plus, using the official strawberry relay integration

### DIFF
--- a/strawberry_django/fields/base.py
+++ b/strawberry_django/fields/base.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Mapping, Optional, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Optional, TypeVar, cast
 
 from strawberry import LazyType, relay
 from strawberry.annotation import StrawberryAnnotation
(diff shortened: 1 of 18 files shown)


strawberry_django/fields/base.py
```diff
@@ -43,11 +60,86 @@ def is_basic_field(self) -> bool:
         """
         return False
 
-    @property
-    def type(  # noqa: A003
+    @cached_property
+    def django_type(self) -> type[utils.WithStrawberryDjangoObjectDefinition] | None:
```
patrick91, line 64: nit: could we do this?

```suggestion
    def django_type(self) -> type[WithStrawberryDjangoObjectDefinition] | None:
```

strawberry_django/fields/base.py
```diff
@@ -43,11 +60,86 @@ def is_basic_field(self) -> bool:
...
+        origin = self.type
+
+        tdef = get_object_definition(origin)
```
patrick91, line 67: nit, can we change tdef to type_definition or object_definition? :D

patrick91 (approved):
From a quick look it looks great!

strawberry_django/fields/base.py
bellini666: Done! :)

strawberry_django/fields/base.py
bellini666: Done! :)

PR #270, opened 2023-06-27
feat: enum integration

### LINKED ISSUE #65: Support django Choices for graphql enums
With django 3.0 came the choices model field type [See here](https://docs.djangoproject.com/en/3.0/ref/models/fields/#choices). It would be nice if this integration converted those values into an enum for us so we don't have to declare the same enum twice and the source of truth can remain the model.

E.G.

Declaration in model
```
class PlayerTypes(models.TextChoices):
    FA = "FA", "Free Agent"
    PermFA = "PermFA", "Permanent Free Agent"
    Signed = "Signed", "Signed"
    Inactive = "Inactive", "Inactive"
```

Example usage?
```
@strawberry.django.type(models.Player)
class Player:
    name: auto
    type = strawberry.enum(models.PlayerTypes)
```

### LINKED ISSUE #66: Strawberry.enum does not return a valid class type
When defining an enum without the class definition since I am dynamically building it, I get

image

If I try to use strawberry.enum on the actual field I get

image

bellini666:
feat: enum integration

This enables exposing fields with choices in 2 ways, ported from strawberry-django-plus:

1. It integrates with [django-choices-field](https://github.com/bellini666/django-choices-field)
   to be able to handle the enum itself that already exists

2. It can optionally auto generate an enum from a field's `choices`
   attribute.

Fix #65
Fix #66

### DIFF
--- a/strawberry_django/fields/types.py
+++ b/strawberry_django/fields/types.py
@@ -1,5 +1,7 @@
 import datetime
 import decimal
+import enum
+import inspect
 import uuid
 from typing import (
     TYPE_CHECKING,
@@ -13,6 +15,7 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    cast,
 )
 
 import django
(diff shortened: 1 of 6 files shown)


strawberry_django/fields/types.py
```diff
@@ -340,25 +352,81 @@ def resolve_model_field_type(
...
+
+    # Django choices field
+    if (
+        TextChoicesField is not None
+        and IntegerChoicesField is not None
+        and isinstance(
+            model_field,
+            (TextChoicesField, IntegerChoicesField),
+        )
+    ):
```
patrick91, line 365: I have the feeling we could migrate these if to call a function, to make it clearer what they do 😊

same for the content, but we can do that in another PR

patrick91 (approved):
looks good, I only left one comment, but that's for the future I'd say

PR #271, opened 2023-06-28
feat: query optimizer extension

### LINKED ISSUE #75: Reduce queries with prefetching
Requests can easily blow out to hundreds of queries making it very slow. I'm not sure what the "best way" would be to solve this but I needed to fix it quickly and was able to go from 200 requests in a list of 50 objects to just 1.

Hopefully, there's some better way to have this built-in that would support fragments etc.

```python
@strawberry.django.type(models.InventoryItem, pagination=True)
class InventoryItem:
    id: auto
    name: str
    category: 'InventoryCategory'
    supplier: 'InventorySupplier'

    def get_queryset(self, queryset, info, **kwargs):
        selection_set_node = info.field_nodes[0].selection_set
        fields = [selection.name.value for selection in selection_set_node.selections]
        related = [field for field in fields if field in ['category', 'supplier']]

        return queryset.select_related(*related)
```

bellini666:
Queryset optimizer extension, ported from strawberry-django-plus.

obs. This was built on top of #268, so it is better to merge it first (but it can be
closed and merged together with this one).

Fix #75

### DIFF
--- a/strawberry_django/utils/typing.py
+++ b/strawberry_django/utils/typing.py
@@ -1,8 +1,20 @@
 from __future__ import annotations
 
 import sys
-from typing import TYPE_CHECKING, Any, ClassVar, TypeVar, overload
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Iterable,
+    Sequence,
+    TypeVar,
+    Union,
+    overload,
+)
 
+from django.db.models import Prefetch
+from graphql.type.definition import GraphQLResolveInfo
 from strawberry.annotation import StrawberryAnnotation
 from strawberry.auto import StrawberryAuto
 from strawberry.type import (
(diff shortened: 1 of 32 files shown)


patrick91 (approved):

strawberry_django/utils/typing.py
```diff
@@ -62,27 +82,19 @@ def get_django_definition(
...
+def is_auto(obj: Any) -> bool:
     if isinstance(obj, str):
-        # Support future references
         return obj in ["auto", "strawberry.auto"]
 
     return isinstance(obj, StrawberryAuto)
 
 
-def fields(obj: WithStrawberryObjectDefinition) -> list[StrawberryField]:
-    return obj.__strawberry_definition__.fields
```
bmrobin, line 74: @bellini666 can you tell me about why this was removed?  our team was using this and i am failing to see where this method may have moved to during this refactor.

bmrobin: thanks, i'll prefer the latter since using "d-under" fields in libraries is typically frownded upon.

PR #286, opened 2023-07-07
Handle Django GENERATE_ENUMS_FROM_CHOICES with strawberry.auto

pcraciunoiu:
## Description

It seems that `GENERATE_ENUMS_FROM_CHOICES` breaks `strawberry.auto` types with this error:

```
TypeError: DjangoModelStrawberryType fields cannot be resolved. 'NoneType' object is not iterable
```

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry_django/fields/types.py
+++ b/strawberry_django/fields/types.py
@@ -435,7 +435,7 @@ def resolve_model_field_type(
     elif (
         settings["GENERATE_ENUMS_FROM_CHOICES"]
         and isinstance(model_field, Field)
-        and hasattr(model_field, "choices")
+        and getattr(model_field, "choices", None)
     ):
         choices = cast(List[Tuple[Any, str]], model_field.choices)
         field_type = getattr(model_field, "_strawberry_enum", None)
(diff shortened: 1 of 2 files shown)


strawberry_django/fields/types.py
```diff
@@ -435,7 +435,7 @@ def resolve_model_field_type(
     elif (
         settings["GENERATE_ENUMS_FROM_CHOICES"]
         and isinstance(model_field, Field)
-        and hasattr(model_field, "choices")
+        and getattr(model_field, "choices", None)
```
pcraciunoiu, line 438: In my testing I found that `model_field.choices` was set to `None` even for `CharField`

bellini666:
@pcraciunoiu nice! :)

I just noticed that we don't have a test for this, and that's probably the reason this slipped during the merge.

Could you add a test for it just to make sure it doesn't break again in the future?

bellini666 (changes requested):
Looks great!

Just need to fix the linting stuff that broke from `pre-commit`

pcraciunoiu:
> Looks great!
> 
> Just need to fix the linting stuff that broke from `pre-commit`

Sync! Pushed just now

bellini666 (approved):

PR #288, opened 2023-07-08
Support using `auto` with Postgres `ArrayField`s

colinmcdonald22:
## Description

Previously, any `django.contrib.postgres.fields.ArrayField`s used in models had to be typed manually (say, as `list[MyEnum]`). Now they're compatible with `auto`, and resolve to a `List` of the type of `base_field`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [X] New feature
- [ ] Documentation

Fix https://github.com/strawberry-graphql/strawberry-graphql-django/issues/200

## Checklist

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -78,97 +78,97 @@ files = [
 
 [[package]]
 name = "charset-normalizer"
-version = "3.1.0"
+version = "3.2.0"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 optional = false
 python-versions = ">=3.7.0"
 files = [
-    {file = "charset-normalizer-3.1.0.tar.gz", hash = "sha256:34e0a2f9c370eb95597aae63bf85eb5e96826d81e3dcf88b8886012906f509b5"},
-    {file = "charset_normalizer-3.1.0-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:e0ac8959c929593fee38da1c2b64ee9778733cdf03c482c9ff1d508b6b593b2b"},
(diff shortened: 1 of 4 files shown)


colinmcdonald22:
Currently it seems like there's two issues here:
1. I'm always adding Postgres dependencies to dev installs. I'd imagine we'd want to make sure tests pass w/o Postgres deps installed (for if they're not at runtime) - is there any precedent for splitting dev dependencies like this? I've only seen Django/Python versions handled specially, not any other optional dependencies.
2. Even with Postgres dependencies added (verified w/ Poetry output in GH Actions logs), this is still failing - even though it passes locally. I'll dig more into this soon.

baseplate-admin:
Shouldn't we use psycopg3 ? 

Django by default now uses psycopg3

bellini666:
> Shouldn't we use psycopg3 ?
> 
> Django by default now uses psycopg3

The dependency here is a "dev" one and mostly for tests, meaning that if we used `psycopg3` we would not be able to test on django 4.1 or older

noamsto:
Why was this closed and not merged?
Is there missing todos here?

patrick91:
@bellini666 do you think we should get this in?

bellini666:
> @bellini666 do you think we should get this in?

Yes I do! It makes sense

Not sure why @colinmcdonald22 closed this, but I can take over it to ensure it will get in

noamsto:
@bellini666 
Maybe this PR should be re-opened so it won't be forgotten ?

dperetti:
I need this!

bellini666:
> I need this!

You shall have this https://github.com/strawberry-graphql/strawberry-django/pull/567

Issue #106, opened 2022-03-23
Add support for `all_fields`


patrick91:
In Strawberry's pydantic extension we have support `all_fields`. I think we should add it for strawberry django too :)

https://strawberry.rocks/docs/integrations/pydantic#pydantic-support

viaregio:
Personally I think all_fields is not the way to go. Other libraries like django-ninja or djantic have a more useful mechnanism where you can include or exclude fields.

How do you include all fields except one or two fields?

bellini666:
I have mixed feelings regarding this...

Coming from graphene myself, I feel that automatically exposing all fields from the django model is can lead to a lot of security issues.

For example, I used to love the simplicity of having everything exposed and only defining a few `"exclude"` that should not be exposed, but one day we had an issue with some leaked sensitive information that got exposed by mistake when someone added a new column and forgot to add it to the `"exclude"` list. Since then we changed everything to a "opt-in" approach instead of "opt-out".

Having said that, I don't fully oppose to the ideas here, but I think that if it gets implemented, it should be documented with a "warning" message mentioning that possibility of sensitive data leak by mistake.

Regarding the implementation, we have 2 use cases here:

1) We have an "opt-in" approach, which is basically the way the integration works right now. I don't see a reason to why we would require a `"fields"` list since the current approach of defining that "list" as attributes of the type, and typing them as "auto" is almost as simple as the list itself.

2) We have an "all-in"/"opt-out" approach in which we assume that all fields are included by default, and we can define some to be "excluded". We could go with a new argument `all_fields=True` (false by default) that triggers the "include all fields" introspection, and an optional `exclude_fields` list that defines a list of fields to be excluded when `all_fields is True`

> This would ease a lot the migration from graphene-django, to be fair

There are a lot of people migrating from graphene to strawberry (I'm also included in that list), I imagine if it would be worth to write a small compatibility layer to apply on top of graphene types/fields to transform them into strawberry ones, which would allow people to migrate more easily instead of having to make an "all or nothing" migration.

Lucasmiguelmac:
I am really in favor of implementing the features described in this thread. Though I have some notes on top of what has already been discussed and argued:

* Regarding an `all_fields` implementation:
[...]

PR #293, opened 2023-07-11
Add new keywords "fields" and "exclude" to type decorator for auto-population of Django model fields

coleshaw:
## Description

Per [the discussion around issue 106](https://github.com/strawberry-graphql/strawberry-graphql-django/issues/106), this PR adds two keyword arguments to the `type` decorator -- `fields` and `exclude`. When provided, these arguments help auto-populate the Strawberry fields from the Django model, giving `strawberry.django` similar behavior to the `strawberry.pydantic` and `graphene-django` libraries.

Note that the core of the work was pulled from [the prior repository for `strawberry-django-plus`](https://github.com/blb-ventures/strawberry-django-plus/issues/85), and real credit goes to @mhdismail.

When included, `fields` can be:

* `__all__`. Automatically include all Django model fields, using `strawberry.auto` as the type annotation.
* `[<List of field names>]`. Include the listed Django model fields, using `strawberry.auto` as the type annotation. Invalid names are ignored.

When included, `exclude` can be:

* `[<List of field names>]`. Include all Django model fields **except** for the ones listed here, using `strawberry.auto` as the type annotation. Invalid names are ignored. Must include at least one item for this to take effect, an empty list is ignored.

`fields` has precedence over `exclude`, so if both are provided, `exclude` is ignored.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* #106 

## Checklist

- [] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

I added tests to cover the new code, but I don't have a local project using Strawberry that I can manually test / verify against ... so it might be helpful if someone could try this against a project they have locally?

@bellini666 , I had some time to migrate this over today.

### DIFF
--- a/docs/guide/fields.md
+++ b/docs/guide/fields.md
@@ -122,7 +122,7 @@ strawberry_django.field_type_map.update({
 
     These new keywords should be used with caution, as they may inadvertently lead to exposure of unwanted data. Especially with `fields="__all__"` or `exclude`, sensitive model attributes may be included and made available in the schema without your awareness.
 
-`strawberry.django.type` includes two optional keyword fields to help you more easily populate fields from the Django model, `fields` and `exclude`.
+`strawberry.django.type` includes two optional keyword fields to help you populate fields from the Django model, `fields` and `exclude`.
 
 Valid values for `fields` are:
 
--- a/docs/guide/fields.md
+++ b/docs/guide/fields.md
@@ -122,8 +122,8 @@ strawberry_django.field_type_map.update({
 
 Valid values for `fields` are:
 
-  * `__all__` to assign `strawberry.auto` as the field type for all model fields.
-  * `[<List of field names>]` to assign `strawberry.auto` as the field type for the enumerated fields. These can be combined with manual type annotations if needed.
+- `__all__` to assign `strawberry.auto` as the field type for all model fields.
+- `[<List of field names>]` to assign `strawberry.auto` as the field type for the enumerated fields. These can be combined with manual type annotations if needed.
 
 ```{.python title=fields_all.py}
 @strawberry.django.type(models.Fruit, fields="__all__")
@@ -145,7 +145,7 @@ class FruitType:
 
 Valid values for `exclude` are:
 
-  * `[<List of field names>]` to exclude from the fields list. All other Django model fields will included and have `strawberry.auto` as the field type. These can also be overriden if another field type should be assigned. An empty list is ignored.
+- `[<List of field names>]` to exclude from the fields list. All other Django model fields will included and have `strawberry.auto` as the field type. These can also be overriden if another field type should be assigned. An empty list is ignored.
 
 ```{.python title=exclude.py}
 @strawberry.django.type(models.Fruit, exclude=["name"])
--- a/docs/guide/fields.md
+++ b/docs/guide/fields.md
@@ -159,7 +159,7 @@ class FruitType:
     color: int
 ```
 
-Note that if `fields` has precedence over `exclude`, so if both are provided, then `exclude` is ignored.
+Note that `fields` has precedence over `exclude`, so if both are provided, then `exclude` is ignored.
 
 ## Overriding the field class (advanced)
 
--- a/docs/guide/fields.md
+++ b/docs/guide/fields.md
@@ -120,7 +120,7 @@ strawberry_django.field_type_map.update({
 
 !!! warning
 
-    These new keywords should be used with caution, as they may inadvertently lead to exposure of unwanted data. Especially with `fields="__all__"` or `exclude`, sensitive model attributes may be included and made available without your awareness.
+    These new keywords should be used with caution, as they may inadvertently lead to exposure of unwanted data. Especially with `fields="__all__"` or `exclude`, sensitive model attributes may be included and made available in the schema without your awareness.
 
 `strawberry.django.type` includes two optional keyword fields to help you more easily populate fields from the Django model, `fields` and `exclude`.
 
--- a/docs/guide/fields.md
+++ b/docs/guide/fields.md
@@ -118,6 +118,10 @@ strawberry_django.field_type_map.update({
 
 ## Including / excluding Django model fields by name
 
+!!! warning
+
+    These new keywords should be used with caution, as they may inadvertently lead to exposure of unwanted data. Especially with `fields="__all__"` or `exclude`, sensitive model attributes may be included and made available without your awareness.
+
 `strawberry.django.type` includes two optional keyword fields to help you more easily populate fields from the Django model, `fields` and `exclude`.
 
 Valid values for `fields` are:
@@ -125,19 +129,19 @@ Valid values for `fields` are:
 - `__all__` to assign `strawberry.auto` as the field type for all model fields.
 - `[<List of field names>]` to assign `strawberry.auto` as the field type for the enumerated fields. These can be combined with manual type annotations if needed.
 
-```{.python title=fields_all.py}
+```{.python title="All Fields"}
 @strawberry.django.type(models.Fruit, fields="__all__")
 class FruitType:
     pass
 ```
 
-```{.python title=fields_enumerated.py}
+```{.python title="Enumerated Fields"}
 @strawberry.django.type(models.Fruit, fields=["name", "color"])
 class FruitType:
     pass
 ```
 
-```{.python title=fields_overriden.py}
+```{.python title="Overriden Fields"}
 @strawberry.django.type(models.Fruit, fields=["color"])
 class FruitType:
     name: str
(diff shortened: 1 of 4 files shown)


docs/guide/fields.md
```diff
@@ -116,6 +116,51 @@ strawberry_django.field_type_map.update({
...
+## Including / excluding Django model fields by name
+
+`strawberry.django.type` includes two optional keyword fields to help you more easily populate fields from the Django model, `fields` and `exclude`.
+
+Valid values for `fields` are:
+
+- `__all__` to assign `strawberry.auto` as the field type for all model fields.
+- `[<List of field names>]` to assign `strawberry.auto` as the field type for the enumerated fields. These can be combined with manual type annotations if needed.
+
+```{.python title=fields_all.py}
```
bellini666, line 128: In this case, since all those would not actually go to difference files, maybe define the title to something more semantic?

Like `title="All Fields"`, and `title="Enumerated Fields"` below, and so on

docs/guide/fields.md
```diff
@@ -116,6 +116,51 @@ strawberry_django.field_type_map.update({
 })
 ```
 
+## Including / excluding Django model fields by name
+
```
bellini666, line 120: I would add a `warning` here specifying that this should be used with caution because it can expose unwanted data by mistake, specially when using `__all__`

You can check an example of a warning at the `docs/guide/authentication.md` file

strawberry_django/type.py
```diff
@@ -394,6 +415,8 @@ def type(  # noqa: A001
     select_related: Optional[TypeOrSequence[str]] = None,
     prefetch_related: Optional[TypeOrSequence[PrefetchType]] = None,
     disable_optimization: bool = False,
+    fields: Optional[Union[List[str], str]] = None,
```
bellini666, line 418:
```suggestion
    fields: Optional[Union[List[str], Literal["__all__"]] = None,
```

strawberry_django/type.py
```diff
@@ -79,9 +80,29 @@ def _process_type(
...
```
bellini666, line 83:
```suggestion
    fields: Optional[Union[List[str], Literal["__all__"]] = None,
```

strawberry_django/type.py
```diff
@@ -79,9 +80,29 @@ def _process_type(
...
     **kwargs,
 ) -> _O:
     is_input = kwargs.get("is_input", False)
+
+    if not exclude:
+        exclude = []
+
+    if fields == "__all__":
+        model_fields = list(model._meta.fields)
+    elif isinstance(fields, collections.abc.Sequence):
```
bellini666, line 94: Since you are checking here for `Sequence`, you should type it as `Sequence` as well in the arguments

Also, you should be importing `Sequence` from `typing` to avoid the issue pyright is raising

bellini666: Also, I would type this as a `Collection`, since you are not actually using this for iteration but instead to check if a name is "contained" in that

strawberry_django/type.py
```diff
@@ -79,9 +80,29 @@ def _process_type(
...
     disable_optimization: bool = False,
+    fields: Optional[Union[List[str], str]] = None,
+    exclude: Optional[List[str]] = None,
     **kwargs,
 ) -> _O:
     is_input = kwargs.get("is_input", False)
+
+    if not exclude:
+        exclude = []
+
```
bellini666, line 91: You don't actually need this because you `isinstance` should be enough protection

strawberry_django/type.py
```diff
@@ -79,9 +80,29 @@ def _process_type(
...
+    if fields == "__all__":
+        model_fields = list(model._meta.fields)
+    elif isinstance(fields, collections.abc.Sequence):
+        model_fields = [f for f in model._meta.fields if f.name in fields]
+    elif isinstance(exclude, collections.abc.Sequence) and len(exclude) > 0:
```
bellini666, line 96: Ditto here

strawberry_django/type.py
```diff
@@ -79,9 +80,29 @@ def _process_type(
...
+        model_fields = [f for f in model._meta.fields if f.name not in exclude]
+    else:
+        model_fields = []
+
+    for f in model_fields:
+        if cls.__annotations__.get(f.name):
```
bellini666, line 102: You need to use `get_annotations` here to also consider annotations from parents

```suggestion
    existing_annotations = set(get_annotations(cls).keys())
    for f in model_fields:
        if f.name in existing_annotations:
```

tests/test_types.py
```diff
@@ -202,3 +202,169 @@ class Query:
...
+    assert name_field is not None
+    assert name_field.type == str
+
+    color_field = type_def.get_field("color")
+    assert color_field is not None
+
+    sweetness_field = type_def.get_field("sweetness")
+    assert sweetness_field is not None
+    assert sweetness_field.type == int
+    assert sweetness_field.description == Fruit._meta.get_field("sweetness").help_text
```
bellini666, line 231: Another way of testing this all would be to test the printed schema.

You can see in the `tests/test_queries.py` how `test_field_name` does that as an example.

Since those are tests I'm fine with not doing that, but I'll leave this comment here in case you want to do that :)

bellini666 (changes requested):
Left some comments/suggestions :)

tests/test_types.py
coleshaw: Thank you for the suggestion, I like the printed schema method better 👍

coleshaw (commented):
Thank you for the great suggestions and feedback! Updated the PR.

bellini666:
@coleshaw everything seems fine now :)

There's just a small question I left because I don't think we should be forcing `__annotations__` to `{}` (unless there's a reason to do so?)

coleshaw:
> @coleshaw everything seems fine now :)
> 
> There's just a small question I left because I don't think we should be forcing `__annotations__` to `{}` (unless there's a reason to do so?)

Yeah, I agree, it looks kind of weird ... so the problem I was running into was that on Python 3.8 / 3.9, I think that classes do not have a default `__annotations__` object when there are no type annotations provided. That was my assumption from looking at the CI results, example [here](https://github.com/strawberry-graphql/strawberry-graphql-django/actions/runs/5522484569/jobs/10072140926?pr=293#step:8:1275), where everything passes on Python 3.10 / 3.11 but fails on Python 3.8 / 3.9.

If you comment out that assignment on the current code, you get the same `AttributeError` exception:

```
        existing_annotations = get_annotations(cls)
        # if not existing_annotations:
        #     cls.__annotations__ = {}

        for f in model_fields:
            if existing_annotations.get(f.name):
                continue
>           cls.__annotations__[f.name] = strawberry.auto
E           AttributeError: type object 'FruitType' has no attribute '__annotations__'
```

The docs [recommend](https://docs.python.org/3/howto/annotations.html#best-practices-for-annotations-in-any-python-version) that if you have to set `__annotations__`, to use a `dict` ...

Not sure if there's a better way to approach it, or if I missed something?

bellini666:
> Not sure if there's a better way to approach it, or if I missed something?

Sorry for the delay in the reply.

And gotcha regarding the issue! :).

But as you pointed out in the docs, it says that `You should avoid assigning to the __annotations__ member of objects directly. Let Python manage setting __annotations__.` and `You should avoid modifying __annotations__ dicts.`. Although your workaround is fine for 3.8/3.9, not that on 3.10+ you are in fact modifying `__annotations__` to be another dict in case it, for some reason, has no annotations in it.

`__annotations__` are lazily created if the class don't have one. For example, on python 3.11:

```python
>>> class Foo:
...     ...
... 
>>> Foo.__dict__
mappingproxy({'__module__': '__main__', '__dict__': <attribute '__dict__' of 'Foo' objects>, '__weakref__': <attribute '__weakref__' of 'Foo' objects>, '__doc__': None})
>>> Foo.__annotations__
{}
>>> Foo.__dict__
mappingproxy({'__module__': '__main__', '__dict__': <attribute '__dict__' of 'Foo' objects>, '__weakref__': <attribute '__weakref__' of 'Foo' objects>, '__doc__': None, '__annotations__': {}})
```

And on python 3.8:

```python
>>> class Foo:
...     ...
... 
>>> Foo.__dict__
mappingproxy({'__module__': '__main__', '__dict__': <attribute '__dict__' of 'Foo' objects>, '__weakref__': <attribute '__weakref__' of 'Foo' objects>, '__doc__': None})
>>> Foo.__annotations__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'Foo' has no attribute '__annotations__'
```

So, this seems to me that a better solution would be this:

```python
try:
    cls.__annotations__
except AttributeError:
    # <explain here why this is needed for 3.8/3.9>
    cls.__annotations__ = {}
```

coleshaw:
@bellini666 , no worries at all, I assume you do other things besides maintain this library 😃 And I really appreciate all of your time and effort in looking this over and providing suggestions! 🙌

I've updated that line to a `try / except` as suggested, it does look cleaner 👍 , thank you!

bellini666 (approved):

PR #297, opened 2023-07-12
Fix missing model docstring crash

Mapiarz:
If TYPE_DESCRIPTION_FROM_MODEL_DOCSTRING is set to True this will lead to crashes in many instances when there's no Python docstring provided.

This looks like a typo.


bellini666 (approved):
It indeed is a typo! Thanks for the fix

PR #302, opened 2023-07-13
refactor!: use a setting to decide if we should map fields to relay types or not

bellini666:
`strawberry-django-plus` used to do that, but when migrating this
I thought it would be enough to check if the type itself inherits from
`relay.Node`.

It turns out that it is not, specially for filters as those doesn't
inherit from `relay.Node`.

This adds an option to allow the user to use `ID` or `GlobalID`
for their auto fields, meaning that someone that only uses `Node`
types can set `MAP_AUTO_ID_AS_GLOBAL_ID=True` to make sure that
it will only use `GlobalID` to map those.

### DIFF
--- a/docs/guide/filters.md
+++ b/docs/guide/filters.md
@@ -30,7 +30,7 @@ input FruitFilter {
 
     If you are using the [relay integration](relay.md) and working with types inheriting
     from `relay.Node` and `GlobalID` for identifying objects, you might want to set
-    `MAP_AUTO_ID_AS_GLOBAL_ID=Trye` in your [strawberry django settings](../settings)
+    `MAP_AUTO_ID_AS_GLOBAL_ID=True` in your [strawberry django settings](../settings)
     to make sure `auto` fields gets mapped to `GlobalID` on types and filters.
 
 ## Lookups
--- a/docs/guide/filters.md
+++ b/docs/guide/filters.md
@@ -26,6 +26,13 @@ input FruitFilter {
 }
 ```
 
+!!! tip
+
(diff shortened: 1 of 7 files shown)


docs/guide/filters.md
```diff
@@ -26,6 +26,13 @@ input FruitFilter {
 }
 ```
 
+!!! tip
+
+    If you are using the [relay integration](relay.md) and working with types inheriting
+    from `relay.Node` and `GlobalID` for identifying objects, you might want to set
+    `MAP_AUTO_ID_AS_GLOBAL_ID=Trye` in your [strawberry django settings](../settings)
```
patrick91, line 33:
```suggestion
    `MAP_AUTO_ID_AS_GLOBAL_ID=True` in your [strawberry django settings](../settings)
```

docs/guide/relay.md
```diff
@@ -45,6 +45,13 @@ Behind the scenes this extension is doing the following for you:
 You can also define your own `relay.NodeID` field and your resolve, in the same way as
 `some_model_conn_with_resolver` is doing. In those cases, they will not be overridden.
 
+!!! tip
+
+    If you are only working with types inheriting from `relay.Node` and `GlobalID`
+    for identifying objects, you might want to set `MAP_AUTO_ID_AS_GLOBAL_ID=Trye`
```
patrick91, line 51:
```suggestion
    for identifying objects, you might want to set `MAP_AUTO_ID_AS_GLOBAL_ID=True`
```

patrick91 (approved):

PR #363, opened 2023-09-15
Support inList and nInList lookup in filters on enum

cpontvieux-systra:
## Description

Enum used in filters are evaluated to their `.value` for doing Django queries.

Unfortunately if a FilterLookup is used and a list is used for a lookup, then this transformation is not done, resulting in non-working django filter.

## Types of Changes
- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

The correction is to check is the value is a non-empty list for at least one Enum and then create a new list with the `.value` of each enum.
I have added some unit tests.

*

## Checklist

- [X] My code follows the code style of this project. (to my knowledge)
- [X] I have read the CONTRIBUTING document. (the file resides in `docs` but yes)

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -132,6 +132,12 @@ def build_filter_kwargs(
 
         if isinstance(field_value, Enum):
             field_value = field_value.value
+        elif (
+            isinstance(field_value, list)
+            and len(field_value) > 0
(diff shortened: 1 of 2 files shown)


bellini666 (approved):
Thank you! :)

Issue #372, opened 2023-09-21
GlobalID not returned via query


sdobbelaere:
I've setup Strawberry django for a Company model, and created types, input types, queries via relay and mutations.
Also, I've setup the `MAP_AUTO_ID_AS_GLOBAL_ID: True` setting.

When I look at the query result, I can see that the object ID is mapped as a GlobalID.
<img width="1304" alt="Screenshot 2023-09-21 at 16 32 48" src="https://github.com/strawberry-graphql/strawberry-graphql-django/assets/16462709/8571de91-9489-4046-8470-48efdbd6a814">

However, when I want to try and remove that object, the mutation is complaining that my ID is of the wrong type and should be a base64 string.  
<img width="1318" alt="Screenshot 2023-09-21 at 16 33 19" src="https://github.com/strawberry-graphql/strawberry-graphql-django/assets/16462709/e1f62e61-8b60-4abb-98f3-f40f79859c4c">

If the mutation is correct, shouldn't the query supply me with an actual converted GlobalID instead of a stringified id?
Or perhaps I'm missing something in my setup?

Some more detail on the models, types, inputtypes and queries below:

```python
class Company(models.Model):
    name = models.CharField(max_length=100)

@strawberry.django.type(Company, filters=CompanyFilter, order=CompanyOrder, pagination=True, fields='__all__')
class CompanyType(relay.Node):
    pass

@strawberry_django.input(Company)
class CompanyInput:
    id: auto
    name: auto

@strawberry.type(name="Query")
class ContactsQuery:
    company: CompanyType = strawberry_django.field(extensions=[IsAuthenticated()])
    companies: ListConnectionWithTotalCount[CompanyType] = strawberry_django.connection(extensions=[IsAuthenticated()])


@strawberry.type(name="Mutation")
class ContactsMutation:
    create_company: CompanyType = mutations.create(CompanyInput, extensions=[IsAuthenticated()])
    create_companies: List[CompanyType] = mutations.create(CompanyInput, extensions=[IsAuthenticated()])
    update_company: List[CompanyType] = mutations.update(CompanyPartialInput, extensions=[IsAuthenticated()])
    delete_company: CompanyType = mutations.delete(NodeInput, extensions=[IsAuthenticated()])
    delete_companies: List[CompanyType] = mutations.delete(NodeInput, extensions=[IsAuthenticated()])

```

bellini666:
Hi @sdobbelaere ,

[...]

PR #373, opened 2023-09-22
Exclude id from model fields to avoid overriding the id: type

sdobbelaere:
## Description

Excluding the id from model_fields when MAP_AUTO_ID_AS_GLOBAL_ID is set to True to allow the previously set annotation to handle the id without overriding it.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

#372 

## Checklist

### DIFF
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -95,6 +95,16 @@ def _process_type(
     else:
         model_fields = []
 
+    # If MAP_AUTO_ID_AS_GLOBAL_ID is True, we can no longer set the id
+    # from fields or it will override the GlobalID and return the default 
+    # django id instead in the query-result. This adjustment however still 
+    # does not fix if the id was set to auto manually on the ModelType.
+    MAP_AUTO_ID_AS_GLOBAL_ID = django_settings().get('MAP_AUTO_ID_AS_GLOBAL_ID', False)
+    if MAP_AUTO_ID_AS_GLOBAL_ID:
+        for f in model_fields[:]:
+            if f.name in ['id']:
+                model_fields.remove(f)
+
     existing_annotations = get_annotations(cls)
     try:
         cls.__annotations__  # noqa: B018


strawberry_django/type.py
```diff
@@ -95,6 +95,16 @@ def _process_type(
     else:
         model_fields = []
 
+    # If MAP_AUTO_ID_AS_GLOBAL_ID is True, we can no longer set the id
+    # from fields or it will override the GlobalID and return the default
+    # django id instead in the query-result. This adjustment however still
+    # does not fix if the id was set to auto manually on the ModelType.
+    MAP_AUTO_ID_AS_GLOBAL_ID = django_settings().get("MAP_AUTO_ID_AS_GLOBAL_ID", False)
+    if MAP_AUTO_ID_AS_GLOBAL_ID:
```
bellini666, line 103: I personally don't like using UPPER_CASED for variable names other than constants.

I suggest either setting the variable lower cased, or better yet, use the settings directly like in my suggestion:

```suggestion
    if django_settings().get("MAP_AUTO_ID_AS_GLOBAL_ID", False):
```

strawberry_django/type.py
```diff
@@ -95,6 +95,16 @@ def _process_type(
...
+        for f in model_fields[:]:
+            if f.name in ["id"]:
+                model_fields.remove(f)
```
bellini666, line 106: nit

```suggestion
        model_fields = [f for f in model_fields if f.name != "id"]
```

bellini666 (changes requested):
Looks good! :)

Just a couple of minor suggestions and should be good to merge

strawberry_django/type.py
sdobbelaere: I can see that, it was more to indicate it came from the settings.  But fair enough.
What I wouldn't do it fetch it directly from the settings inside the if statement.  It makes for hard to read code.

Issue #375, opened 2023-09-26
Getting subscriptions running


sdobbelaere:
Hi Everyone!

I'm trying to get subscriptions working in some models, so that I can push updated-information to the the frontend after our task-manger has executed certain actions.  

However I seem to be stuck on getting a basic subscription running from the docs:
```python
from strawberry import type, subscription
import asyncio

@type
class Subscription:
    @subscription
    async def count(self, target: int = 100) -> int:
        for i in range(target):
            yield i
            await asyncio.sleep(0.5)
``` 
When I try out this subscription in graphiql this is the response:

```json
// subscription{count(target:32)}
// yields

{
  "errors": [
    {
      "isTrusted": true
    }
  ]
}
```

How did I setup the subscription?

```python
# urls
urlpatterns = [
    path('admin/', admin.site.urls),
    path('graphql/', AsyncGraphQLView.as_view(
        schema=schema,
        graphiql=True,
        subscriptions_enabled=True)),
]

# Schema
schema = strawberry.Schema(
    query=Query,
    mutation=Mutation,
    subscription=Subscription,
    extensions=[DjangoOptimizerExtension()]
)
```

I must admit I was getting confused that needs to be done to actually get the subscriptions running as in one part of the docs you just set `AsyncGraphQLView.as_view. subscriptions_enabled` to True.  But in other doc's you need to create a new AsyncgraphQlView. Both approaches didn't seem to get me further, so I must be missing something crucial.

Help is tremendously appreciated, feel like I'm going in circles. :-)

bellini666:
Hi @sdobbelaere ,

Probably we need to improve the documentation regarding subscription support within django.

But basically, to use subscriptions you need to be running with ASGI even in development (tip: you can use `daphne` which will replace your `runserver` with an ASGI one), and that ASGI server needs to be able to handle websockets connections. For that you can use the [channels integration](https://strawberry.rocks/docs/integrations/channels)

sdobbelaere:
@bellini666 I see, that put things into perspective indeed.  

I'm also noticing in the docs that for non-channel connoisseurs, the section that gives back the channel controle makes things harder than they should.  

I'll have a look to clarify the docs while I'm working my way though things to get it running. Feel free to add comments on changes on that pull request when it comes.

sdobbelaere:
@bellini666 I've created a pull-request with my findings getting things running on my end. This should help new users getting things on the rails at least with a counter example.  

[...]

PR #376, opened 2023-09-27
Docs improvement Subscriptions

sdobbelaere:
As referred to in issue #375 some improved guidance on getting subscriptions running including the missing pieces related to the test-server.


docs/guide/subscriptions.md
```diff
@@ -1,6 +1,136 @@
...
+This guide will give you a minimal working example to get you going.
+There are 3 parts to this guide:
+
+1. Making Django compatible
+2. Setup local testing
+3. Creating your first subscription
+
+## Making Django compatible
+
+It's important to realise that Django doesnt support websockets out of the box.
```
bellini666, line 15:
```suggestion
It's important to realise that Django doesn't support websockets out of the box.
```

docs/guide/subscriptions.md
```diff
@@ -1,6 +1,136 @@
...
+To resolve this, we can help the platform along a little.
+
+Edit your `MyProject.asgi.py` file and replace it with the following content.
+Ensure that you replace the relevant code with your setup.
```
bellini666, line 19: I think it is probably interesting to point out that [channels](https://channels.readthedocs.io/en/latest/) can be used to add websockets capabilities to django, and that the user can follow its instructions on how to setup it (as it will probably be more updated and have more content than us).

After that we can provide a minimal example to get started

docs/guide/subscriptions.md
```diff
@@ -1,6 +1,136 @@
...
+...
+
+urlpatterns = [
+	...
+    path('graphql/', AsyncGraphQLView.as_view(
+        schema=schema,
+        graphiql=settings.DEBUG,
+        subscriptions_enabled=True
+        )
+    ),
```
bellini666, line 56: nit to what `black` would do

```suggestion
    path(
        'graphql/',
        AsyncGraphQLView.as_view(
            schema=schema,
            graphiql=settings.DEBUG,
            subscriptions_enabled=True,
        ),
    ),
```

docs/guide/subscriptions.md
```diff
@@ -1,6 +1,136 @@
...
+]
+
+```
+
+Note, django-channels allows for a lot more complexity. Here we just cover the basic framework to get
+subscriptions to run on Django with minimal effort.
+
+## Setup local testing
+
+The classic `./manage.py runserver` will not support subscriptions. However, Django has daphne support out of the box to ensure that we can actually use Daphne for the runserver command. [Source](https://docs.djangoproject.com/en/4.2/howto/deployment/asgi/daphne/)
```
bellini666, line 67: Maybe point out that the default `runserver` implementation runs on WSGI mode, meaning the channels integration would not work with it.

Then mention that about `daphne`, that is an ASGI protocol server, provides a custom `runserver` that runs on ASGI.

Maybe also important to mention that `daphne` can be used to help with local development, even if the user is running another ASGI protocol server in production, like `uvicorn`, `hypercorn`, etc

docs/guide/subscriptions.md
```diff
@@ -1,6 +1,136 @@
...
+```python
+INSTALLED_APPS = [
+	...
+    'daphne',
+    'django.contrib.staticfiles',
+    ...
+]
+```
+
+and add your ASGI_APPLICATION setting in your settings.py
```
bellini666, line 86: nit

```suggestion
and add your `ASGI_APPLICATION` setting in your settings.py
```

docs/guide/subscriptions.md
```diff
@@ -1,6 +1,136 @@
...
+Once you've taken care of those 2 setup steps, your first subscription is a breeze.
+Go and edit your schema-file and add:
+
+```python
+from strawberry import type, subscription
+import asyncio
+
+@type
+class Subscription:
+    @subscription
```
bellini666, line 112: nit to align with the pattern we use across other documentations

```suggestion
import asyncio
import strawberry

@strawberry.type
class Subscription:
    @strawberry.subscription
```

bellini666 (changes requested):
Very nice in general! :)

sdobbelaere:
- Added the proposed improvements to the docs
- Added a new router that is authentication enabled. (Original router doesn't provide the user object, thus not very batteries included as it needs fixing straight out of the box)
- Consolidated how to find the user-object through get_current_user(info)

strawberry_django/auth/queries.py
```diff
@@ -1,10 +1,28 @@
 import strawberry_django
 
 
+def get_current_user(info):
+    """Get and return the current user based on various scenarios."""
+    try:
+        user = info.context.request.user
```
bellini666, line 7: This can also cause errors when calling this inside an async function

Maybe also provide an `async def aget_current_user(info): return sync_to_async(get_current_user)(info)` bellow for async cases

strawberry_django/auth/queries.py
```diff
@@ -1,10 +1,28 @@
...
+        except AttributeError:
+            # When using this through websockets / subscriptions, scope sits inside of the request
+            user = info.context.get("request").scope.get("user")
+
+    return user
+
+
 def resolve_current_user(info):
-    if not info.context.request.user.is_authenticated:
+    user = get_current_user()
```
bellini666, line 20:
```suggestion
    user = get_current_user(info)
```

strawberry_django/auth/queries.py
```diff
@@ -1,10 +1,28 @@
 import strawberry_django
 
 
+def get_current_user(info):
```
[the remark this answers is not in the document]
bellini666: Also, it would be nice to add typing to the `info` argument and type the return value as `AbstractBaseUser`

tests/projects/schema.py
```diff
@@ -379,7 +380,7 @@ class Query:
 
     @strawberry_django.field
     def me(self, info: Info) -> Optional[UserType]:
-        user = info.context.request.user
+        user = get_current_user()
```
bellini666, line 383:
```suggestion
        user = get_current_user(info)
```

bellini666 (changes requested):

strawberry_django/routers.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
sdobbelaere: I looked at the original too, and that's the one that put me on the wrong trail related to authentication.  
As I was trying to make things more 'battery included' it seems like a wise idea to include the new version.  

Since the original doesn't leave much room for subclassing I decided to copy-paste it, and add the authentication to avoid breaking other peoples installs.

strawberry_django/auth/queries.py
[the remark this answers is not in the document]
sdobbelaere: Importing AbstractBaseUser seems to be giving me some issues.  I'll skip on that for now. But I like the utils.py suggestion a lot, the typing I've added as well.

strawberry_django/auth/queries.py
```diff
@@ -1,10 +1,28 @@
...
+        user = info.context.request.user
+    except AttributeError:
+        try:
+            # When running queries/mutations in ASGI mode, the user is moved into the consumer scope
+            user = info.context.get("request").consumer.scope["user"]
+        except AttributeError:
+            # When using this through websockets / subscriptions, scope sits inside of the request
+            user = info.context.get("request").scope.get("user")
+
+    return user
```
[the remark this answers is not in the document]
sdobbelaere: Sure, I'll do that.  Any thoughts on the id for it? Literally added XXX for now.

strawberry_django/auth/queries.py
bellini666: You marked this as resolved, but it is still missing the info =P

strawberry_django/auth/utils.py
```diff
@@ -0,0 +1,23 @@
...
+    except AttributeError:
+        try:
+            # When running queries/mutations in ASGI mode, the user is moved into the consumer scope
+            user = info.context.get("request").consumer.scope["user"]
+        except AttributeError:
+            # When using this through websockets / subscriptions, scope sits inside of the request
+            user = info.context.get("request").scope.get("user")
+
+    # Access an attribute inside the user object to force loading it in async contexts.
+    user.is_authenticated
```
bellini666, line 17: You will probably need to make `ruff` happy here by adding a `noqa` to a rule

Try to run pre-commit locally as there are some issues to be solved: https://results.pre-commit.ci/run/github/308695886/1696080990.Uxt5tuPWSO2Hu7uePKlx4g

strawberry_django/auth/utils.py
```diff
@@ -0,0 +1,23 @@
+from strawberry.types import Info
+
+
+def get_current_user(info: Info):
```
bellini666, line 4: Still missing the return annotation in here

I think it should be `Optional[AbstractUser]`

strawberry_django/auth/utils.py
bellini666, line 17: Also, `user` might be `None` in here
```suggestion
    if user is not None:
        user.is_authenticated
```

strawberry_django/auth/utils.py
```diff
@@ -0,0 +1,23 @@
...
+            user = info.context.get("request").scope.get("user")
+
+    # Access an attribute inside the user object to force loading it in async contexts.
+    user.is_authenticated
+
+    return user
+
+
+async def aget_current_user(info: Info):
+    return sync_to_async(get_current_user)(info)
```
bellini666, line 23: This needs to be imported

bellini666 (changes requested):

strawberry_django/auth/utils.py
sdobbelaere: importing `from django.contrib.auth.models import AbstractUser` or from django.contrib.auth.models import AbstractBaseUser won't work as the django application isn't fully loaded at this point.  

If you have a solution., happy to hear it.

docs/guide/types.md
```diff
@@ -103,12 +103,15 @@ You can use that `info` parameter to, for example,
 limit access to results based on the current user in the request:
 
 ```{.python title=types.py}
+from stawberry_django.auth.queries import get_current_user
```
bellini666, line 106: This needs to be updated to the new path

```suggestion
from stawberry_django.auth.utils import get_current_user
```

strawberry_django/auth/utils.py
bellini666: import it inside a `if TYPE_CHECKING:`

bellini666 (changes requested):
Almost there :)

strawberry_django/auth/utils.py
sdobbelaere: Ruff keeps kicking out TYPE_CHECKING.  

I added:

```python
from typing import Optional, TYPE_CHECKING

from asgiref.sync import sync_to_async
from strawberry.types import Info

if TYPE_CHECKING:
    from django.contrib.auth.models import AbstractBaseUser
```

after committing, it turns into:

```python
from typing import Optional

from asgiref.sync import sync_to_async
from django.contrib.auth.models import AbstractBaseUser
from strawberry.types import Info
```

Thoughts?

bellini666: You probably need to annotate it as a string, like use `"AbstractBaseUser"` instead of `AbstractBaseUser`

That way ruff will allow you to leave it inside the `if TYPE_CHECKING`.

sdobbelaere: Actually, I tried to run it before committing as well - and it just doesn't work. Django will not allow you to load the user-model outside of your Django-project.  Or at the very least before all apps are loaded, which causes similar issues.  

Perhaps you want to give it a shot yourself?  I've put as much time in this particular detail as I can.

bellini666: The solution was doing something like this:

image

But having said that, I fixed it for you and actually used our `UserType`, which is a union of `AbstractBaseUser` and `AnonymousUser`

bellini666 (approved):

bellini666:
@sdobbelaere I just solved the remaining pyright issues for you.

If the tests are passing fine, is there anything else missing from this or can it be merged?

docs/guide/subscriptions.md
```diff
@@ -1,6 +1,140 @@
...
+            graphiql=settings.DEBUG,
+            subscriptions_enabled=True,
+        ),
+    ),
+    ...
+]
+
+```
+
+Note, django-channels allows for a lot more complexity. Here we just cover the basic framework to get
```
bellini666, line 64: obs. pre-commit alex won't allow the word `just` to be used. It might need to be reworded here.

https://results.pre-commit.ci/run/github/308695886/1696869627.klCewzb2Qp6uMthM-uq5iw

bellini666 (changes requested):

docs/guide/subscriptions.md
sdobbelaere: Thanks for the help.  I've adjusted the wording.  
I did add a get_current_user override to allow for quick testing earlier today.  If you like what I did, let me know and I'll add it to the docs as well.

sdobbelaere:
> @sdobbelaere I just solved the remaining pyright issues for you.
> 
> If the tests are passing fine, is there anything else missing from this or can it be merged?

There is commit b0a464a455afe3ca3f6fd27048d0e12a5c5c7a08 which allows you to manually set a user in the context for testing, on the condition that you mark it as "is_test_mode". 

Why did I add this?  Running tests needs access to a user in my case, this allows to construct a simple context when running tests like so:

```python
from django.contrib.auth import get_user_model
from MyProject.schemas import schema

user = get_user_model().objects.last()
dummy_context = {'is_test_mode': True, 'user': user}

resp = schema.execute_sync(query=query, context_value=dummy_context)
```

 Happy to add it to the docs.

strawberry_django/auth/utils.py
```diff
@@ -0,0 +1,65 @@
...
+            user = info.context.get("request").consumer.scope["user"]
+        except AttributeError:
+            # websockets / subscriptions move scope inside of the request
+            try:
+                user = info.context.get("request").scope.get("user")
+            except AttributeError:
+                # When running tests, we must manually add a context.
+                # For now let's grab it here:
+                is_test_mode = info.context.get("is_test_mode")
+                if is_test_mode:
```
bellini666, line 33: I'm not a huge fan of this. Usually we would want to mock this or solve it in another way

What issue were you trying to fix with this code?

bellini666 (changes requested):

strawberry_django/auth/utils.py
bellini666: Oh, I just saw your comment in the PR. I'll reply there

bellini666:
> > @sdobbelaere I just solved the remaining pyright issues for you.
> > If the tests are passing fine, is there anything else missing from this or can it be merged?
> 
> Thank you for the help, much appreciated. The only thing left is commit [b0a464a](https://github.com/strawberry-graphql/strawberry-graphql-django/commit/b0a464a455afe3ca3f6fd27048d0e12a5c5c7a08) which allows you to manually set a user in the context for testing, on the condition that you mark it as "is_test_mode".
> 
> Why did I add this? Running tests needs access to a user in my case, this allows to construct a simple context when running tests like so:
> 
> ```python
> from django.contrib.auth import get_user_model
> from MyProject.schemas import schema
> 
> user = get_user_model().objects.last()
> dummy_context = {'is_test_mode': True, 'user': user}
> 
> resp = schema.execute_sync(query=query, context_value=dummy_context)
> ```
> 
> Happy to add it to the docs.

As I mentioned there, I don't like that approach very much, specially because we have a way for forcing a user into the request: https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/strawberry_django/test/client.py#L75

You can see it being used in here for example: https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/tests/test_permissions.py#L49

strawberry_django/auth/utils.py
sdobbelaere: I like the current approach much better indeed.  My addition was reverted, and the docs which were missing the `with` statement are fixed for whoever needs it after me.

bellini666 (changes requested):
Just a small nitpick and it should be good to merge :)

Thanks for taking the time to adjust everything I asked

PR #377, opened 2023-09-27
Support annotate parameter in field to allow ORM annotations

fjsj:
## Description

Adds support for ORM annotations as fields. Looks like this:
```python
class ProjectType(relay.Node):
    milestones_count: int = strawberry_django.field(annotate=Count("milestone"))  # here!
```

And this:
```python
@strawberry_django.type(Milestone)
class MilestoneType(relay.Node):
    @strawberry_django.field(
        annotate={
            "_my_bugs_count": lambda info: Count(
                "issue",
                filter=Q(
                    issue__issue_assignee__user_id=info.context.request.user.id,
                    issue__kind=Issue.Kind.BUG,
                ),
            ),
        },
    )
    def my_bugs_count(self) -> int:
        return self._my_bugs_count  # type: ignore
```

There are some shortcomings and implementation details that I discuss more in-depth in PR comments. The main one is this only works if `DjangoOptimizerExtension` is enabled. Although that doesn't look like a major issue to me, because custom prefetches also have the same limitation (see the existing `tests/test_optimizer.py::test_query_prefetch_with_callable`).

**As of Sept 27 2023, this PR is still missing docs. I'll wait for feedback on implementation and field "syntax" before making the docs.** I already added some tests, but perhaps we need more. I still also have to test manually a bit more.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [X] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -19,6 +19,7 @@ from typing import (
 from django.db import models
 from django.db.models import Prefetch
 from django.db.models.constants import LOOKUP_SEP
+from django.db.models.expressions import BaseExpression
 from django.db.models.fields.reverse_related import (
     ManyToManyRel,
     ManyToOneRel,
@@ -50,8 +51,11 @@ from .utils.inspect import (
     get_selections,
 )
 from .utils.typing import (
+    AnnotateCallable,
+    AnnotateType,
     PrefetchCallable,
     PrefetchType,
+    TypeOrMapping,
     TypeOrSequence,
     get_django_definition,
 )
@@ -72,6 +76,7 @@ __all__ = [
 _M = TypeVar("_M", bound=models.Model)
 
 _sentinel = object()
+_annotate_placeholder = "__annotated_placeholder__"
 _interfaces: defaultdict[
     Schema,
     dict[StrawberryObjectDefinition, list[StrawberryObjectDefinition]],
@@ -92,6 +97,8 @@ class OptimizerConfig:
             Enable `QuerySet.select_related` optimizations
         enable_prefetch_related:
             Enable `QuerySet.prefetch_related` optimizations
+        enable_annotate:
+            Enable `QuerySet.annotate` optimizations
         prefetch_custom_queryset:
             Use custom instead of _base_manager for prefetch querysets
 
@@ -100,6 +107,7 @@ class OptimizerConfig:
     enable_only: bool = dataclasses.field(default=True)
     enable_select_related: bool = dataclasses.field(default=True)
     enable_prefetch_related: bool = dataclasses.field(default=True)
+    enable_annotate: bool = dataclasses.field(default=True)
     prefetch_custom_queryset: bool = dataclasses.field(default=False)
 
 
@@ -121,14 +129,18 @@ class OptimizerStore:
     only: list[str] = dataclasses.field(default_factory=list)
     select_related: list[str] = dataclasses.field(default_factory=list)
     prefetch_related: list[PrefetchType] = dataclasses.field(default_factory=list)
+    annotate: dict[str, AnnotateType] = dataclasses.field(default_factory=dict)
 
     def __bool__(self):
-        return any([self.only, self.select_related, self.prefetch_related])
+        return any(
+            [self.only, self.select_related, self.prefetch_related, self.annotate],
+        )
 
     def __ior__(self, other: OptimizerStore):
         self.only.extend(other.only)
         self.select_related.extend(other.select_related)
         self.prefetch_related.extend(other.prefetch_related)
+        self.annotate.update(other.annotate)
         return self
 
     def __or__(self, other: OptimizerStore):
@@ -139,6 +151,7 @@ class OptimizerStore:
             only=self.only[:],
             select_related=self.select_related[:],
             prefetch_related=self.prefetch_related[:],
+            annotate=self.annotate.copy(),
         )
 
     @classmethod
@@ -148,6 +161,7 @@ class OptimizerStore:
         only: TypeOrSequence[str] | None = None,
         select_related: TypeOrSequence[str] | None = None,
         prefetch_related: TypeOrSequence[PrefetchType] | None = None,
+        annotate: TypeOrMapping[AnnotateType] | None = None,
     ):
         return cls(
             only=[only] if isinstance(only, str) else list(only or []),
@@ -160,6 +174,13 @@ class OptimizerStore:
                 [prefetch_related]
                 if isinstance(prefetch_related, (str, Prefetch, Callable))
                 else list(prefetch_related or [])
+            ),
+            annotate=(
+                # placeholder here,
+                # because field name is evaluated later on .annotate call:
+                {_annotate_placeholder: annotate}
+                if isinstance(annotate, (BaseExpression, Callable))
+                else dict(annotate or {})
             ),
         )
 
@@ -178,10 +199,18 @@ class OptimizerStore:
             else:  # pragma:nocover
                 assert_never(p)
 
+        annotate = {}
+        for k, v in self.annotate.items():
+            if isinstance(v, Callable):
+                assert_type(v, AnnotateCallable)
+                v = v(info)  # noqa: PLW2901
+            annotate[f"{prefix}{LOOKUP_SEP}{k}"] = v
+
         return self.__class__(
             only=[f"{prefix}{LOOKUP_SEP}{i}" for i in self.only],
             select_related=[f"{prefix}{LOOKUP_SEP}{i}" for i in self.select_related],
             prefetch_related=prefetch_related,
+            annotate=annotate,
         )
 
     def apply(
@@ -252,6 +281,15 @@ class OptimizerStore:
 
         if config.enable_only and self.only:
             qs = qs.only(*self.only)
+
+        if config.enable_annotate and self.annotate:
+            to_annotate = {}
+            for k, v in self.annotate.items():
+                if isinstance(v, Callable):
+                    assert_type(v, AnnotateCallable)
+                    v = v(info)  # noqa: PLW2901
+                to_annotate[k] = v
+            qs = qs.annotate(**to_annotate)
 
         return qs
 
@@ -399,6 +437,18 @@ def _get_model_hints(
         # Add annotations from the field if they exist
         field_store = getattr(field, "store", None)
         if field_store is not None:
+            if (
+                len(field_store.annotate) == 1
+                and _annotate_placeholder in field_store.annotate
+            ):
+                # This is a special case where we need to update the field name,
+                # because when field_store was created on __init__,
+                # the field name wasn't available.
+                # This allows for annotate expressions to be declared as:
+                # Instead of the more redundant:
+                field_store.annotate = {
+                    field.name: field_store.annotate[_annotate_placeholder],
+                }
             store |= (
                 field_store.with_prefix(prefix, info=info) if prefix else field_store
             )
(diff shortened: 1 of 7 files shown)


strawberry_django/utils/typing.py
```diff
@@ -37,10 +39,13 @@
 _Type = TypeVar("_Type", bound="StrawberryType | type")
 
 TypeOrSequence: TypeAlias = Union[_T, Sequence[_T]]
+TypeOrMapping: TypeAlias = Union[_T, Mapping[str, _T]]
 TypeOrIterable: TypeAlias = Union[_T, Iterable[_T]]
 UserType: TypeAlias = Union["AbstractBaseUser", "AnonymousUser"]
 PrefetchCallable: TypeAlias = Callable[[GraphQLResolveInfo], Prefetch]
 PrefetchType: TypeAlias = Union[str, Prefetch, PrefetchCallable]
+AnnotateCallable: TypeAlias = Callable[[GraphQLResolveInfo], BaseExpression]
+AnnotateType: TypeAlias = Union[BaseExpression, AnnotateCallable]
```
fjsj, line 48: I have to double-check if `BaseExpression` is the best or `Expression` is better. `BaseExpression` also allows `Subquery`, and I believe I have examples in some projects were we had to use `annotate` with `Subquery` because other annotation types didn't cover our use cases. I'll check.

strawberry_django/optimizer.py
```diff
@@ -768,10 +822,11 @@ def optimize(
             return qs
 
         config = OptimizerConfig(
-            enable_only=self.enable_ony
+            enable_only=self.enable_only
```
fjsj, line 825: Typo fix (I guess?).

strawberry_django/optimizer.py
```diff
@@ -399,6 +437,18 @@ def _get_model_hints(
...
+                and _annotate_placeholder in field_store.annotate
+            ):
+                # This is a special case where we need to update the field name,
+                # because when field_store was created on __init__,
+                # the field name wasn't available.
+                # This allows for annotate expressions to be declared as:
+                # Instead of the more redundant:
+                field_store.annotate = {
+                    field.name: field_store.annotate[_annotate_placeholder],
+                }
```
fjsj, line 453: I don't the necessity to use `_annotate_placeholder`. So please check if my comment here really makes sense. I think the `field` object at instantiation time cannot know about what attr name it has inside the parent `Type` class, because there's no metaclass in `Type`. In other libraries (such as Django REST Framework), the fields knows its attr in the parent class thanks to metaclasses.

strawberry_django/utils/typing.py
```diff
@@ -37,10 +39,13 @@
 _Type = TypeVar("_Type", bound="StrawberryType | type")
 
 TypeOrSequence: TypeAlias = Union[_T, Sequence[_T]]
+TypeOrMapping: TypeAlias = Union[_T, Mapping[str, _T]]
```
fjsj, line 42: I don't love that `annotate` needs a dict instead of a list, but I see no other way of specifying the annotation name in a case like this:
@strawberry_django.type(Milestone)
class MilestoneType(relay.Node):
    @strawberry_django.field(
        annotate={
            "_issues_count": lambda info: Count("issue")
        },
    )
    def issues_count(self) -> int:
        return self._issues_count  # type: ignore

tests/django_settings.py
```diff
@@ -11,6 +11,8 @@
 
 DEBUG = True
 SECRET_KEY = 1
+USE_TZ = True
+TIME_ZONE = "UTC"
```
fjsj, line 15: Had to add this to avoid test warnings. Those are the Django 5.0 defaults.

tests/projects/schema.py
```diff
@@ -83,6 +92,13 @@ class ProjectType(relay.Node):
     name: strawberry.auto
     due_date: strawberry.auto
     milestones: List["MilestoneType"]
+    milestones_count: int = strawberry_django.field(annotate=Count("milestone"))
+    is_delayed: bool = strawberry_django.field(
+        annotate=ExpressionWrapper(
+            Q(due_date__lt=Now()),
+            output_field=BooleanField(),
+        ),
```
fjsj, line 100: Complex annotation example.

tests/projects/schema.py
```diff
@@ -134,6 +150,20 @@ class MilestoneType(relay.Node):
...
+                "issue",
+                filter=Q(
+                    issue__issue_assignee__user_id=info.context.request.user.id,
+                    issue__kind=Issue.Kind.BUG,
+                ),
+            ),
+        },
+    )
+    def my_bugs_count(self) -> int:
+        return self._my_bugs_count  # type: ignore
```
fjsj, line 165: Complex callable annotation example. Considers current user.

tests/test_optimizer.py
```diff
@@ -694,3 +696,118 @@ def test_query_nested_fragments(db, gql_client: GraphQLTestClient):
...
+                    res = gql_client.query(query, {"node_id": e["id"]})
+                    assert res.data == {"project": e}
+            else:
+                # myBugsCount requires the optimizer to be turned on
+                res = gql_client.query(
+                    query,
+                    {"node_id": e["id"]},
+                    asserts_errors=False,
+                )
+                assert res.errors
```
fjsj, line 813: Please LMK about additional tests I should implement.

tests/test_optimizer.py
```diff
@@ -694,3 +696,118 @@ def test_query_nested_fragments(db, gql_client: GraphQLTestClient):
...
+            res = gql_client.query(
+                query,
+                {"node_id": e["id"]},
+                asserts_errors=False,
+            )
+            assert res.errors
+
+
+@pytest.mark.django_db(transaction=True)
+def test_query_annotate_with_callable(db, gql_client: GraphQLTestClient):
```
fjsj, line 755: I'm not sure if this test is testing the case of prefetch + annotate, i.e., the annotate is called inside the custom queryset of a `prefetch_related`. This is something we should test.

aprams:
Would it make sense to split the annotate functionality for `ModelProperty`, `PrefetchInspector` and potentially types to a separate PR to avoid complexity creep here?

Regarding type support: The current field support supports also works on the type's queryset itself, correct? I can see how this might help for downstream tasks on the objects which are not explicit fields on the type.

fjsj:
> Would it make sense to split the annotate functionality for ModelProperty, PrefetchInspector and potentially types to a separate PR to avoid complexity creep here?

IMHO, yes, as long as nothing that is expected to work in regular field actually doesn't.

> Regarding type support: The current field support supports also works on the type's queryset itself, correct?

Yes, since `type` already supports `select_related` and `prefetch_related` and pass this down to `field` init, I guess we can just do the same here for `annotate`.

tests/test_optimizer.py
```diff
@@ -694,3 +696,121 @@ def test_query_nested_fragments(db, gql_client: GraphQLTestClient):
...
+            # thus they will be counted
+            for i in IssueFactory.create_batch(4, milestone=m, kind=Issue.Kind.BUG):
+                Assignee.objects.create(user=user, issue=i)
+                m_res["myBugsCount"] += 1
+
+    assert len(expected) == 2
+    for e in expected:
+        with gql_client.login(user):
+            if DjangoOptimizerExtension.enabled.get():
+                with assert_num_queries(4):
```
bellini666, line 806: I haven't tried the code, but shouldn't this be 3 queries?

- 1 for the logged user
- 1 to retrieve the project 
- 1 to prefetch the milestones, with the `bugsCount` aggregate in it

tests/test_optimizer.py
bellini666: Of course more tests are always nice, but I think you got basically all of the core features covered nicely :)

tests/test_type.py
```diff
@@ -82,3 +82,29 @@ def some_field(self, my_input: SomeModelInput) -> str: ...
...
+        SomeModel,
+        only=["name", "other", "other_name"],
+        select_related=["other"],
+        prefetch_related=["other"],
+        annotate={"other_name": models.F("other__name")},
+    )
+    class SomeModelType:
+        name: str
+
+    store = SomeModelType.__strawberry_django_definition__.store
```
bellini666, line 105: nit

obs. you need to `from strawberry_django.utils.typing import get_django_definition` as well

```suggestion
    store = get_django_definition(SomeModelType, strict=True), store
```

strawberry_django/optimizer.py
bellini666: Getting back to this comment after checking the whole PR: Maybe we don't need this, and instead we can just keep the expression in `annotate` directly, and this if would be `if isinstance(field_store.annotate, BaseExpression)` or even `if field_store.annotate is not None and not isinstance(field_store.annotate, dict)`?

I mean, it would make sense to use `_annotate_placeholder` if we could have more than 1 annotation together, but I'm assuming that it will never be the case. Or am I missing something?

bellini666 (changes requested):
This looks amazing!!

The -1 is for a couple of nitpicks

strawberry_django/optimizer.py
fjsj: I guess the problem of not using `_annotate_placeholder` is that typing will get more confusing in the `OptimizerStore`. But I'll check if this is doable.

tests/test_optimizer.py
fjsj: Here are the queries:

```
(Pdb) ctx.connection.queries[0]['sql']
'SELECT "django_session"."session_key", "django_session"."session_data", "django_session"."expire_date" FROM "django_session" WHERE ("django_session"."expire_date" > \'2023-09-28 17:19:11.112006\' AND "django_session"."session_key" = \'xepa46zwe72z635b6nby9jlvdl4zl7cl\') LIMIT 21'
(Pdb) ctx.connection.queries[1]['sql']
'SELECT "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined" FROM "auth_user" WHERE "auth_user"."id" = 1 LIMIT 21'
(Pdb) ctx.connection.queries[2]['sql']
'SELECT "projects_project"."id", "projects_project"."name" FROM "projects_project" WHERE "projects_project"."id" = 1 ORDER BY "projects_project"."id" ASC LIMIT 1'
(Pdb) ctx.connection.queries[3]['sql']
'SELECT "projects_milestone"."id", "projects_milestone"."name", "projects_milestone"."project_id", COUNT("projects_issue"."id") FILTER (WHERE ("projects_assignee"."user_id" = 1 AND "projects_issue"."kind" = \'b\')) AS "_my_bugs_count" FROM "projects_milestone" LEFT OUTER JOIN "projects_issue" ON ("projects_milestone"."id" = "projects_issue"."milestone_id") LEFT OUTER JOIN "projects_assignee" ON ("projects_issue"."id" = "projects_assignee"."issue_id") WHERE "projects_milestone"."project_id" IN (1) GROUP BY "projects_milestone"."id", "projects_milestone"."name", "projects_milestone"."project_id"'
```

The second query is for the user, which happens also on `test_query_prefetch_with_callable`. So I guess this is fine?

bellini666: Oh, there's an extra query for the `session`

Yeah, it is fine :)

strawberry_django/optimizer.py
fjsj: If we make `annotate: dict[str, AnnotateType] | AnnotateType = dataclasses.field(default_factory=dict)` inside `OptimizerStore`, this conditional logic will need handling in `__ior__`, `copy`, `with_prefix` and `apply` methods. Right now, the placeholder logic is only in `with_hints`. Therefore, to avoid spreading this logic, I guess it's best to leave as is.

bellini666: Yes, forgot about that. I agree with you, let's leave as it for now :)

bellini666 (approved):
LGTM 👍🏼

bellini666:
The typing issue is due to pyright's yesterday release.

Going to merge this as is and fix the pyright issue myself, as it is not this PRs fault.

Issue #379, opened 2023-09-28
DjangoOptimizer prefetch_related on nested objects modifies prefetch


aprams:
## Describe the Bug

When using the DjangoOptimizer extension and adding prefetching to type's field, the field's store gets corrupted on each query.

Example:

```
@strawberry_django.type(
    models.MyModel,
)
class MyModelType(RandomPkInterface):
    main_identifier: str = strawberry_django.field(
        field_name="main_identifier",
        prefetch_related=[
            Prefetch(
                "identifiers",
                queryset=Identifiers.objects.all(),
                to_attr="identifiers"
            ),
        ]
    )
```

Here, I have a custom property on my model that gets the main identifier from all the identifiers, so I want to do a prefetch here to avoid n + 1 issues. 

The following query will now cause issues, though:

```
query {
    myParentObject {
      myModel {
        mainIdentifier
      }
    }
}
```

Calling this query once works fine, however, the second request will lead to this error:

`Cannot find 'my_parent' on MyModel object, 'my_parent_model__my_model__my_parent_model__my_model__identifiers' is an invalid parameter to prefetch_related()`

Further subsequent requests each add a `my_parent_model__my_model__` to the beginning of the invalid prefetch. 

My debugging lead me to the conclusion, that the culprit is line 444 in strawberry-django/optimizer.py :
`store |= f_store.with_prefix(path, info=info)`

In `f_store.with_prefix` the actual Prefetch object will prepend the path. This is however a by reference pass of the actual Prefetch object of the field, as this is never copied along the way. This way, after the first call, the prefetch stacks up the prefixes and pollutes/corrupts the field for subsequent calls.

## System Information

 - Operating system: Mac OS
 - Strawberry version (if applicable):
 - strawberry-graphql==0.209.0
 - strawberry-graphql-django==0.17.0

## Additional Context

My suggestion would be to have a copy of the prefetch object / the field's store when getting it in line 400 in optimizer.py. This would avoid downstream issues. 

If you agree with this approach, I can also attempt fixing it as my first issue.

PR #380, opened 2023-09-30
Fix: DjangoOptimizerExtension corrupts nested objects' fields' prefetch objects

aprams:
## Description

This PR aims to resolve issue #379 .

I included a reproducible test and a suggested fix, where the `Prefetch` object gets deepcopied to avoid the side effects from `add_prefix`. 

Considerations:
- There were multiple ways to do this, including copying the OptimizerStore at a higher level in the execution stack, but this should be the least invasive one 
- I used deepcopy now as Prefetch didn't offer an easier way of copying the object that I know of, I'm happy about suggestions on improving this

I'm also very open for improving the test case included. If you know a better way on how to do it without the custom type setup, please let me know.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #379 

## Checklist

I love the work you do here, thanks a lot for the really awesome work! ❤️

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 import contextlib
 import contextvars
+import copy
 import dataclasses
 import itertools
 from collections import defaultdict
-from copy import deepcopy
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -198,7 +198,7 @@ class OptimizerStore:
                 prefetch_related.append(f"{prefix}{LOOKUP_SEP}{p}")
             elif isinstance(p, Prefetch):
                 # add_prefix modifies the field's prefetch object, so we copy it before
-                p_copy = deepcopy(p)
+                p_copy = copy.copy(p)
                 p_copy.add_prefix(prefix)
                 prefetch_related.append(p_copy)
             else:  # pragma:nocover
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -5,6 +5,7 @@ import contextvars
 import dataclasses
 import itertools
 from collections import defaultdict
+from copy import deepcopy
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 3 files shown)


strawberry_django/optimizer.py
```diff
@@ -196,8 +197,10 @@ def with_prefix(self, prefix: str, *, info: GraphQLResolveInfo):
             if isinstance(p, str):
                 prefetch_related.append(f"{prefix}{LOOKUP_SEP}{p}")
             elif isinstance(p, Prefetch):
-                p.add_prefix(prefix)
-                prefetch_related.append(p)
+                # add_prefix modifies the field's prefetch object, so we copy it before
+                p_copy = deepcopy(p)
```
bellini666, line 201: Hrm, the only solution I can think right now would be to create a `Prefetch` object by hand, but then we would need to update it in case it gets changed in django.

The only question here would be: Any reason to use `deepcopy` instead of `copy`? Unless I'm missing something, a copy should be enough to fix the issue here

tests/test_optimizer.py
```diff
@@ -814,3 +818,72 @@ def test_query_annotate_with_callable(db, gql_client: GraphQLTestClient):
...
+
+@pytest.mark.django_db(transaction=True)
+def test_user_query_with_prefetch():
+    @strawberry_django.type(
+        Project,
+    )
+    class ProjectTypeWithPrefetch:
+        @strawberry_django.field(
+            prefetch_related=[
+                Prefetch(
```
bellini666, line 831: I can see the issue here =P

I usually define `Prefetch` objects as callables (i.e. `prefetch_related=[lambda info: Prefetch(...)]`), meaning each call will have its own prefetch, so I didn't notice the possibility of this issue

Nice catch!

bellini666 (changes requested):
Thanks for this! :)

The -1 is due to the copy vs deepcopy comment I've made. Solving it and we should be good to merge

strawberry_django/optimizer.py
aprams: Yeah, I tried that and it's also not super trivial with the internal `levels` and protected variables. Kind of a messy class.

I don't really have a good reason for deepcopy, fair point, I changed it to `copy.copy` and the test still works :)

tests/test_optimizer.py
aprams: He, didn't think of that, I just stumbled upon it by using the same syntax from the test, defining the `Prefetch` objects directly.

bellini666 (approved):

PR #381, opened 2023-09-30
feat: deprecate nSomething in favor of using NOT

### LINKED ISSUE #361: Deprecate `nSomething` in favour of using `NOT`
Deprecate via GraphQL, and via a code warning, we can use `NOT` instead of duplicating filters, similar to what we do for `AND` and `OR` :)

bellini666:
Fix #361


patrick91 (approved):
awesome!

PR #393, opened 2023-10-13
compatibility ASGI/websockets get_request, login and logout

sdobbelaere:
The context object can turn into a dict when using ASGI.  This pull requests accounts for both ASGI and WSGI conditions and replaces the use of `info.context.request` with `get_request(info)`

### DIFF
--- a/strawberry_django/auth/mutations.py
+++ b/strawberry_django/auth/mutations.py
@@ -40,7 +40,7 @@ def resolve_login(info: Info, username: str, password: str) -> AbstractBaseUser 
             # request is actually channelsrequest
             try:
                 scope = request.consumer.scope  # type: ignore
-                async_to_sync(channels_auth.login)(scope, user)
+                async_to_sync(channels_auth.login)(scope, user)  # type: ignore
                 # According to channels docs you must save the session
                 scope["session"].save()
             except (AttributeError, NameError):
@@ -63,7 +63,7 @@ def resolve_logout(info: Info) -> bool:
     except AttributeError:
         try:
             scope = request.consumer.scope  # type: ignore
-            async_to_sync(channels_auth.logout)(scope)
+            async_to_sync(channels_auth.logout)(scope)  # type: ignore
         except (AttributeError, NameError):
             # When Django-channels is not installed,
             # this code will be non-existing
--- a/strawberry_django/auth/mutations.py
+++ b/strawberry_django/auth/mutations.py
@@ -11,6 +11,7 @@ from strawberry_django.mutations import mutations, resolvers
 from strawberry_django.mutations.fields import DjangoCreateMutation
 from strawberry_django.optimizer import DjangoOptimizerExtension
 from strawberry_django.resolvers import django_resolver
+from strawberry_django.utils.requests import get_request
 
 if TYPE_CHECKING:
     from django.contrib.auth.base_user import AbstractBaseUser
@@ -19,18 +20,20 @@ if TYPE_CHECKING:
 
 @django_resolver
 def resolve_login(info: Info, username: str, password: str) -> AbstractBaseUser | None:
-    request = info.context.request
+    request = get_request(info)
     user = auth.authenticate(request, username=username, password=password)
+
     if user is not None:
         auth.login(request, user)
         return user
+
     auth.logout(request)
     return None
 
 
 @django_resolver
 def resolve_logout(info: Info) -> bool:
-    request = info.context.request
+    request = get_request(info)
     ret = request.user.is_authenticated
     auth.logout(request)
     return ret
--- a/strawberry_django/auth/mutations.py
+++ b/strawberry_django/auth/mutations.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
+import contextlib
 import functools
 from typing import TYPE_CHECKING, Any, Type, cast
 
 import strawberry
 from asgiref.sync import async_to_sync
-from channels import auth as channels_auth
 from django.contrib import auth
 from django.contrib.auth.password_validation import validate_password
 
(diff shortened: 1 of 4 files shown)


strawberry_django/auth/utils.py
```diff
@@ -16,15 +17,17 @@ def get_current_user(info: Info, *, strict: bool = False) -> Optional[UserType]:
...
+    request = get_request(info)
+
     try:
-        user = info.context.request.user
+        user = request.user
     except AttributeError:
         try:
             # queries/mutations in ASGI move the user into consumer scope
-            user = info.context.get("request").consumer.scope["user"]
+            user = request.consumer.scope["user"]
```
bellini666, line 27: Probably need a `# type: ignore` in here

strawberry_django/auth/utils.py
```diff
@@ -16,15 +17,17 @@ def get_current_user(info: Info, *, strict: bool = False) -> Optional[UserType]:
...
         except AttributeError:
             # websockets / subscriptions move scope inside of the request
-            user = info.context.get("request").scope.get("user")
+            user = request.scope.get("user")
```
bellini666, line 30: Ditto

bellini666 (changes requested):
Looks good! :)

Just need to fix the pyright tests and we should be good to merge

sdobbelaere:
It would seem that login and logout mutations were also non compatible with ASGI/websockets. This was addressed by using the channels authentication rather then the default django.

sdobbelaere:
It would seem that login and logout mutations were also non compatible with ASGI/websockets. This was addressed by using the channels authentication rather then the default django.

sdobbelaere:
@bellini666 Not sure if you need anything else from mee on this pull-request?  All tests seem to be passing.

strawberry_django/auth/mutations.py
```diff
@@ -19,20 +23,39 @@
...
+            # ASGI in combo with websockets needs the channels login functionality.
+            # to ensure we're talking about channels, let's veriy that our
+            # request is actually channelsrequest
+            scope = request.consumer.scope  # type: ignore
+            async_to_sync(channels_auth.login)(scope, user)
+            # According to channels docs you must save the session
+            scope["session"].save()
         return user
+
+    # FIXME: Why call logout inside login??
```
bellini666, line 42: Indeed... I don't use this mutation in my projects, and I'm not the author of it.

IMO this seems wrong. Do you want to fix this `FIXME` by removing the `logout` call?

bellini666 (approved):
Sorry for not replying earlier to this.

I was out during last week for the djangocon event (I was a speaker there https://2023.djangocon.us/talks/building-high-performance-type-safe-graphql-apis-with-strawberry-and-django/).

I made a small suggestion regarding the `logout` that you pointed out. If you want to fix it, I'll merge it after. Otherwise just tell me and this is fine to be merged already

strawberry_django/auth/mutations.py
sdobbelaere: @bellini666 I'll have a look at it, as this logout doesnt seem to actually do anything - which makes me suspect the actual logout mutation might not work.  I would appreciate some thoughts on #394 though.

sdobbelaere:
> Sorry for not replying earlier to this.
> 
No problem :)

> I was out during last week for the djangocon event (I was a speaker there https://2023.djangocon.us/talks/building-high-performance-type-safe-graphql-apis-with-strawberry-and-django/).
> 
Nice!  Some publicity for the library is a good idea indeed!

> I made a small suggestion regarding the `logout` that you pointed out. If you want to fix it, I'll merge it after. Otherwise just tell me and this is fine to be merged already

Yes, I'll take a look.  I hope it's a detail, but I fear that logout function may not work as advertised otherwise the login wouldn't work - which it does.

strawberry_django/auth/mutations.py
sdobbelaere: Seems to be fine to remove.  Logout mutation itself works fine.

strawberry_django/auth/mutations.py
```diff
@@ -4,13 +4,17 @@
 from typing import TYPE_CHECKING, Any, Type, cast
 
 import strawberry
+from asgiref.sync import async_to_sync
+from channels import auth as channels_auth
```
bellini666, line 8: Just a small thing here. Since `channels` is an optional dependency, trying to import it here directly will raise an `ImportError` for anyone not using it.

We should import this inside a `try: except ImportError` and only try to use it when available

bellini666 (changes requested):
A small important thing that I didn't notice in my previous review =P

strawberry_django/auth/mutations.py
sdobbelaere: Indeed, an important thing. Thx for noticing.
Ruff did force me to use another approach, but I've just pushed this change

sdobbelaere:
All done. Ready for merge from my end @bellini666

strawberry_django/auth/mutations.py
```diff
@@ -1,16 +1,25 @@
...
 from strawberry_django.mutations import mutations, resolvers
 from strawberry_django.mutations.fields import DjangoCreateMutation
 from strawberry_django.optimizer import DjangoOptimizerExtension
 from strawberry_django.resolvers import django_resolver
+from strawberry_django.utils.requests import get_request
+
+with contextlib.suppress(ModuleNotFoundError):
+    # Django-channels is not always used/intalled,
+    # therefore it shouldn't be it a hard requirement.
+    from channels import auth as channels_auth
```
bellini666, line 22: Probably use `ImportError` and a simple `try`/`except` to assign `channels_auto = None` on errors, otherwise the name will not be defined below

bellini666 (changes requested):

strawberry_django/auth/mutations.py
sdobbelaere: hmm, That was my original approach, but then ruff complains.

bellini666: > hmm, That was my original approach, but then ruff complains.

You can `# noqa: XYZ` on what ruff complains in that case =P

Search for `ImportError` in the repo, there are some similar approaches in it

sdobbelaere: Interesting that setting the missing import to None seems to get ignored by ruff.  

So I picked `ModuleNotFoundError` over `ImportError` since in this case we're talking about a package not being installed rather then struggling to import https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError

bellini666 (approved):

Issue #388, opened 2023-10-10
Mutations on proxy models issue


sdobbelaere:
It would seem that mutations on proxy-models aren't working in my setup. 
Can you see the mistake in my setup?

A simplified, reproducible example with details included:

Managers:
```python
from django.db.models import Queryset, Manager

class ClientQueryset(Queryset):
    def create(self, **kwargs):
        kwargs.update({'is_client': True})
        return super().create(**kwargs)

class ClientManager(Manager):
    def get_queryset(self):
        return ClientQueryset(self.model, using=self._db).filter(is_client=True)
````

Models:
```python
from django.db import models
from .managers import ClientManger

class Company(models.Model):
    name = models.CharField(max_length=100)
    is_client = models.BooleanField(default=False)

class Client(Company):
    objects = ClientManager()
    class Meta:
        proxy = True
```

Based on this setup, I'm perfectly capable of creating the client objects in the classic ORM way:

```python
from contacts.models import Client

client = Client.objects.create(name='client orm')
print(client.is_client, client.name)

```

Next up, let's setup the strawberry mutations.

schema:
```python
from strawberry import relay, auto
import strawberry
import strawberry_django
from typing import List

from .models import Company, Client


@strawberry_django.filter(Company)
class CompanyFilter:
    id: auto
    name: auto

@strawberry_django.filter(Client)
class ClientFilter:
    id: auto
    name: auto


@strawberry_django.input(Company, fields="__all__")
class CompanyInput:
    pass


@strawberry_django.partial(Company, fields="__all__")
class CompanyPartialInput(strawberry_django.NodeInput):
    pass


@strawberry_django.input(Client, fields="__all__")
class ClientInput:
    pass


@strawberry_django.partial(Client, fields="__all__")
class ClientPartialInput(strawberry_django.NodeInput):
    pass


@strawberry_django.order(Company)
class CompanyOrder:
    name: auto


@strawberry_django.order(Client)
class ClientOrder:
    name: auto


@strawberry_django.type(Company, filters=CompanyFilter, order=CompanyOrder, pagination=True, fields='__all__')
class CompanyType(strawberry.relay.Node):
    pass


@strawberry_django.type(Client, filters=ClientFilter, order=ClientOrder, pagination=True, fields='__all__')
class ClientType(strawberry.relay.Node):
    pass


@strawberry.type
class Mutation:
    create_company: CompanyType = strawberry_django.mutations.create(CompanyInput)
[...]

PR #394, opened 2023-10-14
Refactor create method to ensure proxy-model compatibility

sdobbelaere:
As discussed in issue #388 proxy-models with manager overrides were not respected. This pull-request attempts to rectify that. 

In short, the create() method has been refactored to no longer use update() - but instead leverage the ORM Manager functionality rather then the naive approach of manually setting an instance and calling save() at the end.

There are a few things that are worth considering:
- pre-save-hook is questionable - now being run after the create as there is no instance to run it on before creation.  Line 301
- refresh_from_db() really needed? The update method seemed to do something similar. Unsure why. Line 324

Feedback, thought and improvements are welcomed.

EDIT:
- [x] Fix create mutations to respect objects.create approach
- [x] Ensure pre_save_hook as run in the traditional method


sdobbelaere:
Looks like some a test related to user registration is failing with this new approach.  any thoughts?

sdobbelaere:
Independently of this PR, the proxy models seem to be giving me one more issue.

They are returning the wrong parent type.

Following the example in #388 

When querying for the object via it's id with node()

the `companyType` is returned instead of `clientType`.  This works well in with `connection()` listings, but not with `node()`.

Example:
```graphql
query client{
  client(id:"Q29tcGFueVR5cGU6MTE="){
    id
    name
    isCustomer
  }
}
```
yields unsuccessfully
```json
{
  "data": null,
  "errors": [
    {
      "message": "Expected value of type 'ClientType' but got: <Company instance>.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "customer"
      ]
    }
  ]
}
```

sdobbelaere:
Some notes on the type issue.

Let's use this query:
```graphql
            query customer($id: GlobalID!) {
                customer(id: $id) {
                    id
                    name
                    __typename
                    isCustomer
                }
            }
```

Running that query with the strawberry-test-client correctly yields:
```
{
'customer': 
    {
    'id': 'Q3VzdG9tZXJUeXBlOjEz', 
     '__typename': 'CustomerType', 
    'isCustomer': True
    }
}
```
Yet when running it via Daphne in graphiql, the result is incorrectly a company-instance rather then the customer-instance.
```
{
  "data": null,
  "errors": [
    {
      "message": "Expected value of type 'CustomerType' but got: <Company instance>.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "customer"
      ]
    }
  ]
}
```

Some help on this PR would be greatly appreciated.

strawberry_django/utils/query.py
```diff
@@ -101,7 +101,7 @@ def filter_for_user_q(
         except KeyError:  # pragma:nocover
             # If we are not running async, retrieve it
             ctype = (
-                ContentType.objects.get_for_model(model)
+                ContentType.objects.get_for_model(model, for_concrete_model=False)
```
bellini666, line 104: I'm not very used to proxy models, but does this have any bad side effect for non-proxy ones?

strawberry_django/mutations/resolvers.py
```diff
@@ -351,7 +448,8 @@ def update(
         for field, value in m2m:
             update_m2m(info, instance, field, value)
 
-        retval.append(instance.__class__._default_manager.get(pk=instance.pk))
+        instance.refresh_from_db()
```
bellini666, line 451: If I remember correctly, the reason I decided to retrieve the object again instead of refreshing it from the db was to make sure that one-to-many and many-to-many relations would be refetched as well.

Wondering if we should do `instance._prefetched_objects_cache.clear()` and `instance._state.fields_cache.clear()` as well. Maybe it is not required though

strawberry_django/mutations/resolvers.py
```diff
@@ -216,19 +216,116 @@ def create(
     full_clean: bool | FullCleanOptions = True,
     pre_save_hook: Callable[[_M], None] | None = None,
 ):
+    # Before creating your instance, verify this is not a bulk create
+    # if so, add them one by one. Othewise, get to work.
     if isinstance(data, list):
         return [create(info, model, d, full_clean=full_clean) for d in data]
 
+    # Also, the appraoch below will use the manager to create the instnace
```
bellini666, line 224:
```suggestion
    # Also, the approach below will use the manager to create the instance
```

strawberry_django/mutations/resolvers.py
```diff
@@ -216,19 +216,116 @@ def create(
...
+    # Now that the instance has been created, go and assign
+    # files and many2many fields.
+    for file_field, value in files:
+        file_field.save_form_data(instance, value)
+
+    for field, value in m2m:
+        update_m2m(info, instance, field, value)
+
+    # FIXME: Should you really need to refresh the object from db?
+    instance.refresh_from_db()
```
bellini666, line 326: In this case, I don't think so.

It is mostly useful for update to make sure we are reloading the latest version of the object, but for create we should be fine without it

strawberry_django/mutations/resolvers.py
```diff
@@ -216,19 +216,116 @@ def create(
...
+        ]
+    ] = []
+    m2m: list[tuple[ManyToManyField | ForeignObjectRel, Any]] = []
+    create_kwargs = {}
+
     if dataclasses.is_dataclass(data):
-        data = vars(cast(object, data))
+        data = vars(data)
+
+    for name, value in data.items():
```
bellini666, line 254: This is mostly duplicating the code from `update` into create.

I do understand why, but can we maybe move some of the duplicated code to a separated utility function and call it in both places?

bellini666 (changes requested):
Thanks for the PR and sorry for taking long to review it. I was at the djangocon conference last week and am just now getting back to this.

Made a couple of comments here, but from what I understand there's still some things missing, correct?

bellini666:
> It seems to be very hard to track down, but is actually obstructing the proxy-models to work. From what I can tell, it seems to be related to the dynamic way models are fetched.
> 
> I've tried to track down the base source of where an object is fetched and how the query sets are determined. But seem to be chasing my tail with that. Also re-fetching the object doesn't seem to resolve the issue.
> 
> Perhaps someone with more in-depth knowledge of the library can figure this out?

`node()` will be calling `resolve_node()`/`resolve_nodes()` from the type, which for django types get injected here when not defined: https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/strawberry_django/type.py#L206

Feel free to ping me on discord in case you want to discuss more about this. As I mentioned, I have not used proxy models so far, so I'm not really sure what adjustments that needs to be done here to make them work properly, but hopefully it should be a matter of details.

strawberry_django/utils/query.py
sdobbelaere: Shouldn't have any effect.  It merely allows for fetching of proxy-model content-types.  https://docs.djangoproject.com/en/4.2/ref/contrib/contenttypes/#django.contrib.contenttypes.models.ContentTypeManager.get_for_model

strawberry_django/mutations/resolvers.py
sdobbelaere: Shouldn't refresh_from_db() do that as well?  
I see in other projects, that if I have out-of-date info through FK and M2M - adding refresh_from_db() resolves that without having to refetch.  

There is in my experience only one case where I need to refetch in a similar way to here.  That's in combination with Django-polymorphic. That that's due to the app bungling up the content types.  That's not the case here.

strawberry_django/mutations/resolvers.py
sdobbelaere: ok, removed these lines.

strawberry_django/mutations/resolvers.py
sdobbelaere: I'll have a look to clean that up once we've managed to find the source of the last remaining issue.
Why does the model content-type get reverted to its parent when the object is not present in the qs.

strawberry_django/mutations/resolvers.py
bellini666: Hrm, you are correct!

I remember having to do that in another project, but maybe it is not required anymore :)

Considering this as resolved

strawberry_django/mutations/resolvers.py
sdobbelaere: I've done a few things:
- cleaned up the repeated code
- restructured the update a little to ensure a single flow of data rather then trying to accommodate single and multiple instances
- moved dataclasses into a separate file

Ideally I wanted the new helper in another file, but there was just too much circular imports that made it quite pointless.

sdobbelaere:
@bellini666 My apologies for the 10.000 commits.  I believe I've finally managed to complete this PR, including the typing.  Please re-review.

strawberry_django/mutations/resolvers.py
```diff
@@ -37,20 +34,20 @@
 )
 from strawberry_django.utils.inspect import get_model_fields
 
+from .dataclasses import (
+    _M,
+    _T,
```
bellini666, line 39: Those should not be imported from there but instead need to be defined directly here

strawberry_django/mutations/resolvers.py
```diff
@@ -37,20 +34,20 @@
...
+    FullCleanOptions,
+    ParsedObject,
+    ParsedObjectList,
+    _InputListTypes,
```
bellini666, line 43: This one makes sense, but I would change it to be `InputListTypes` (without the `_`) in the dataclasses module

strawberry_django/mutations/resolvers.py
```diff
@@ -37,20 +34,20 @@
 )
 from strawberry_django.utils.inspect import get_model_fields
 
+from .dataclasses import (
```
bellini666, line 37: Maybe call this module `types.py`?

dataclasses may be conflicting, and types is IMO better because you are actually defining types there, not only dataclasses (e.g. there is a typeddict in there)

strawberry_django/mutations/resolvers.py
```diff
@@ -185,6 +159,87 @@ def parse_input(info: Info, data: Any):
...
+            if value is None and not value_data:
+                value = None  # noqa: PLW2901
+            elif value is None:
+                value = field.related_model._default_manager.create(  # noqa: PLW2901
+                    **value_data,
+                )
+            else:
+                update(info, value, value_data, full_clean=full_clean)
+
+        if direct_field_value:
```
bellini666, line 235: From what I can tell this if will always be `True`, because it gets set to True at the beggining of the for loop, and everytime it gets set to `False` you call `continue`

strawberry_django/mutations/resolvers.py
```diff
@@ -185,6 +159,87 @@ def parse_input(info: Info, data: Any):
     return data
 
 
+def prepare_create_update(
+    *,
+    info: Info,
+    instance: Model,
+    data: dict[str, Any],
+    full_clean: bool | FullCleanOptions = True,
+) -> tuple[Model, dict, list, list]:
```
bellini666, line 168: Cam we type those dict and lists? :)

bellini666 (changes requested):
Np at all for the amount of commits :)

Sorry for giving you more trouble in this review as well. Ping me on discord if you need any help with anything

strawberry_django/mutations/resolvers.py
sdobbelaere: Fair suggestions, it's a more relevant word indeed.

strawberry_django/mutations/resolvers.py
sdobbelaere: Honestly, the original had continue statements without comments.  Personally didn't see the point - but decided to leave them.  There were no comments why they are called.  But yes, happy to remove them, they don't seem to serve any purpose. Unless I'm missing what continue does.

sdobbelaere: I've reworked it a little. Both versions seem to work fine though.

strawberry_django/mutations/resolvers.py
sdobbelaere: Sure. I'll have a look in the morning.

bellini666:
@sdobbelaere I fixed the merge conflict for you, but tests are breaking now after the latest changes

sdobbelaere:
@bellini666 can you check and merge if you're happy with it?  
I did add some comments on  strawberry_django/mutations/resolvers.py:472, as the code seems to be doing nothing from what I can tell.  Tests are also passing when you remove it.

bellini666 (approved):

Issue #398, opened 2023-10-19
Perform validation check on updating nested fields


tokr-bit:
## Summary
If a nested field is updated via the parent field, a validation on the nested field should be performed, if some validators are provided on the django field.

## Feature Request Type

- [ ] Core functionality
- [x] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description

Assuming we have a Parent model with some nested fields:

```python
class Parent(models.Model);
   foo_field = models.CharField()

class Child(models.Model):
   url = models.URLField(validators=[validators.validate_url])
   parent_field = models.ForeignKey(to=Parent, related_name="children")
```

We create some types and a create mutation:

```python
@strawberry.type(Child, exclude=["parent_field"])
class Child:
    id: auto
    url: auto

@strawberry.input(Child)
class ChildInput(Child):
   pass

@strawberry.type(Parent)
class Parent:
   id: auto
   foo_field: auto
   children: List[Child]

@strawberry.input(Parent)
class ParentInput(Parent):
   pass

@strawberry.type
class Mutation:
    parent: Parent = mutations.create(types.ParentInput)
```

No matter which value is passed to the URL Field of the `Child`, the validator is not executed.

## Solution

Reason is that the `update_m2m` method in **resolvers.py** does not execute a `full_clean` on the respective object.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry-graphql-django/issues/398">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/398/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/398/pledge.svg">
</picture>
</a>

bellini666:
Hey @tokr-bit ,

Thanks for reporting this!

I'm wondering if https://github.com/strawberry-graphql/strawberry-graphql-django/pull/394 will also fix this issue?

cc @sdobbelaere

tokr-bit:
Hey,
#394 won't fix the issue. If you and @sdobbelaere support the feature request, I can supply a PR.

[...]

PR #405, opened 2023-10-31
feat: Enforce validation for updating nested relations

tokr-bit:
## Description

If nested one to many / many to many relations are updated, a `full_clean` is performed on the object to be updated. This ensures that Validators are actually executed.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

#398 

## Checklist

### DIFF
--- a/strawberry_django/mutations/resolvers.py
+++ b/strawberry_django/mutations/resolvers.py
@@ -518,8 +518,8 @@ def update_m2m(
         need_remove_cache = need_remove_cache or bool(value.add)
         for v in value.add or []:
             obj, data = _parse_data(info, manager.model, v)
-            obj.full_clean()
             if obj and data:
+                obj.full_clean()
                 manager.add(obj, **data)
             elif obj:
                 # Do this later in a bulk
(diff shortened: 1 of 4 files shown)


bellini666 (approved):
ty! :)

strawberry_django/mutations/resolvers.py
```diff
@@ -474,6 +474,7 @@ def update_m2m(
                 if data:
                     for k, inner_value in data.items():
                         setattr(obj, k, inner_value)
+                    obj.full_clean()
```
bellini666, line 477: I just noticed something. This needs to respect the `full_clean` options that are passed to the update/create methods

bellini666 (changes requested):

strawberry_django/mutations/resolvers.py
tokr-bit: Done in this commit: https://github.com/strawberry-graphql/strawberry-graphql-django/pull/405/commits/3f37550105b9bdc9ca157569871c346536e06a77 Thanks for mentioning.

bellini666 (approved):

PR #409, opened 2023-11-05
fix: keep ordering sequence

### LINKED ISSUE #150: Ordering order being ignored
Hi,
when ordering with 2 and more values order of values is bound to the definition of Order object thus negating user defined order.

Am I mistaken in assuming these two should not be equivalent?
```
query {
  milestoneList(
    order: {
      project: {
        name: ASC
      }
      name: ASC
    }
  ) {
    name
    project {
      name
    }
  }
}
```
```
query {
  milestoneList(
    order: {
      name: ASC
      project: {
        name: ASC
      }
    }
  ) {
    name
    project {
      name
    }
  }
}
```

The result of both of these queries is the same, seems like the actual order is determined via schema definition and not input.

bellini666:
Make sure that the sequence the ordering was given by the client is respected.

Fix #150

### DIFF
--- a/strawberry_django/ordering.py
+++ b/strawberry_django/ordering.py
@@ -1,3 +1,4 @@
+import dataclasses
 import enum
 from typing import (
     TYPE_CHECKING,
@@ -6,7 +7,6 @@ from typing import (
     List,
     Optional,
     Sequence,
-    Tuple,
     Type,
     TypeVar,
     Union,
@@ -21,7 +21,7 @@ from strawberry.field import StrawberryField, field
 from strawberry.type import WithStrawberryObjectDefinition, has_object_definition
 from strawberry.types import Info
 from strawberry.unset import UnsetType
-from typing_extensions import Self, TypeAlias, dataclass_transform
+from typing_extensions import Self, dataclass_transform
 
 from strawberry_django.fields.base import StrawberryDjangoFieldBase
 from strawberry_django.utils.typing import is_auto
@@ -34,9 +34,14 @@ if TYPE_CHECKING:
 
 _T = TypeVar("_T")
 _QS = TypeVar("_QS", bound="QuerySet")
-_OrderSequence: TypeAlias = Dict[str, Tuple[int, "_OrderSequence"]]
 
 ORDER_ARG = "order"
+
+
+@dataclasses.dataclass
+class _OrderSequence:
+    seq: int = 0
+    children: Optional[Dict[str, "_OrderSequence"]] = None
 
 
 @strawberry.enum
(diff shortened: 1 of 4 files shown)


strawberry_django/ordering.py
```diff
@@ -39,9 +43,18 @@ class Ordering(enum.Enum):
...
+    *,
+    sequence: Optional[_OrderSequence] = None,
+    prefix: str = "",
+):
+    sequence = sequence or {}
     args = []
-    for f in order.__strawberry_definition__.fields:
+    for f in sorted(
+        order.__strawberry_definition__.fields,
+        key=lambda f: (seq := sequence.get(f.name, (0,))) and seq[0],
```
patrick91, line 56: I'd make this into a function, I'm not 100% sure what it does 😊

strawberry_django/ordering.py
```diff
@@ -51,17 +64,42 @@ def generate_order_args(order: WithStrawberryObjectDefinition, prefix: str = "")
...
+    queryset: _QS,
+    info: Optional[Info] = None,
+) -> _QS:
     if order in (None, strawberry.UNSET):  # noqa: PLR6201
         return queryset
 
-    args = generate_order_args(order)
+    sequence: _OrderSequence = {}
+    if info is not None and info._raw_info.field_nodes:
+        field_node = info._raw_info.field_nodes[0]
```
patrick91, line 87: we should abstract this in strawberry, let's remember to chat about this in future 😊

strawberry_django/ordering.py
```diff
@@ -51,17 +64,42 @@ def generate_order_args(order: WithStrawberryObjectDefinition, prefix: str = "")
...
+        for arg in field_node.arguments:
+            if arg.name.value != "order" or not isinstance(arg.value, ObjectValueNode):
```
patrick91, line 89: will the argument always be called order?

bellini666: Yes! =P

But you gave me an idea to move this to a constant, together with the one in filters

strawberry_django/ordering.py
bellini666: I refactored it, let me know if it is better :)

patrick91 (approved):

strawberry_django/ordering.py
```diff
@@ -51,17 +74,44 @@ def generate_order_args(order: WithStrawberryObjectDefinition, prefix: str = "")
...
+    sequence: Dict[str, _OrderSequence] = {}
+    if info is not None and info._raw_info.field_nodes:
+        field_node = info._raw_info.field_nodes[0]
+        for arg in field_node.arguments:
+            if arg.name.value != ORDER_ARG or not isinstance(
+                arg.value, ObjectValueNode
+            ):
+                continue
+
+            def parse_and_fill(field: ObjectValueNode, seq: Dict[str, _OrderSequence]):
```
cpontvieux-systra, line 104: @bellini666 
This function definition does not depend on any value of the for-loop. Maybe define it just before the for-loop?

bellini666: In this case the for is just to find the order argument, so this code should happen only once

cpontvieux-systra: Yes I understand, but the `def parse_and_fill` is defined for every iteration of the `for`. It should probably defined just before the for-loop or you prefer to be defined in the for-loop for scope reason? So that `parse_and_fill` only exist within the for-loop?

bellini666: TBH I didn't think too much about that, the only reason I created the function was because I had a recursive code in there =P

Because I created it just to be used exactly in that next line, I decided to create it as close as possible.

Also, as I said, although the function in theory is defined for every iteration, in reality it isn't because all other iterations will `continue` in the previous `if`, only one will actually hit this code

PR #412, opened 2023-11-07
fix: Use _RESOLVER_TYPE as the type for the resolver on field, so the…

### LINKED ISSUE #411: Field with resolver (strawberry_django.field(resolver=resolver)) returns incorrect type
## Describe the Bug

When using strawberry_django.field with resolve in this style,
```python
def resolver(root: "FruitType") -> str:
    return f"{root.id} - {root.name}"

@strawberry_django.type(model=Fruit)
class FruitType:
    id: int
    name: str
    id_name: str = strawberry_django.field(resolver=resolver)
```
the type returned is going to be `StrawberryDjangoField` instead of `str`.
This happens only for resolvers with arguments. When the resolver uses no arguments, the correct type is returned.

This is similar to [this](https://github.com/strawberry-graphql/strawberry/issues/2169) issue reported at `strawberry-graphql/strawberry`

As proposed by the above issue, a fix could be replacing this (https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/strawberry_django/fields/field.py#L387):
```python
@overload
def field(
    *,
    ...
    resolver: Callable[[], _T],
    ...
) -> _T: ...
```
for this
```python
@overload
def field(
    *,
    ...
    resolver: Callable[..., _T],
    ...
) -> _T: ...
```
Another options is to take the same approach used by `strawberry` [here](https://github.com/strawberry-graphql/strawberry/pull/2172/commits/c14a4964098fbcbd252fbac029f87dcc398c701c#diff-00af9d43c9b59404ba170e72e470939f2a6dd50e2eae24d0e24ef105431e5414R42-R47)

```python
_RESOLVER_TYPE = Union[
    StrawberryResolver[T],
    Callable[..., T],
    Callable[..., Coroutine[T, Any, Any]],
    Callable[..., Awaitable[T]],
    "staticmethod[Any, T]",
    "classmethod[Any, Any, T]",
]

@overload
def field(
    *,
    ...
    resolver: _RESOLVER_TYPE[T],
    ...
) -> _T: ...
```
## System Information

 - Operating system: macOS Sonoma 14.1
 - Strawberry version (if applicable): not applicable

## Additional Context
I can create a PR with the preferred solution if this is indeed an issue.

guizesilva:
… correct type is returned when using a resolver with arguments.

## Description

Relate to this [issue](https://github.com/strawberry-graphql/strawberry-graphql-django/issues/411), use the type `_RESOLVER_TYPE` for field's resolver type, so it returns the correct type when using a resolver with arguments, and stop pyright from complaining about incorrect types.

## Types of Changes

- [] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #411

## Checklist

### DIFF
--- a/strawberry_django/fields/field.py
+++ b/strawberry_django/fields/field.py
@@ -384,7 +384,7 @@ class StrawberryDjangoConnectionExtension(relay.ConnectionExtension):
 def field(
     *,
     field_cls: type[StrawberryDjangoField] = StrawberryDjangoField,
-    resolver: Callable[[], _T],
+    resolver: _RESOLVER_TYPE[_T],
     name: str | None = None,
     field_name: str | None = None,
     is_subscription: bool = False,
@@ -439,7 +439,7 @@ def field(
 
 @overload
 def field(
-    resolver: StrawberryResolver | Callable | staticmethod | classmethod,
+    resolver: _RESOLVER_TYPE,
     *,
     field_cls: type[StrawberryDjangoField] = StrawberryDjangoField,
     name: str | None = None,


bellini666 (approved):
Thank you :)

PR #417, opened 2023-11-15
perf: cache nested import of get_user_or_annonymous to improve performance

bellini666:


strawberry_django/permissions.py
```diff
@@ -48,18 +48,28 @@
 from .utils.query import filter_for_user
 from .utils.typing import UserType
 
-try:
-    # Try to use the smaller/faster cache decorator if available
-    _cache = functools.cache  # type: ignore
-except AttributeError:
-    _cache = functools.lru_cache
-
```
bellini666, line 56: this was not being used lol

strawberry_django/permissions.py
```diff
@@ -48,18 +48,28 @@
...
 
 
 _M = TypeVar("_M", bound=Model)
 
+_sentinel: Any = object()
+_user_or_anonymous_getter: Optional[Callable[[UserType], UserType]] = _sentinel
+
+
+def _get_user_or_anonymous_getter() -> Optional[Callable[[UserType], UserType]]:
+    global _user_or_anonymous_getter  # noqa: PLW0603
```
patrick91, line 62: wouldn't this be nicer with the lru cache?

bellini666: Wasn't sure if that was an overkill, but you are probably right! Going to change it

bellini666: Done!

patrick91 (approved):

PR #418, opened 2023-11-17
Fix typing issues

patrick91:
This should fix an issue I was having with latest pyright 😊

Not sure if Prefetch in django is actually generic (I'm using django types), that's why I put it behind quotes 😊


bellini666 (approved):
Nice! :)

PR #421, opened 2023-11-19
Login and CurrentUser queries yield broken responses

sdobbelaere:
When doing a login with an incorrect username/password or querying me/current_user both cases result in a cryptical reponse:
```json{
  "data": null,
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Query.me.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "me"
      ]
    }
  ]
}
```

This PR corrects this behaviour by raising an error and with an understandable error.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation


strawberry_django/auth/mutations.py
```diff
@@ -28,29 +30,30 @@
...
 
 @django_resolver
-def resolve_login(info: Info, username: str, password: str) -> AbstractBaseUser | None:
+def resolve_login(info: Info, username: str, password: str) -> AbstractBaseUser:
     request = get_request(info)
     user = auth.authenticate(request, username=username, password=password)
 
-    if user is not None:
+    if user is None:
+        raise IncorrectUsernamePasswordError()  # noqa: RSE102
```
bellini666, line 38: I don't think we should be creating our own exceptions here, but instead using django's `ValidationError`

If we want custom exceptions (is there a good reason for one?) then we should inherit those from `ValidationError` directly too

strawberry_django/auth/queries.py
```diff
@@ -1,15 +1,23 @@
...
+    from strawberry_django.utils.typing import UserType
+
 
-def resolve_current_user(info: Info):
+def resolve_current_user(info: Info) -> UserType:
     user = get_current_user(info)
 
     if not getattr(user, "is_authenticated", False):
-        return None
+        raise UserNotLoggedInError()  # noqa: RSE102
```
bellini666, line 20: Ditto

strawberry_django/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class DefaultMessageError(Exception):
```
bellini666, line 1: As mentioned, don't think we need this

bellini666 (changes requested):
This looks good in general :)
Some minor changes and we should be good to merge

strawberry_django/auth/mutations.py
sdobbelaere: That's fine, I don't mind either way.  I find custom exceptions extremely useful for catching errors cleanly - but I do agree they don't expose to the frontend - so not as useful as they could be.

I'll revert it to ValidationError

bellini666 (approved):
Thank you! :)

Issue #422, opened 2023-11-21
"OR" filter not working for OneToOne-related models


TWeidi:
## Describe the bug

When running a query including an "OR" filter the filter gets passed as "AND" (according to the query shown in the Django Debug Toolbar)

### Django Model
```python
class SomeBaseModel(models.Model):
     some_base_attr = models.CharField(...)
    ...

class SomeModel(SomeBaseModel):
    some_other_attr = models.FloatField(...)
    ...
```

### Types
```python
@type(models.SomeBaseModel, filters=filters.SomeBaseFilter)
class SomeBaseNode(relay.Node):
    some_base_attr: auto
    ...

@type(models.SomeModel, filters=filters.SomeFilter)
class SomeNode(SomeBaseNode):
    some_other_attr: auto
    ...
```

### Filters
```python
@filter(models.SomeBaseModel, lookups=True)
class SomeBaseFilter:
    some_base_attr: auto
    id: auto
    ...

@filter(models.SomeModel, lookups=True)
class SomeFilter(SomeBaseFilter):
    some_other_attr: auto
    id: auto
    ...
```

### Query
```
{
  someQuery(filters: {
    id: {exact: "U29tZU5vZGU6MjMwNg=="}
    OR: {id: {exact: "U29tZU5vZGU6MjI3NA=="}}
  }) {
    edges {
      node {
        id
      }
    }
  }
}
```

I guess it has to do with the `build_filter_kwargs` function in filters.py. Trying to figure out how the function works, but, unfortunately, it's a bit complicated ;)

## System Information

 - Operating system: Ubuntu 20.04.6 LTS
 - Strawberry version (if applicable):
   - strawberry-graphql: version 0.215.1
   - strawberry-graphql-django: 0.25.0

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry-graphql-django/issues/422">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/422/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/422/pledge.svg">
[...]

PR #424, opened 2023-11-22
Guarantee 'AND', 'OR', and 'NOT' filter fields get evaluated last by …

TWeidi:
…sorting 'filters.__strawberry_definition__.fields'; fixed issue #422;

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

#422

## Checklist

### DIFF
--- a/tests/filters/test_filters.py
+++ b/tests/filters/test_filters.py
@@ -285,8 +285,26 @@ def test_resolver_filter_with_inheritance(vegetables):
 
     query = utils.generate_query(Query)
     result = query(
-        "{ vegetables(filters:  { worldProduction: { gt: 100e6 } "
-        'OR: { name: { exact: "cucumber" } } }) { id name } }'
+        """
+      {
+        vegetables(
+          filters: {
+            worldProduction: {
+              gt: 100e6
+            }
+            OR: {
+              name: {
+                exact: "cucumber"
+              }
+            }
+          }
+        )
+        {
+          id
+          name
+        }
+      }
+    """
     )
     assert isinstance(result, ExecutionResult)
     assert not result.errors
--- a/tests/filters/test_filters.py
+++ b/tests/filters/test_filters.py
@@ -12,6 +12,18 @@ import strawberry_django
 from tests import models, utils
 
 
+@strawberry_django.filter(models.NameDescriptionMixin)
+class NameDescriptionFilter:
+    name: auto
+    description: auto
+
+
+@strawberry_django.filter(models.Vegetable, lookups=True)
+class VegetableFilter(NameDescriptionFilter):
+    id: auto
+    world_production: auto
+
+
 @strawberry_django.filters.filter(models.Color, lookups=True)
 class ColorFilter:
     id: auto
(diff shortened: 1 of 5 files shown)


bellini666:
It would be nice to have a test that would break without this change, not only to evaluate that this change is in fact fixing the issue, but also to avoid regressions in the future

TWeidi:
Ok, I'll try to add the two test cases :)

TWeidi:
@bellini666:

I'm having a hard time using the debugger in VSCode to debug into the tests. My breakpoints are are outright ignored when i try to run a test in debug mode. Do i have to provide specific settings to correctly enable debugging?

bellini666:
@TWeidi not sure TBH as I don't use VSCode and I'm also a print debugger dev =P

@patrick91 do you have any knowledge to share regarding that?

TWeidi:
Hey guys,

### Regarding test debugging in VSCode
I found [this](https://github.com/microsoft/vscode-python/issues/693) article, which states that debugging tests in vscode does not work with pytest's --cov option, so commenting out the line starting with 'addopts' in pyproject.toml led to test debugging working again (just in case someone else encounters the same issue).

### Regarding the requested test
The test `test_resolver_filter_with_inheritance` in _tests/filters/test_filters.py_ fulfills your requirement to fail with the current version but succeeds with the proposed fix.

I added another test (test_filter_field_order_with_inheritance) in tests/filters/test_types.py just to show, that the order of fields of a filter class depends on class inheritance. This test will fail for both the original and the fixed version and is not supposed to be included in case this PR is merged. 
Maybe it would be a better solution to fix the issue by defining a base Filter class with a dedicated  __init_subclass__() method that takes care about the order of filter fields and which must be subclassed by every filter, just in case the fields order is relevant in other parts of the code as well?

tests/filters/test_filters.py
```diff
@@ -253,6 +273,30 @@ def fruits(self, filters: FruitFilter) -> List[Fruit]:
...
+        def vegetables(self, filters: VegetableFilter) -> List[Vegetable]:
+            queryset = models.Vegetable.objects.all()
+            return cast(
+                List[Vegetable], strawberry_django.filters.apply(filters, queryset)
+            )
+
+    query = utils.generate_query(Query)
+    result = query(
+        "{ vegetables(filters:  { worldProduction: { gt: 100e6 } "
+        'OR: { name: { exact: "cucumber" } } }) { id name } }'
```
bellini666, line 289: nit: can we have this formatted properly using a multiline string? check `tests/test_optimizer.py` for some examples

bellini666 (approved):
Thanks for you this :)

Left a small nit, other than that this seems fine to merge

TWeidi:
Do you want me to remove the failing test (it was just to show you the issue arising from the fields order) or will you do that after investigating it further?

bellini666:
> Do you want me to remove the failing test (it was just to show you the issue arising from the fields order) or will you do that after investigating it further?

If you can modify it in a way that it would break without this changes but passes with them is better :)

TWeidi:
Ok, I looked into this and I would rather drop the failing test. The mixed up order is not a wrong behavior, but the one expected from the typing module:

> The order of the fields in all of the generated methods is the order in which they appear in the class definition. ([1](https://docs.python.org/3.7/library/dataclasses.html#dataclasses.dataclass))

So, the better approach to the issue would be to not rely at all on the order of fields in the `build_filter_kwargs` function in filters.py.
Unfortunately, I don't know your library well enough to dare to rewrite that function myself. As a quickfix I guess this PR will do, though.

bellini666:
@TWeidi nice, this LGTM

Do you want to remove it from draft? Then I can merge it

PR #430, opened 2023-12-04
Fix: ordering when dealing with camelCased field

he0119:
## Description

I found that `f.name.value` at L110 is camelCase value while at L63 `f.name` is underscore value which makes `sort_key` function always return 0.

https://github.com/strawberry-graphql/strawberry-graphql-django/blob/c90a7907913188cb5b70ee74e3fbe4a1ead7c8a2/strawberry_django/ordering.py#L110
https://github.com/strawberry-graphql/strawberry-graphql-django/blob/c90a7907913188cb5b70ee74e3fbe4a1ead7c8a2/strawberry_django/ordering.py#L62-L65

Try to fix it by converting `f.name` to camelCase value, don't know if this is the right way to fix the bug (will `f.name.value` always be camelCase value?).

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

<https://github.com/strawberry-graphql/strawberry-graphql-django/issues/150#issuecomment-1835527261>

## Checklist

### DIFF
--- a/strawberry_django/ordering.py
+++ b/strawberry_django/ordering.py
@@ -21,6 +21,7 @@ from strawberry.field import StrawberryField, field
 from strawberry.type import WithStrawberryObjectDefinition, has_object_definition
 from strawberry.types import Info
 from strawberry.unset import UnsetType
+from strawberry.utils.str_converters import to_camel_case
 from typing_extensions import Self, dataclass_transform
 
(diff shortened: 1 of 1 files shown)


bellini666 (approved):
Ty! :)

Issue #432, opened 2023-12-05
manually using strawberry_django.filters.apply does not work anymore after v0.27


devkral:
my queries which use OR and AND doesn't work anymore. I think the sorting breaks stuff, you may need to convert the key lambda to

`lambda f: (f.name in {"OR", "AND", "NOT"}, f.name`)

but I didn't investigated the bug in detail.

Note:  I use strawberry_django.filters.apply manually

## System Information

 - Operating system:
 - Strawberry version (if applicable):

devkral:
maybe f.name isn't required at all as well as reverse (note: in this case the suggested code must be changed to f.name in ...)

bellini666:
@devkral ohh thank you for raising that issue.

The changes was done in [this PR](https://github.com/strawberry-graphql/strawberry-graphql-django/pull/424) to fix https://github.com/strawberry-graphql/strawberry-graphql-django/issues/422

Apparently you have made some changes that should fix your issue while not regressing on that fix. Would you like to open a PR with that? I'll try to review it ASAP!

PR #435, opened 2023-12-06
fix: fix #424, really push OR, AND and NOT to the end

devkral:
Fix for #432

really and only push AND, OR and NOT to the end.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fix https://github.com/strawberry-graphql/strawberry-graphql-django/issues/432

## Checklist


devkral:
thanks, note: sorting should be only used in emergencies as it has a bad performance.

There should be maybe an evaluation where sorting is used and not cached or where user input is sorted (risk of denial of service, therefor just pushing NOT, AND and OR back). But this is another topic

PR #437, opened 2023-12-08
fix: HasPerm on async fields, fix missing query in another test

devkral:
The HasPerm extension  called in an async context get_current_user. This caused decorated async fields to crash

The test schema was updated for an example.

This PR fixes also a minor bug in another test: the query execution was missing

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


devkral:
note: I am not quite familar with pytest. There is most probably a better way to exclude the test from non async contexts

bellini666 (approved):

bellini666:
Ty! :)

PR #441, opened 2023-12-15
Fix code example on updating `field_type_map`.

alimony:
## Description

The `field_type_map` example docs is broken because the import path is wrong. This fixes the import, and also adds a models import so this code actually works.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation


bellini666 (approved):
Thank you very much for this! :)

PR #159, opened 2022-08-02
Feature: Ability to set single object lookup key

benhowes:
## Description

This PR adds the ability to alter the behaviour of a single object lookup. Currently the behaviour is that a single object fetch can only operate on the `pk` of that object (with `pk` as the variable name), but in many cases developers may want to use other names/types.

For maximum flexibility, this PR adds 3 options to the `strawberry.django.field` definition which can be used in many combinations, such as:

```python

@strawberry.type
class Query:
   fruit: Fruit = strawberry.django.field(
      lookup_key="id",
      lookup_key_django_name="uuid",
      lookup_key_type=UUID
   )

```

or

```python

@strawberry.type
class Query:
   fruit: Fruit = strawberry.django.field(
      lookup_key="public_id",
      lookup_key_type=str
   )

```

or

```python

@strawberry.type
class Query:
   fruit: Fruit = strawberry.django.field(
      lookup_key="id",
   )

```

It defaults to the existing behaviour of the lookup field being `pk` and the type being `strawberry.ID`.

The only "breaking" change is that the lookup key is now a required field, so this will slightly alter the generated schema. This can be considered more of a bugfix since the lookups would not work without an argument, other than in the special case where there was only a single record of that type (in which case `.get()` on the queryset _would_ work). I would hope that's not behaviour that anyone is relying on though...  

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

- Closes #107 
- Closes #121 
- Closes #130

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation. - **This PR should build on #147 when that's merged**
- [x] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #446, opened 2023-12-25
Feature: Add setting DEFAULT_PK_FIELD_NAME

### LINKED ISSUE #121: Possibility to rename pk to id or something else
Currently, when a filter is created for a single object, "pk" is used as the ID name. There is no possibility to specify an alternative name.

## Feature Request Type

- Change (enhancement/optimization) of an existing feature.

## Description
Ability to set custom name for "pk".

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry-graphql-django/issues/121">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/121/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

noamsto:
This PR was inspired by #159.
I took a different approach which is less granular but easier to implement.
If a fine-grain control is needed it can be implemented in the future.

## Description

Add a `DEFAULT_PK_FIELD_NAME` setting which default to `"pk"`.
This should affect all builtin Django CRUDs, and is backward compatible. 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* Closes #121 

## Checklist


noamsto:
@bellini666 
What do you think of this?
I know #159 is abandoned and this is a partial solution, but I know I will use it at least :).

I can add more tests if you think are needed.

Also, need your help regarding this: https://github.com/strawberry-graphql/strawberry-graphql-django/pull/446/files#diff-87c07c1c9f6db39e2dacd257d14f3123d74b65d0a1db4029364d69f4a707013cR55

strawberry_django/filters.py
```diff
@@ -36,10 +36,13 @@
 
 from .arguments import argument
 from .fields.base import StrawberryDjangoFieldBase
+from .settings import strawberry_django_settings as django_settings
 
 if TYPE_CHECKING:
     from django.db.models import QuerySet
 
+
+settings = django_settings()
```
bellini666, line 45: I prefer to have this instantiated where it is used to make sure we get any changes made to settings.

strawberry_django/mutations/fields.py
```diff
@@ -257,10 +262,13 @@ def create(self, data: dict[str, Any], *, info: Info):
             )
 
 
+settings = strawberry_django_settings()
```
bellini666, line 265: Ditto

strawberry_django/mutations/mutations.py
```diff
@@ -28,6 +30,7 @@
 )
 
 _T = TypeVar("_T")
+settings = strawberry_django_settings()
```
bellini666, line 33: Ditto

strawberry_django/filters.py
```diff
@@ -49,7 +52,7 @@
 
 @strawberry.input
 class DjangoModelFilterInput:
-    pk: strawberry.ID
+    pk: strawberry.ID  # TODO: How to override pk here with settings["DEFAULT_PK_FIELD_NAME"] ?
```
bellini666, line 55: In theory we could do:

```python

class DjangoModelFilterInput:
    pass

DjangoModelFilterInput.__annotations__[settings["DEFAULT_PK_FIELD_NAME"]] = strawberry.ID
DjangoModelFilterInput = strawberry.input(DjangoModelFilterInput)
```

But that would mean forcing django to be imported earlier (which can lead to some issues). I'll try to think if there's a more elegant solution.

What do you think @patrick91 ?

bellini666 (commented):
@noamsto I think this is mostly fine, and I commented on the issue you requested my attention, although I'm not sure it is the correct option in there

strawberry_django/filters.py
noamsto: I can wrap it in a function to prevent the import from Module scope

strawberry_django/mutations/mutations.py
noamsto: I was thinking,
Maybe it's redundant here?
Anyway `key_attr` is propagating to [`DjangoMutationCUD`](https://github.com/strawberry-graphql/strawberry-graphql-django/pull/446/files#diff-c5989633bdbcc1fe4c92d5b3bb2b37ce9bbe1dc3d793adea4552d94caadd89adR178) and there it will be handled for Update and Delete right?

strawberry_django/filters.py
noamsto: ack

bellini666 (commented):
Looking good! :)

I think the only thing missing is is the `DjangomodelFilterInput`

I was thinking about it and probably what we want to do it to create it dynamically if `DEFAULT_PK_FIELD_NAME != "pk"` like I suggested at the point where it is being used (it is only used in one place) and save it in a global variable for further usage.

What do you think?

noamsto:
> Looking good! :)
> 
> I think the only thing missing is is the `DjangomodelFilterInput`
> 
> I was thinking about it and probably what we want to do it to create it dynamically if `DEFAULT_PK_FIELD_NAME != "pk"` like I suggested at the point where it is being used (it is only used in one place) and save it in a global variable for further usage.
> 
> What do you think?

I don't mind doing that, though the usage is in a different file; defined in: `strawberry_django/filters.py`, used in: `strawberry_django/fields/types.py` it's a bit confusing no?

It is also used in `test_types.py` wouldn't it break the test?

bellini666 (approved):

PR #449, opened 2024-01-04
Allow updates of nested fields

tokr-bit:
# Summary

This PR allows updates of nested fields, i.e.

- update the field by setting the ID
- add copies of Child by leaving the ID unset
- remove a Child by omitting it

## Description

The `update_m2m`, `parse_data` and the `_parse_pk` methods were changed.  `_parse_pk` does now return the id as well. If `parse_data` finds an UNSET id, the id is still returned. The id is used to check if a new object is to be created or not. If we have an Input type, the ID is not part of the `data`-dict. This means, we perform a `get_or_create`.

Since `key_attr` can be used to change the primary key from `id` to an "arbitrary" value, major API changes had to be made. This was proposed in https://github.com/strawberry-graphql/strawberry-graphql-django/issues/383.

I am not sure, if this leads to breaking changes - a couple of methods were affected.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

https://github.com/strawberry-graphql/strawberry-graphql-django/issues/383

## Checklist

### DIFF
--- a/strawberry_django/mutations/resolvers.py
+++ b/strawberry_django/mutations/resolvers.py
@@ -83,7 +83,7 @@ def _parse_pk(
 def _parse_data(
     info: Info, model: type[_M], value: Any, *, key_attr: str | None = "pk"
 ):
-    obj, data = _parse_pk(value, model, key_attr)
+    obj, data = _parse_pk(value, model, key_attr=key_attr)
     parsed_data = {}
     if data:
         for k, v in data.items():
@@ -127,21 +127,23 @@ def parse_input(info: Info, data: Any, *, key_attr: str | None = "pk") -> Any: .
 
 def parse_input(info: Info, data: Any, *, key_attr: str | None = "pk"):
     if isinstance(data, dict):
-        return {k: parse_input(info, v, key_attr) for k, v in data.items()}
+        return {k: parse_input(info, v, key_attr=key_attr) for k, v in data.items()}
 
     if isinstance(data, list):
-        return [parse_input(info, v, key_attr) for v in data]
+        return [parse_input(info, v, key_attr=key_attr) for v in data]
 
     if isinstance(data, relay.GlobalID):
         return data.resolve_node_sync(info, required=True)
 
     if isinstance(data, NodeInput):
-        pk = cast(Any, parse_input(info, getattr(data, "id", UNSET), key_attr))
+        pk = cast(Any, parse_input(info, getattr(data, "id", UNSET), key_attr=key_attr))
         parsed = {}
         for field in dataclasses.fields(data):
             if field.name == "id":
                 continue
-            parsed[field.name] = parse_input(info, getattr(data, field.name), key_attr)
+            parsed[field.name] = parse_input(
+                info, getattr(data, field.name), key_attr=key_attr
+            )
 
         return ParsedObject(
             pk=pk,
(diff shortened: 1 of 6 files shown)


strawberry_django/mutations/resolvers.py
```diff
@@ -54,8 +54,7 @@
 
 
 def _parse_pk(
-    value: ParsedObject | strawberry.ID | _M | None,
-    model: type[_M],
+    value: ParsedObject | strawberry.ID | _M | None, model: type[_M], key_attr: str
```
bellini666, line 57: Give it a default value to avoid it being required

```suggestion
    value: ParsedObject | strawberry.ID | _M | None,
    model: type[_M],
    *,
    key_attr: str = "id",
```

strawberry_django/mutations/resolvers.py
```diff
@@ -67,67 +66,71 @@ def _parse_pk(
...
+                return model._default_manager.get(pk=obj_pk), value
         return None, value
 
     return model._default_manager.get(pk=value), None
 
 
-def _parse_data(info: Info, model: type[_M], value: Any):
-    obj, data = _parse_pk(value, model)
-
+def _parse_data(info: Info, model: type[_M], value: Any, key_attr: str):
```
bellini666, line 80: Ditto, and the same applies to every other places in this PR

```suggestion
def _parse_data(info: Info, model: type[_M], value: Any, *, key_attr: str = "id"):
```

bellini666 (changes requested):
Looking good! :)

Asked for some changes and also take a look at the typing issues

strawberry_django/mutations/resolvers.py
tokr-bit: Thank you! I am not sure about the default value, shouldn't it be `pk`, as this is the default value set in the `DjangoCUDMutation` (https://github.com/strawberry-graphql/strawberry-graphql-django/blob/main/strawberry_django/mutations/fields.py#L178)

tokr-bit: A friendly ping @bellini666  :)

bellini666: @tokr-bit oh, sorry, somehow I lost the previous notification :)

Yeah I think you are correct! The default is usually `pk` because it will work for any model's primary key even if you set it to something other than `id`

The `id` itself needs to be checked for relay stuff

tokr-bit: @tokr-bit Thanks for getting back to me! I followed your requests. In my opinion, the branch is ready to be merged.

strawberry_django/mutations/resolvers.py
```diff
@@ -67,67 +69,81 @@ def _parse_pk(
         return value.parse(model)
 
     if isinstance(value, dict):
+        if key_attr in value:
+            obj_pk = value[key_attr]
+            if obj_pk != strawberry.UNSET:
+                if key_attr == "id":
+                    obj_pk = int(obj_pk)
```
bellini666, line 76: I think this can lead to issues. e.g. if your `id` is a `UUID` or something else

Note that `SomeModel.objects.get(pk='1234')` will work fine as django itself will cast the string to int if the column is an int

```suggestion
```

bellini666 (changes requested):
Hey @tokr-bit ,

Thanks for this PR :). LGTM now! Only left one small detail to change, let me know if you have any concerns on that.

After that has been solved this can be merged :)

strawberry_django/mutations/resolvers.py
bellini666: Not related to your PR :)

Thanks for this!

bellini666 (approved):

PR #456, opened 2024-01-18
Documentation improvements

thclark:
## Description

This PR contains (what I think are) documentation improvements.

Mainly, the thrust is to orient the quickstart toward users familiar with django, but less familiar with graphql and strawberry.

*My assumption is that a large majority of users with this library are more experienced with django than with strawberry* - I think the typical pattern will be discovering strawberry while trying to add a graph API to their django app, rather than vice versa. That said, I think this updated version is accessible to folks who are also inexperienced with django.

## Changes

- Clarified library purpose and proposition on the homepage.
- Minor tweaks to homepage design (eg not displaying the 'home' H1 line and bolding the strapline)
- Links to the example app front and centre (I didn't know it existed until after many hours of struggling)
- Deduplicate and enhance quick-start guide
  - A very similar guide was duplicated twice, once in `index.md` and once in `quick-start.md`.
  - I started with the `index.md` one because the other seemed more oriented to teaching django
    - In particular the original `quick-start` stepped slowly through adding a ForeignKey - a django basic. I felt that was less valuable than focusing on the strawberry-django specifics.
  - I split the steps to delineate descriptions about what Types and Queries are for, how they relate to models, and what the advantages of `strawberry-django` are in particular.
    - Including a warning about `fruits: list[Fruit]` syntax is for where it's used in two different cases, which I struggled with before the penny dropped.
- Removed the deprecated project cross reference to strawberry-django-plus, to reduce confusion for new users hitting an archived repo link (I'd been looking back at the discord, for example, and tried to install that to do something before realising it was already here - it's time to move on and clear up).
- Replace deprecated emoji extension creating warnings in the doc builds

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [x] Documentation

## Checklist


sdobbelaere:
I was one of those familiar with Django and not with Strawberry.  One of the things that threw me off at first was the use of the model names as type names.  

What do you think of changing:

```python
import strawberry_django

from strawberry import auto
from typing import List

@strawberry_django.type(models.Fruit)
class Fruit:
    id: auto
    name: auto
    color: "Color"

@strawberry_django.type(models.Color)
class Color:
    id: auto
    name: auto
    fruits: List[Fruit]
```

into

```python
import strawberry_django
from fruits.models import Fruit, Color
from strawberry import auto
from typing import List

@strawberry_django.type(Fruit)
class FruitType:
    id: auto
    name: auto
    color: "Color"

@strawberry_django.type(Color)
class ColorType:
    id: auto
    name: auto
    fruits: List[FruitType]
```

That would have helped me a lot reading through the guides at the time.

bellini666 (approved):
This looks awesome!! :)

Thank you so much for this

bellini666:
> I was one of those familiar with Django and not with Strawberry. One of the things that threw me off at first was the use of the model names as type names.

@sdobbelaere I totally agree on that.

The use of model names as type names in the documentation predates me, I also prefer the `Type` suffix. In my personal projects I'll usually do:

```python
class Fruit(models.Model):
    ...

@strawberry_django.type(Fruit, name="Fruit")
class FruitType:
    ...
```

That way I don't get them confused but the name is still exposed as `Fruit` to the schema.

As this PR is basically ready and what @thclark basically followed the style we previously had, I'm going to merge it, but I think it is worth opening another one to make those changes in the documentation

thclark:
Yeah I'm in agreement with both of you, that would be a good clarifying change.

PR #460, opened 2024-01-25
fix handling of null for lookup filters

### LINKED ISSUE #459: lookups doesn't accept null
when defining a query filter like:

```
query {
  users(filter: {username: {contains: null}}){
  }
}
```

the query fails with with: "Cannot use None as a query value" except for some filters like exact. And for exact this is a spec violation:

filters shall accept Unset or None to not filter

This bug causes errors when testing with schemathesis.
PR follows

## System Information

 - Strawberry version (if applicable): 0.29

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry-graphql-django/issues/459">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/459/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

devkral:
When a lookup is detected, ignore null values.
This PR is a bit conservative and also ignores UNSET but this is most probably not necessary.

This prevents django errors for null specifications and ignores null like the rest of the API does.

Is an exception for the filter exact wanted (e.g. foo_exact=null)? Shall I make a settings switch?

note: exact=null is the same like is_null=true, so there would be no functionality loss, when exact=null would not filter anymore

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*  #459

## Checklist

### DIFF
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+import warnings
 from enum import Enum
 from types import FunctionType
 from typing import (
@@ -251,12 +252,19 @@ def build_filter_kwargs(
             )
             filter_kwargs &= subfield_filter_kwargs
             filter_methods.extend(subfield_filter_methods)
-        elif not is_lookup or (field_value is not UNSET and field_value is not None):
+        elif not is_lookup or field_value is not None:
+            assert field_value is not UNSET, "UNSET should have been filtered before"
             # cannot use in set because field_value can be a list which is unhashable
             filter_kwarg = Q(**{f"{path}{field_name}": field_value})
             if negated:
                 filter_kwarg = ~filter_kwarg
             filter_kwargs &= filter_kwarg
+        elif is_lookup and field_name.endswith("exact"):
+            warnings.warn(
+                "exact can't be used with null anymore, use isnull",
+                DeprecationWarning,
+                stacklevel=1,
+            )
 
     return filter_kwargs, filter_methods
 
--- a/strawberry_django/filters.py
+++ b/strawberry_django/filters.py
@@ -244,6 +244,15 @@ def build_filter_kwargs(
             ):
                 continue
 
+        if is_lookup and field_name.endswith("exact"):
+            warnings.warn(
+                "exact can't be used with null anymore (will be ignored), use isnull",
+                DeprecationWarning,
+                stacklevel=1,
+            )
+            # a bit hacky but is just for a deprecation
+            is_lookup = False
+
         if has_object_definition(field_value):
             subfield_filter_kwargs, subfield_filter_methods = build_filter_kwargs(
                 cast(WithStrawberryObjectDefinition, field_value),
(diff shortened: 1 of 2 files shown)


strawberry_django/filters.py
```diff
@@ -169,8 +169,7 @@ def _resolve_global_id(value: Any):
 
 
 def build_filter_kwargs(
-    filters: WithStrawberryObjectDefinition,
-    path="",
+    filters: WithStrawberryObjectDefinition, path="", is_lookup=False
```
bellini666, line 172: question: why do we need to pass `is_lookup` here? Isn't it just a matter of doing `is_lookup = isinstance(field_value, FilterLookup)` at the beggining of the function?

strawberry_django/filters.py
```diff
@@ -248,10 +247,12 @@ def build_filter_kwargs(
             subfield_filter_kwargs, subfield_filter_methods = build_filter_kwargs(
                 cast(WithStrawberryObjectDefinition, field_value),
                 f"{path}{field_name}__",
+                is_lookup=isinstance(field_value, FilterLookup),
             )
             filter_kwargs &= subfield_filter_kwargs
             filter_methods.extend(subfield_filter_methods)
-        else:
+        elif not is_lookup or (field_value is not UNSET and field_value is not None):
+            # cannot use in set because field_value can be a list which is unhashable
```
bellini666, line 255: Not sure about this... Specially after looking at the PR, silently ignoring this IMO is even worse, because someone might expect this to work.

I think the error in this case is better. But, we might try to find a way to workaround this? e.g. `name: {exact: null}` can be translated to `name__isnull=True`

devkral: that is my thought.

strawberry_django/filters.py
devkral: it looses that information because of the cast, so this is the only place to know

strawberry_django/filters.py
bellini666: Hrm, it is an option, but I would still raise an error in this case, like "exact can't be used with `null`, use `isnull` instead" or something like this

devkral: sounds good

devkral: I issue now a UserWarning, or should I use an DeprecationWarning and revert for exact to the old logic?

strawberry_django/filters.py
devkral: do you have a better idea?

strawberry_django/filters.py
bellini666: I think this should error out, not just a warning.

Looking from the backend perspective, I understand that the user is going to be able to see the warning. But anyone using it from the frontend can get very confused when a query for null doesn't work as expected, and silently ignores the given parameter without any error

devkral: I see your point, but what is if the user wants null values to not error (like me)? Do we want to have setting?
In this case we can handle both ways

devkral: Or would it be better to simply deprecate this feature? This would be the easiest way.

devkral: I think a proper deprecation is the best way. This way we don't break anything and warn the user .

devkral: you can use: -Wdeprecation for that.

If you still want a runtime error, I will move the error logic into an extra function, so I can overwrite it in my tests

devkral:
fixed in an another PR

PR #465, opened 2024-01-27
feat(optimizer): Do not defer select_related fields if no only was specified

### LINKED ISSUE #455: Select/prefetch related fields with async query.
## Field cannot be both deferred and traversed using select_related at the same time.

While trying to use `select_related` in `@strawberry_django.type` decorator I am getting this error:
```
Field Product.type cannot be both deferred and traversed using select_related at the same time.
```

I am using a chained relations in my models like this:
`Cart--CartItem--Product--ProductType `
and appropriate `strawberry_django.type`s chain, i.e.:
```
@strawberry_django.type(models.Product, select_related="type")
class ProductType:
    name: strawberry.auto
    description: strawberry.auto
    sku: strawberry.auto
    
    @strawberry.field(name="type", description="Ecom Product Type code")
    async def _type(self) -> str:
        return self.type.code
   
@strawberry_django.type(models.CartItem)
class CartItemType:
    product: ProductType
    quantity: strawberry.auto
    price: strawberry.auto
    meta: strawberry.auto


@strawberry_django.type(models.Cart)
class CartType:
    items: typing.List[CartItemType]
    total: strawberry.auto
```

in addition, there is an async version of queries/mutations like this:
```
@strawberry.field
    async def cart(...) -> CartType: ...
```

The problem is resolved just if I prepare the queryset on the top level in query with all possible (which could be requested) selected and prefetched models.
```
@strawberry.field
    async def cart(...) -> CartType: 
    return await Cart.objects.select_related(...).prefetch_related("cart_items__product__product_type").aget(...)
```

## System Information

 - Operating system: MacOS 14.2.1
strawberry-graphql = "0.209.2"
strawberry-graphql-django = "0.19.0"
django="4.2.9"

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry-graphql-django/issues/455">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/455/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

bellini666:
When doing `select_related=["foo"]` and not doing `only=["foo_<something>"]`, we should add
`"foo"` to the only field to make sure none of its fields will be deferred.

Fix #455

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -283,11 +283,21 @@ class OptimizerStore:
             # "lookup was already seen with a different queryset" error
             qs = qs.prefetch_related(None).prefetch_related(*to_prefetch.values())
 
+        only_set = set(self.only)
+        select_related_only_set = set()
+
         if config.enable_select_related and self.select_related:
             qs = qs.select_related(*self.select_related)
 
-        if config.enable_only and self.only:
-            qs = qs.only(*self.only)
+            for select_related in self.select_related:
+                if select_related in only_set:
+                    continue
+
+                if not any(only.startswith(select_related) for only in only_set):
+                    select_related_only_set.add(select_related)
+
+        if config.enable_only and only_set:
+            qs = qs.only(*(only_set | select_related_only_set))
 
         if config.enable_annotate and self.annotate:
             to_annotate = {}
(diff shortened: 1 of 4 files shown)


patrick91 (approved):

Ivan-Vinitskyi:
@bellini666 should it works without `DjangoOptimizerExtension`?

bellini666:
@Ivan-Vinitskyi no, optimization hints only work when `DjangoOptimizerExtension` is enabled

PR #467, opened 2024-01-28
Extending subscription support

sdobbelaere:
As promised on https://github.com/strawberry-graphql/strawberry-graphql-django/issues/454 I'm hereby publishing my subscription implementation for [OneSila.  ](https://github.com/OneSila/OneSilaHeadless). 

What does it do:
- Allow you to expose every model to the frontend via subscriptions.
- Easy to extend and modify.

What it doesn't do:
- Support query and mutation extensions
- Follow any other internal lib safety features or restrictions. 

This PR is intended as first step towards comprehensive subscriptions support for models out of the box. The reason this wasn't published sooner is because this PR should be stamped as experimental.  

## Types of Changes
- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* https://github.com/strawberry-graphql/strawberry-graphql-django/issues/454 

## Todos:
- [ ] Investigate query/mutation extension support
- [ ] Create one liner field declarations
- [ ] More flexible access controle
- [ ] Improve extendability
- [ ] Tests
- [ ] Add receivers import into the app.py file

## Want to help?
- Test need figuring out as this is Django-channels based. 
- ACL and extensions are an important part of query/mutations. This probably needs investigating.
- model_subscriber() wrapper is a not a great approach towards extendability. 
- ...


strawberry_django/apps.py
```diff
@@ -4,3 +4,6 @@
 class StrawberryDjangoConfig(AppConfig):
     name = "strawberry_django"
     verbose_name = "Strawberry django"
+
+    def ready(self):
+        pass
```
bellini666, line 9: Is there something missing from here? Otherwise you might want to remove this

strawberry_django/receivers.py
```diff
@@ -0,0 +1,17 @@
+import contextlib.suppress
```
bellini666, line 1: Imoporting `contextlib` is enough. It will also solve the typing issue being raised

```suggestion
import contextlib
```

strawberry_django/subscriptions/utils.py
```diff
@@ -0,0 +1,11 @@
+def get_group(instance):
+    return f"{instance.__class__.__name__}"
```
bellini666, line 2: I don't think the class name is enough, as 2 models from different modules can have similar names.

You can get `instance.__class__._meta` and use `f"{meta.app_label}.{meta.model_name}"`

Also, maybe prefix it with `strawberry-subscription-{}` would be nice to make sure this doesn't conflict by mistake with anything else?

strawberry_django/subscriptions/receivers.py
```diff
@@ -0,0 +1,12 @@
...
+
+from .utils import get_group, get_msg
+
+
+def refresh_subscription_receiver(instance):
+    group = get_group(instance)
+    msg = get_msg(instance)
+
+    channel_layer = channels.layers.get_channel_layer()
+    async_to_sync(channel_layer.group_send)(group=group, message=msg)
```
bellini666, line 12: Maybe this should be `async` and the caller itself should use `async_to_sync`?

strawberry_django/subscriptions/publishers.py
```diff
@@ -0,0 +1,110 @@
...
+        self.model = model
+
+        self.ws = info.context["ws"]
+        self.channel_layer = self.ws.channel_layer
+
+    async def verify_logged_in(self):
+        user = get_current_user(self.info)
+
+        if user.is_anonymous:
+            raise ValidationError("Anynomous users are not allowed to subscribe.")
```
bellini666, line 41: Hrm, why not?

I mean, which user is allowed to subscribe should be defined by the application. Maybe it is an application which allows that?

What do you think about adding a way to check for permissions using the model_subscriber function? We can even try to use the permissioning system we already have for this (it might need some adjustments)

bellini666 (changes requested):
I've used similar solutions in some projects, and I like the way you organized this.

Left some suggestions. Let me know if I can help you with anything here (either ping me here or on discord)

cc @patrick91 what do you think of this?

strawberry_django/apps.py
sdobbelaere: Yes, looks like the receivers need to be added still.

strawberry_django/subscriptions/utils.py
sdobbelaere: Sounds like a plan.  I'll add that.

strawberry_django/subscriptions/receivers.py
sdobbelaere: I'll need to play with it. On that makes sense, since we're looking at async first.

strawberry_django/subscriptions/publishers.py
sdobbelaere: This is from the OneSila project, most endpoints are not available unless logged in.  I left it here as a reminder to investigate if we can leverage the existing extensions somehow.  Would prefer to stay away from a second set of permissions. 

Having said that, I also still want to reshuffle a few things in order to ensure the Publisher class is easy to extend and override for custom behaviour on the Subscription class.

PR #469, opened 2024-01-30
validate files at dummy-instance level

sdobbelaere:
Referring to issue: https://github.com/strawberry-graphql/strawberry-graphql-django/issues/468
This PR will validate the images at dummy-instance level.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

Fix https://github.com/strawberry-graphql/strawberry-graphql-django/issues/468

## Checklist

### DIFF
--- a/strawberry_django/mutations/resolvers.py
+++ b/strawberry_django/mutations/resolvers.py
@@ -338,6 +338,11 @@ def create(
         full_clean=full_clean,
         key_attr=key_attr,
     )
+    # Don't forget to add the files to the dummy_instance
+    # without saving to ensure it will also trigger
+    # in the full_clean method
+    for file_field, value in files:
+        file_field.save_form_data(dummy_instance, value)
 
     # Creating the instance directly via create() without full-clean will
     # raise ugly error messages. To generate user-friendly ones, we want
--- a/tests/models.py
+++ b/tests/models.py
@@ -67,7 +67,7 @@ class Fruit(FruitAbstract):
     pass
 
 
-class FruitWithRequiredPicture(models.Model):
+class FruitWithRequiredPicture(FruitAbstract):
     picture = models.ImageField(
         null=False,
         blank=False,
--- a/tests/models.py
+++ b/tests/models.py
@@ -26,7 +26,7 @@ class Vegetable(NameDescriptionMixin):
     world_production = models.FloatField()
 
 
-class FruitAbstract(models.Model):
+class Fruit(models.Model):
     name = models.CharField(max_length=20)
     color_id: Optional[int]
     color = models.ForeignKey(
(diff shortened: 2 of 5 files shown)


strawberry_django/mutations/resolvers.py
```diff
@@ -338,6 +338,11 @@ def create(
         full_clean=full_clean,
         key_attr=key_attr,
     )
+    # Don't forget to add the files to the dummy_instance
+    # without saving to ensure it will also trigger
+    # in the full_clean method
+    for file_field, value in files:
```
yergom, line 344: I think this is correct, as save_form_data is not committing the file to the django File Backend. (that happens only during save). However, I still think that the change is insufficient.

yergom: However, the code bellow:
```
instance = model._default_manager.create(**create_kwargs)
```

Is incorrect? during the create method, the instance will be ultimately saved. If you have a model with a ```FileField(blank=False, null=False),```, the insertion to the database will fail, as the field cannot be null. I think that the solution is to add the file as part of create_kwargs. Moreover, this also saves one db query.

yergom (changes requested):

strawberry_django/mutations/resolvers.py
sdobbelaere: Unless I'm wrong, files and direct data are actually saved in two steps.  
Take a look at the test I added @yergom, this one seems to succeed without errors. 

Interestingly enough, I couldn't actually replicate the issue without creating a strawberry input type that has the field missing.  This is because the input-type will already do the validation itself based on the settings of your model.   You'll see what I mean if you look at the test-code in details, in combination with the model and type settings.

yergom: Could it be that it works because the INSERT + UPDATE happen within a transaction, and the Null=False condition check is delayed until the transaction commits? But still...I do not understand the need for dividing direct data and file fields :S

bellini666:
Hey @sdobbelaere , thanks a lot for this PR :)

The code looks correct to me, but I'm not sure if there's something missing based on @yergom 's comments.

What do you 2 think?

yergom:
@bellini666 @sdobbelaere I think it's safe to be merged. Thanks for all the effort!

strawberry_django/mutations/resolvers.py
sdobbelaere: The direct data vs all fields is to ensure that in there is a separation that allows individual logic to set the files and m2m relations. 

This was split up because the update method has pre_save hooks which I wanted to keep in there.  This is from what I noticed used for creating users and setting passwords. 

As far as correction functionality, I believe it should be correct.  But I do agree that the test makes some assumptions based on the graphql response rather than inspecting the actual model instance itself.  

I'm a bit pre-occupied this week to extend the test, but if you're up for it. Feel free to make some adjustments. @yergom

bellini666:
@sdobbelaere anything missing from this or can I merge and release it?

PR #473, opened 2024-02-05
Deduce model field names from custom prefetches

diesieben07:
## Description

Currently the QuerySet optimizer is oblivious to custom `prefetch_related` like the following:
```python
@strawberry_django.type(models.Project)
class ProjectType(models.Project):
      @strawberry_django.field(
        prefetch_related=lambda _: Prefetch(
            "milestones",
            to_attr="next_milestones_pf",
            queryset=Milestone.objects.filter(due_date__isnull=False).order_by("due_date")
        )
    )
    def next_milestones(self) -> "list[MilestoneType]":
        return self.next_milestones_pf
```

The current optimizer cannot understand how to optimize "through" this prefetch, because it cannot find the model field (`milestones` in `Project`) that the field uses.
This results in the optimization essentially stopping at the `next_milestones` field and any further nested relations will cause N+1 issues.

This pull request attempts to address this shortcoming.

## Types of Changes

Now the optimizer first inspects the field configuration to find any custom `Prefetch`. A custom prefetch is one which has a custom `to_attr`  set. The optimizer then uses these custom Prefetches to perform further optimization instead of stopping here.

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry-graphql-django/issues/389

## Checklist

### DIFF
--- a/tests/projects/schema.py
+++ b/tests/projects/schema.py
@@ -103,6 +103,8 @@ class ProjectType(relay.Node):
     cost: strawberry.auto = strawberry_django.field(extensions=[IsAuthenticated()])
     is_small: strawberry.auto
 
+    next_milestones_property: strawberry.auto
+
     @strawberry_django.field(
         prefetch_related=lambda _: Prefetch(
             "milestones",
--- a/tests/projects/schema.py
+++ b/tests/projects/schema.py
@@ -115,7 +115,7 @@ class ProjectType(relay.Node):
         )
     )
     def next_milestones(self, root: Milestone) -> "List[MilestoneType]":
-        """The milestones for the project ordered by their due date"""
+        """Return the milestones for the project ordered by their due date."""
         if hasattr(root, "next_milestones_pf"):
             return cast(List[MilestoneType], root.next_milestones_pf)
         return root.milestones.filter(due_date__isnull=False).order_by("due_date")  # type: ignore
--- a/tests/projects/schema.py
+++ b/tests/projects/schema.py
@@ -1,7 +1,7 @@
 import asyncio
 import datetime
 import decimal
-from typing import Iterable, List, Optional, Type, cast, Union
+from typing import Iterable, List, Optional, Type, Union, cast
 
 import strawberry
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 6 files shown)


diesieben07:
Some guidance on the test failures would be great. I am not 100% sure what the correct approach would be for the changed GraphQL schema.

bellini666 (changes requested):
Wow, sorry for taking to long to review this. For some reason I started reviewing and stopped, had a some comments drafted...

> Some guidance on the test failures would be great. I am not 100% sure what the correct approach would be for the changed GraphQL schema.

Most of the issues I'm seeing here are related to typing. Try to run `pyright` locally and it will help you fix those. If you need specific guidance feel free to reach me on [discord](https://strawberry.rocks/discord) :)

There are some related to features not supported on older python versions. e.g. `X | Y` cannot be used yet because we are targeting python 3.8+, `Annotated` needs to be imported from `typing_extensions` (it was introduced on 3.9), etc

About the schema, just run pytest with `--snapshot-update` and it should update the schema file :)

diesieben07:
@bellini666 I think all the linting and type errors should be resolved now and the snapshot test fixed.

tests/projects/schema.py
```diff
@@ -296,6 +313,16 @@ class ProjectConnection(ListConnectionWithTotalCount[ProjectType]):
...
 
 
+ProjectFeedItem = Annotated[
+    Union[IssueType, MilestoneType], strawberry.union("ProjectFeedItem")
+]
+
+
+@strawberry.type
+class ProjectFeedConnection(relay.Connection[ProjectFeedItem]):
+    pass
```
bellini666, line 323: Is this a left over from a test attempt? =P

This is not being used, so it should be removed

bellini666 (changes requested):
The solution don't seem to be working for django 3.2 and 4.0 (see the tests that broke)

diesieben07:
I tried to revive this yesterday, but decided to abandon this approach. Instead I will look into creating a different pull request that makes using `optimize` inside a custom prefetch function easier.

PR #474, opened 2024-02-06
fix: cache definitions in optimizer

yergom:
Only 1 subtype of an interface types (e.g Node) is optimized. This PR fixes the optimizer's ```_interfaces`` cache, so that all interfaces are considered for optimization. I assume that the cache was introduced to avoid a performance hit (having to iterate over all the schema's types) every time an interface has to be optimized.

Before this PR, the following would happen. Suppose that we have A and B, both implementing the Node interface:
- We resolve node for an instance of A: The ``_interfaces`` cache is empty. The optimizer will add to the ``_interfaces`` cache only A, which will get all the optimization hints
- We resolve a node for an instance of B: The ``_interfaces`` cache contains only A. The optimizer will fail to optimize B.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -695,7 +695,9 @@ def optimize(
                 interface_definitions = []
                 for t in schema.schema_converter.type_map.values():
                     t_definition = t.definition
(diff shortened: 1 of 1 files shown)


bellini666 (approved):
LGTM, thanks for this! :)

PR #483, opened 2024-02-18
feat: Avoid joins when only fetching an ID

mselee:
## Description

Consider the following query
```gql
query Query {
  books {
    id
    author {
      id
    }
  }
}
```
A join will be generated even though we really don't have to, as the required information already exists through `book.author_id`.
The idea here is to try and optimize away this join, by constructing the `Author` object in-memory using the value from `book.author_id`.

I'm not very familiar with the code base, so I'm not sure if this is has other drawbacks.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


Sanyambansal76:
@mselee want to check if you are waiting for help to update the PR?

PR #493, opened 2024-03-04
fix: use str() to trigger eventual django's gettext_lazy string

fabien-michel:
The generate enum from choices feature fail when choice description is a django's lazy string (gettext_lazy used with translation framework) 

## Description

Use `str()` on description to trigger string translation when generating Enum.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

Fix #489 

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

Issue #489, opened 2024-02-26
Auto Enums does not works with gettext_lazy choice values


fabien-michel:
When a django choices is defined using gettext_lazy strings, it makes the "description is a string" check fail in 
[graphql.type.definition.GraphQLEnumValue](https://github.com/graphql-python/graphql-core/blob/430ee891a52d239adadf133d873fdca2a3fe009b/src/graphql/type/definition.py#L240)

The check `is_description(description)` return False if description is of type `django.utils.functional.lazy.<locals>.__proxy__`

### Context

Strawberry django's settings 
```python
STRAWBERRY_DJANGO = {
    "GENERATE_ENUMS_FROM_CHOICES": True,
}
```

### Failing test

in `tests/test_enums.py` 

Add choice with `gettext_lazy` description

```python
from django.utils.translation import gettext_lazy

class Choice(models.TextChoices):
    """Choice description."""

    A = "a", "A description"
    B = "b", "B description"
    C = "c", "C description"
    D = "d", gettext_lazy("D description")
```    

### Possible fix

As a fix I would suggest to change [this line](https://github.com/strawberry-graphql/strawberry-django/blob/17720f324e8876c41557d3cfb5ece45d6d59c880/strawberry_django/fields/types.py#L463) from:
```python
c[0]: EnumValueDefinition(value=c[0], description=c[1])
```

to
```python
c[0]: EnumValueDefinition(value=c[0], description=str(c[1]))
```

It would also be possible to change [this line](https://github.com/strawberry-graphql/strawberry-django/blob/17720f324e8876c41557d3cfb5ece45d6d59c880/strawberry_django/fields/types.py#L448) from

```python
choices = cast(List[Tuple[Any, str]], model_field.choices)
```

to 

```python
choices = [c[0]: str(c[1]) for c in model_field.choices]
```

I can propose a PR if you agree with one of these changes.

Thanks

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry-django/issues/489">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-django/issues/489/pledge.svg?darkmode=1">
[...]

PR #494, opened 2024-03-05
Fix/489 auto enum value allowed chars

fabien-michel:
Use regex to replace invalid characters in auto generated enums.

This occur because django's choice values are strings and can contain any chars.

Exemple This
```
class Choice(models.TextChoices):
   EN_US = "en-US", "English"
```

`Choice.choices` return `(("en-US", "English"),)`, and hyphen is not a valid char for GraphQL naming convention. So it would fail.
The PR replace non valid chars to `_`.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #489 
* #493

## Checklist


bellini666 (approved):
TY! :)

PR #504, opened 2024-03-19
fix: async with new filter API (assert queryset is wrong)

devkral:
This PR fixes a problem with the new filter API in async environments
It is likely a typo

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


devkral:
interesting: sourcery-ai doesn't filter removed lines. What it proposes is exactly what I am doing.

It is quite a mini fix but this bug breaks my code.

devkral:
we need to upgrade the severity of the issue: the wrong assertion will fail also for empty querysets, so it can cause major breakage

devkral:
I added some tests for the bug

devkral:
workaround: don't use the queryset argument

Issue #509, opened 2024-03-26
Custom filter/order method unexpected value


Kitefiko:
Value of custom filter method is being, by default, resolved via [`filters._resolve_value`](https://github.com/strawberry-graphql/strawberry-django/blob/3b4d753b01f7110d8be0f65c8b969d22e0dd6307/strawberry_django/filters.py#L98-L108)

```python
@strawberry_django.filter_field
def multi_filter(self, value: relay.GlobalID):
    # value here is not GlobalID instance, but already resolved node_id (str)
    # so following cannot be currently done
    
    if value.type_name == "FruitNode":
        ...
    else value.type_name == "VegetableNode":
        ... 
    else:
        raise ValidationError("Incorrect GlobalID type")
```
Solution: add flag to `resolve_value` param to `filter_field` that could be used to change behavior of value.

I believe __default behavior__ for custom filter or order methods should be to __not resolve__. Current behavior is confusing to users, not just for lack of documentation.

What should be default behavior and can be possibly do breaking change @bellini666 ?

I will create pull with changes shortly.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry-django/issues/509">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-django/issues/509/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry-django/issues/509/pledge.svg">
</picture>
</a>

PR #510, opened 2024-03-26
feat: filter_field optional value resolution

Kitefiko:
## Description
Adds option not to resolve value of custom filter method via `@strawberry_django.filter_field(resolve_value=False)`.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* https://github.com/strawberry-graphql/strawberry-django/issues/509

## Checklist

### DIFF
--- a/docs/guide/filters.md
+++ b/docs/guide/filters.md
@@ -209,10 +209,12 @@ class FruitFilter:
 
 !!! tip
 
+    #### process_filters
+
     As seen above `strawberry_django.process_filters` function is exposed and can be
     reused in custom methods. Above it's used to resolve fields lookups
 
-!!! tip
+    #### null values
 
     By default `null` value is ignored for all filters & lookups. This applies to custom
     filter methods as well. Those won't even be called (you don't have to check for `None`).
@@ -221,6 +223,12 @@ class FruitFilter:
 
     This also means that build in `exact` & `iExact` lookups cannot be used to filter for `None`
     and `isNull` have to be used explicitly.
+
+    #### value resolution
+
+    By default `value` parameter of type `relay.GlobalID` is resolved to its `node_id` attribute & `Enum`
+    to is's value (also applied to elements in list). This behavior can be disabled using
+    `strawberry_django.filter_field(resolve_value=False)`
 
 The code above generates the following schema:
 
--- a/docs/guide/filters.md
+++ b/docs/guide/filters.md
@@ -225,10 +225,15 @@ class FruitFilter:
     and `isNull` have to be used explicitly.
 
     #### value resolution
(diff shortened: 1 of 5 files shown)


bellini666 (approved):
LGTM!

Before merging, I'm wondering if we should avoid resolving the value for custom filters by default? What do you think?

bellini666:
@Kitefiko even looking at the example on the issue you created:

```python
@strawberry_django.filter_field
def multi_filter(self, value: relay.GlobalID):
    # value here is not GlobalID instance, but already resolved node_id (str)
    # so following cannot be currently done
    
    if value.type_name == "FruitNode":
        ...
    else value.type_name == "VegetableNode":
        ... 
    else:
        raise ValidationError("Incorrect GlobalID type")
```

It seems strange if `value` in this case is not actually `GlobalID`. i.e. the typing should be respected.

Do you think we can refactor this to basically only convert the value for automatic filters and pass it as is to custom resolvers, to make sure the type is respected?

bellini666 (changes requested):
Changing this to `Request changes` because of my last comment

Kitefiko:
Hi,
yeah, that was precisely my question. 
So just to clear this up for me. You suggest to do breaking change and stop resolving value in custom filters. I assume still leaving the ability to use param to do so, but make it not resolve by default?

bellini666:
> Hi, yeah, that was precisely my question. So just to clear this up for me. You suggest to do breaking change and stop resolving value in custom filters. I assume still leaving the ability to use param to do so, but make it not resolve by default? Another question I found during adjustments, should `RangeLookup` resolve its `start` and `end` fields?
> 
> Edit: Adjusted. Made `RangeLookup` resolve it's field.

Yes.

This is still a new change so I think it is safe

What I mean is that when I do something like this:

```python
@strawberry_django.filter_field()
def some_filter(self, value: relay.GlobalID, prefix: str) -> Q:
    ...
```

I expect that the `value` will be `GloabalID` as it is typed, and not its resolved value. Not `value` not being `GlobalID` just like it is typed there is a bug to IMO

Maybe we can have an option to opt-in to "automatic resolution" of the values, but TBH I think we don't even need that as it seems more confusing. What are your thoughts here?

Kitefiko:
Ok, so currently it is done as you describe BUT with opt-in `resolve_value` (more info in docs - I think it's pretty clear there). And also auto resolving for `RangeLookup` - still don't know, whether this is good choice.

I would probably leave `resolve_value` in? Don't really see reason to rip it out, there might be use-case (custom GlobalID?, different enum value resolution?) for custom Lookups where value resolution might be unwanted and this covers that.

bellini666 (approved):
Thank you! :)

PR #515, opened 2024-04-05
fix(optimizer): handle existing select_related in querysets

taobojlen:
## Description

The strawberry-django optimizer can automatically add `select_related` and `only` clauses to a queryset. However, it currently fails if you enable the `only` optimization while using querysets that already have `select_related`s on them.

We use strawberry at work, and have a model like:
```
class MyManager(models.Manager):
  def get_queryset(self):
    return super().get_queryset().select_related("foo")

class MyModel(models.Model):
  objects = MyManager()
  foo = models.ForeignKey(Foo, on_delete=models.CASCADE)
```

This causes issues with the optimizer if you try to execute a graphQL query that *doesn't* include `foo`: you get errors like `Field cannot be both deferred and traversed using select_related at the same time`. We've actually been running without the `only` optimization for a long time for this reason.

This PR adds some code to the optimizer that inspects the queryset for any existing `select_related`s, and includes them in the `.only()` call.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

N/A, there's no issue for this, though it related to [this old message](https://discord.com/channels/689806334337482765/1045715992086593597/1045715992086593597) in the Strawberry Discord.

## Checklist

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -285,18 +285,40 @@ class OptimizerStore:
 
         only_set = set(self.only)
         select_related_only_set = set()
+        select_related_set = set(self.select_related)
 
-        if config.enable_select_related and self.select_related:
-            qs = qs.select_related(*self.select_related)
+        # inspect the queryset to find any existing select_related fields
+        def get_related_fields_with_prefix(
+            queryset_select_related: dict[str, Any], prefix=""
+        ):
+            fields = []
+            for parent, nested in queryset_select_related.items():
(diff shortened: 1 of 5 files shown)


bellini666:
> Note: this is a draft PR! I am testing this out on our code, but I figured I'd open this draft so you're aware :)

Let me know once it is ready to be reviewed :)

bellini666 (approved):
Thank you for this! :)

PR #537, opened 2024-05-30
refactor: Use graphql-core's collect_sub_fields instead of our own implementation

bellini666:
This allows us to simplify our optimizer implementation, and also paves the way for nested optimizations implementation in the near future.

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -28,6 +28,8 @@ from django.db.models.fields.reverse_related import (
 )
 from django.db.models.manager import BaseManager
 from django.db.models.query import QuerySet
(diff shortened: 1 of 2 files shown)


strawberry_django/optimizer.py
```diff
@@ -473,18 +470,30 @@ def _get_model_hints(
...
+            info,
+            f_selections,
+            field_definition,
+            parent_type,
+        )
+
         # Add annotations from the field if they exist
         field_store = getattr(field, "store", None)
-        if field_store is not None:
+        if field_store:
```
bellini666, line 496: the store has a `__bool__` method that will return `False` if it is empty, which avoids the overhead of doing the `|=` below

patrick91 (approved):

PR #565, opened 2024-06-26
Generated fields type resolution

Mapiarz:
I added support for field type resolution for Django 5 `GeneratedFields`. For a generated field, the `output_field` is mandatory, and this is what I'm using to determine the strawberry field type.

Extended existing test about field type resolution that is backwards compatible with Django <5. Did not modify the project dependencies.

## Summary by Sourcery

This pull request introduces support for resolving field types for Django 5 `GeneratedFields` by utilizing the `output_field` attribute. It also extends existing tests to cover this new functionality while maintaining compatibility with earlier Django versions.

- **New Features**:
    - Added support for resolving field types for Django 5 `GeneratedFields` using the `output_field` attribute.
- **Enhancements**:
    - Extended existing tests to include field type resolution for `GeneratedFields`, ensuring backwards compatibility with Django versions prior to 5.

### DIFF
--- a/strawberry_django/fields/base.py
+++ b/strawberry_django/fields/base.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import functools
 from typing import TYPE_CHECKING, Any, Optional, TypeVar, cast
 
+import django
 from django.db.models import ForeignKey
 from strawberry import LazyType, relay
 from strawberry.annotation import StrawberryAnnotation
--- a/strawberry_django/fields/base.py
+++ b/strawberry_django/fields/base.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
 _QS = TypeVar("_QS", bound="models.QuerySet")
 
 try:
-    from django.db.models.fields.generated import GeneratedField
+    from django.db.models import GeneratedField  # type: ignore
 except ImportError:
     GeneratedField = None
 
@@ -207,7 +207,7 @@ class StrawberryDjangoFieldBase(StrawberryField):
                 self.origin_django_type,
             )
             if is_optional(
-                model_field.output_field
+                model_field.output_field  # type: ignore
                 if GeneratedField is not None
                 and isinstance(model_field, GeneratedField)
                 else model_field,
--- a/strawberry_django/fields/base.py
+++ b/strawberry_django/fields/base.py
@@ -211,7 +211,7 @@ class StrawberryDjangoFieldBase(StrawberryField):
                 model_field, GeneratedField
             )
             field_to_check = (
-                model_field.output_field if is_generated_field else model_field
+                model_field.output_field if is_generated_field else model_field  # type: ignore
             )
             if is_optional(
                 field_to_check,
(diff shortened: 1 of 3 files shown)


tests/fields/test_types.py
```diff
@@ -118,6 +141,17 @@ class Type:
         ("json", JSON),
     ]
 
+    if hasattr(models, "GeneratedField"):
```
bellini666, line 144: Maybe use `if django.VERSION >= (5, 0)` to make sure any typos in here would not make it pass by mistake?

strawberry_django/fields/types.py
```diff
@@ -38,6 +38,11 @@
     IntegerChoicesField = None
     TextChoicesField = None
 
+try:
+    from django.db.models import GeneratedField  # type: ignore
```
bellini666, line 42: Worth adding a comment here that `GeneratedField` is only available at Django 5.0+ so that we remember to remove it when dropping support for 4.2 in the future?

Another suggestion would be to change this to:

```python
if django.VERSION >= (5, 0):
    from django.db.models import GeneratedField
else:
    GeneratedField = None
```

This is self-documented, and is the kind of string that I would search for when searching for code that can be dropped/simplified after a version bump

strawberry_django/fields/base.py
```diff
@@ -202,7 +207,10 @@ def resolve_type(
                 self.origin_django_type,
             )
             if is_optional(
-                model_field,
+                model_field.output_field  # type: ignore
+                if GeneratedField is not None
+                and isinstance(model_field, GeneratedField)
+                else model_field,
```
bellini666, line 213: nit: add parenthesis around those to avoid them formatting in the same indentation, which IMO makes it confusing to see that everything is basically a single argument

obs. might need to run ruff again as it might adjust this differently

```suggestion
                (
                    model_field.output_field  # type: ignore
                    if (
                        GeneratedField is not None
                        and isinstance(model_field, GeneratedField)
                    )
                    else model_field
                ),
```

tests/fields/test_types.py
```diff
@@ -21,6 +22,12 @@
 
 import strawberry_django
 from strawberry_django.fields.field import StrawberryDjangoField
+from strawberry_django.type import _process_type  # noqa: PLC2701
+
+try:
+    from django.db.models import GeneratedField  # type: ignore
+except ImportError:
+    GeneratedField = None
```
bellini666, line 30:
```suggestion
if django.VERSION >= (5, 0):
    from django.db.models import GeneratedField  # type: ignore
else:
    GeneratedField = None
```

strawberry_django/fields/base.py
```diff
@@ -37,6 +37,11 @@
 
 _QS = TypeVar("_QS", bound="models.QuerySet")
 
+try:
+    from django.db.models import GeneratedField  # type: ignore
+except ImportError:
+    GeneratedField = None
```
bellini666, line 43:
```suggestion
if django.VERSION >= (5, 0):
    from django.db.models import GeneratedField  # type: ignore
else:
    GeneratedField = None
```

bellini666 (approved):

PR #570, opened 2024-06-29
Test optimizer with polymorphic relay node

stygmate:
Add node polymorphism tests and related schema
Update asserts in optimizer.py -> To be double checked before merge


stygmate:
@bellini666 here it is :)

bellini666 (approved):
Thanks @stygmate :)

I had to do a small modification from `isinstance(foo, A | B)` to `isinstance(foo, (A, B))` as the new syntax is only compatible with python 3.10+, and we are still supporting 3.8/3.9

Issue #603, opened 2024-07-29
Support multi-level nested objects create/update


EvSpirit:
In the latest strawberry-django version (0.47.1) single-level nested objects CUD operations work fine, however, multiple level nested objects cannot be created/updated. Consider the following types hierarchy: 

```
class Project(models.Model):
    foo = models.CharField(max_length=255)


class Milestone(models.Model):
    bar = models.CharField(max_length=255)
    project = models.ForeignKey(Project,  on_delete=models.PROTECT)  


class Issue(models.Model):
    baz = models.CharField(max_length=255)
    milestone = models.ForeignKey(Milestone, on_delete=models.PROTECT)  
```

### Case 1. Issue - Milestone - Project

When trying to execute the following mutation:
```
mutation CreateIssue ($input: IssueInput!) {
  createIssue (input: $input) {
    ... on IssueType {
      id
      milestone {
        id
        project {
          id
          name
        }
      }
    }
  }
}
```
input:
```
"input": {
    "foo": "Some Issue",
    "milestone": {
        "bar": "Some Milestone",
        "project": {
            "baz": "Some Project",
        },
    },
}
```

Error occurs:
```
ValueError: Cannot assign "UNSET": "Milestone.project" must be a "Project" instance.
```

That happens because of two issues:

* `_parse_data` function check for `None` pk value but not `UNSET`:
https://github.com/strawberry-graphql/strawberry-django/blob/9025ee5a0af7c31ce8fffe2be9ce9d871b289a17/strawberry_django/mutations/resolvers.py#L102-L107

So `UNSET` added to parsed data. After fixing if-clause and checking for both `None` and `UNSET` values another issue arises:

* Incorrect `model` used to create dummy instance and full-clean one

The problem is in the same function:
https://github.com/strawberry-graphql/strawberry-django/blob/9025ee5a0af7c31ce8fffe2be9ce9d871b289a17/strawberry_django/mutations/resolvers.py#L102-L103

parent model (`Milestone` but not `Project`) is used, so even if full clean passes somehow (in unit tests that's the case as all models have `name` field), finally incorrect instance would be created. The fix seems to be quite straightforward also:

```
related_model = get_model_fields(model).get(k).related_model
v = create(info, related_model, v.data or {}) 
```

Using parent model we can extract related model and pass one to `create` function. 

### Case 2. Project - Milestone - Issue (reverse relations)

```
mutation UpdateProject ($input: ProjectInputPartial!) {
[...]

PR #362, opened 2023-09-12
fix: Update related objects with unique_together

zvyn:
## Description

Resolves the issue described in #360: Support updating related objects with `unique_together` on the foreign key where the data identifies a single object in the database.

## Types of Changes

- [x] Bugfix
- [x] Enhancement/optimization

## Issues Fixed or Closed by This PR

* #360 

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

Issue #360, opened 2023-09-12
Updating related objects with `unique_together` including the FK raises IntegrityError


zvyn:
Consider the following setup:

```python
class Parent(models.Model):
    pass

class Child(models.Model):
    name = models.CharField()
    parent = models.ForeignKey(Parent, on_delete=models.CASCADE, related_name="children")

    class Meta:
        unique_togetehr = [("name", "parent")]

@strawberry_django.input(Child)
class ChildInput:
    name: auto

class ParentType:
    id: auto

@strawberry_django.input(Parent)
class ParentInput:
    children = Optional[list[ChildInput]]

@strawberry.type
class Mutation:
    update_parent: ParentType = strawberry_django.mutations.create(types.ParentInput)
```

Assuming a `Parent` with ID 1 but no child exists this works on the first attempt only:

```graphql
mutation {
  updateParent(data: {id: 1, children: [{name: "foo"}]}) {
    id
  }
}
```

On the second attempt an error is raised:

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "child_parent_id_name_12345678_uniq"
```

This is due to the `update_m2m` creating new objects, even if the data is uniquely identifying an existing relation.

Proposed solution: Use `manager.get_or_create` instead of ` manager.create`.

PR #604, opened 2024-07-29
Fix nested fields update

EvSpirit:
## Description

This PR addresses issues described in #603. Kindly ask you to review the code and share your ideas.

While working on these two major bugs I also faced another issue related to previously added Django `get_or_create` shortcut (see PR #362). The current implementation seems a bit inconsistent for me because `get_or_create` is used only in m2m relations, more than that, this shortcut relies not only on unique fields.  So it definitely fixes the issue mentioned in #360 , but at the same time there may be an opposite situation when duplicated records won't be created. First I thought to identify all unique fields and constraints (e.g. using Django `total_unique_constraints`) to further understand if `get` should be used before creating an object. But it seems there are  quite many caveats. So what I'm suggesting is to have the following convention: if an object input contains explicitly passed `None` / `null` pk value (not `UNSET`), it would indicate an intention to force create an object. If omitted, `get_or_create` approach will be applied (actually, it's a bit different under the hood, I refactored `full_clean` usage to be invoked only after `get()`, otherwise, `full_clean` may trigger unique constraint violation).

Personally I think that it should not be hardcoded somewhere in the lib and this approach looks only as a workaround for me, probably it's a good idea to further extend input types or probably integrate forms to precisely control fields cleaning and saving, that's for further discussion and may be a a good feature request (as while working on the issue I noticed that `full_clean` options usability may be limited in case of nested relations). 

It the current idea looks good, I think it would be better to update docs respectively.

Other than proposed fixes for the issues mentioned above, there is some minor code refactoring and cleanup:
* `full_clean` options exposed to `DjangoMutationCUD`
* some unused code in tests was removed, inherited schema definition also looked duplicated
* A couple of tests to cover #362 added

Please let me know if it is better to move out such changes to another PR.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #603 

## Checklist

- [x] My change requires a change to the documentation. (probably - to be discussed)

## Summary by Sourcery

This pull request fixes issues with nested fields update, refines the `full_clean` process, and enhances the handling of `get_or_create` for better consistency. It also includes new tests to cover these changes and removes redundant test code.

- **Bug Fixes**:
    - Resolved issues with nested fields update, ensuring proper handling of unique constraints and nested relations.
- **Enhancements**:
    - Refactored the `full_clean` usage to be invoked only after `get()` to prevent unique constraint violations.
    - Exposed `full_clean` options to `DjangoMutationCUD` for better control over validation.
    - Improved handling of `get_or_create` to support explicit `None`/`null` pk values for forced object creation.
- **Tests**:
    - Added tests for multiple-level nested m2m creation and list input creation.
    - Included tests for nested unique together update mutation to ensure no unique constraint failures.
    - Removed unused code in tests and cleaned up inherited schema definitions.


strawberry_django/mutations/resolvers.py
```diff
@@ -99,8 +99,9 @@ def _parse_data(
                 continue
 
             if isinstance(v, ParsedObject):
-                if v.pk is None:
-                    v = create(info, model, v.data or {})  # noqa: PLW2901
+                if v.pk in (None, UNSET):  # noqa: PLR6201
+                    related_model = get_model_fields(model).get(k).related_model
+                    v = create(info, related_model, v.data or {})  # noqa: PLW2901
```
EvSpirit, line 104: Seems like `pk` may be either `UNSET` or `None`, in both cases it's necessary to create nested object. And to further properly perform `full_clean` and `save` we should identify related model (as in test type hierarchy most models use `name` clashing field names, cleaning could pass somehow, but finally object of different type could be created).

strawberry_django/mutations/resolvers.py
```diff
@@ -178,6 +179,9 @@ def parse_input(
         )
 
     if isinstance(data, (OneToOneInput, OneToManyInput)):
+        if not data.set:
+            return None
+
```
EvSpirit, line 184: ListInput set may be explicitly passed as `null`, in such case no need to further parse data. E.g.:

https://github.com/strawberry-graphql/strawberry-django/blob/9025ee5a0af7c31ce8fffe2be9ce9d871b289a17/tests/mutations/test_relationship.py#L31-L42

strawberry_django/mutations/resolvers.py
```diff
@@ -366,16 +376,24 @@ def create(
...
+        # right after. The idea is to bypass cleaning first to allow getting
+        # an object by e.g. unique fields. In case `full_clean()` invoked before,
+        # it would raise unique constraint validation error
+        instance = model._default_manager.get(**create_kwargs)
+    except model.DoesNotExist:
+        # Creating the instance directly via create() without full-clean will
+        # raise ugly error messages. To generate user-friendly ones, we want
+        # full-clean() to trigger form-validation style error messages.
+        full_clean_options = full_clean if isinstance(full_clean, dict) else {}
         dummy_instance.full_clean(**full_clean_options)  # type: ignore
```
EvSpirit, line 392: Instead of using `get_or_create` shortcut I finally came up with a decision to use `get` and `create` separately, but invoke `full_clean` only after `get()` is invoked as `full_clean` may raise unique constraint violation exception (e.g. in use case described in #360 )

strawberry_django/mutations/resolvers.py
```diff
@@ -552,16 +570,33 @@ def update_m2m(
 
     use_remove = True
     if isinstance(field, ManyToManyField):
-        manager = cast("RelatedManager", getattr(instance, field.attname))
+        remote_field_name = field.attname
+        reverse_field_name = field.remote_field.related_name
```
EvSpirit, line 574: Have some doubts here, probably there is any other way to get reverse relationship name.

tests/projects/models.py
```diff
@@ -171,6 +181,21 @@ def name_with_priority(self) -> str:
         return f"{self.kind}: {self.priority}"
 
 
+class Version(NamedModel):
```
EvSpirit, line 184: I tried not to further extend types hierarchy, but this one is required to cover #362 changes

tests/projects/models.py
```diff
@@ -202,14 +227,16 @@ class Meta:  # type: ignore
...
-class Tag(NamedModel):
+class Tag(models.Model):
     issues: "RelatedManager[Issue]"
 
     id = models.BigAutoField(
         verbose_name="ID",
         primary_key=True,
     )
 
+    name = models.CharField(max_length=255, unique=True)
```
EvSpirit, line 238: Also related to #362 . Not only `unique_together` may be the case but also a single unique field.

tests/projects/schema.py
```diff
@@ -282,13 +296,17 @@ class TagInputPartial(NodeInputPartial):
...
     priority: strawberry.auto
     kind: strawberry.auto
     tags: Optional[List[NodeInput]]
     extra: Optional[str] = strawberry.field(default=UNSET, graphql_type=Optional[int])
 
 
+@strawberry_django.input(Issue)
+class IssueInputWithMilestones(IssueInput):
+    milestone: "MilestoneInputPartial"
+
```
EvSpirit, line 308: Workaround not to expose milestone related fields to inherited schema (as related input types were updated to cover nested CUD operations): https://github.com/strawberry-graphql/strawberry-django/blob/9025ee5a0af7c31ce8fffe2be9ce9d871b289a17/tests/projects/test_schema.py#L21-L34

strawberry_django/mutations/resolvers.py
```diff
@@ -655,11 +686,7 @@ def update_m2m(
                 data.pop(key_attr, None)
                 to_add.append(obj)
             elif data:
-                if key_attr not in data:
-                    manager.get_or_create(**data)
-                else:
-                    data.pop(key_attr)
-                    manager.create(**data)
+                create_nested(data)
```
EvSpirit, line 689: I don't fully understand the previous implementation introduced in #449 , why do we need to check input type and use `get_or_create` or just `create`? I tried to unify it regardless the input type, anyway, could you please pay attention to this part of code before changes?

tests/test_input_mutations.py
```diff
@@ -1116,21 +1435,21 @@ def test_mutation_full_clean_with_kwargs(db, gql_client: GraphQLTestClient):
 
     res = gql_client.query(
         query,
-        {"input": {"title": "ABC", "fullCleanOptions": True}},
+        {"input": {"pk": None, "title": "ABC", "fullCleanOptions": True}},
```
bellini666, line 1438: In those tests here, what would happen without this change?

tests/projects/test_schema.py
```diff
@@ -19,28 +14,5 @@ def test_schema(snapshot: Snapshot):
...
-    class Mutation:
-        create_issue: IssueType = mutations.create(
-            IssueInputSubclass,
-            handle_django_errors=True,
-            argument_name="input",
-        )
-
-    schema = strawberry.Schema(query=Query, mutation=Mutation)
     snapshot.snapshot_dir = SNAPSHOTS_DIR
-    snapshot.assert_match(str(schema), "schema_with_inheritance.gql")
```
bellini666, line 46: Why did you have to change this?

bellini666 (commented):
Thanks for the PR :)

I still need to take another better look at it (it is huge =P), but left some comments about some stuff that is worrying me

strawberry_django/mutations/resolvers.py
```diff
@@ -366,16 +376,24 @@ def create(
...
-    full_clean_options = full_clean if isinstance(full_clean, dict) else {}
-    if full_clean:
+    try:
+        # Instead of using `get_or_create` shortcut first use `get()`
+        # (which will also raise `DoesNotExist` in case pk `None` value explicitly
+        # passed), if no records found - invoke `full_clean()` and `create`
+        # right after. The idea is to bypass cleaning first to allow getting
+        # an object by e.g. unique fields. In case `full_clean()` invoked before,
+        # it would raise unique constraint validation error
+        instance = model._default_manager.get(**create_kwargs)
```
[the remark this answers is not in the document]
EvSpirit: I've checked Django code for unique fields validations, there are a few helpful fragments:

1) This one covers `unique_together` and `total_unique_constraints`
https://github.com/django/django/blob/9cf9c796be8dd53bc3b11355ff39d65c81d7be6d/django/contrib/admin/views/main.py#L473-L484

2) `unique` field  + null check
https://github.com/django/django/blob/9cf9c796be8dd53bc3b11355ff39d65c81d7be6d/django/contrib/admin/views/main.py#L441-L445

So what we can try to do is to extract all unique fields using these three "sources" (`unique`, `unique_together` and `UniqueConstraint`), what do you think?

tests/projects/test_schema.py
bellini666: Oh, that `schema_inherited.py` is not actually used anywhere (maybe it is a left over from some old attempt)

I think the ideal here is to remove that file and have the test "as is", as it is usually better to have everything defined inside the test itself.

I'll send a commit to main removing that file

strawberry_django/mutations/resolvers.py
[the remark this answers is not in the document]
bellini666: Although I do like the "identify the unique sources" idea, I'm worried if that could end up reusing some object where we were supposed to create a new one (and if it was unique, the user would maybe expect the code to break)

I'm thinking about a fourth option: What about expanding `key_attr` (or creating another field and deprecating it) to tell the list of fields/field combinations that are to be considered "keys"? Just like `unique_together = [("foo", "bar"), ("bin",)]` for the django model

In that example, someone could only pass `"bin"` to act like "get_or_create", but `("foo", "bar")` would still try to only create, etc.

axieum:
I'd love to see this merged. Happy to help where I can. 😄

bellini666:
> I'd love to see this merged. Happy to help where I can. 😄

I'm mostly waiting for the tests to pass and the adjustment to the conflict to review this again.

Not sure if @EvSpirit is going to tackle those, but if not, I do appreciate the help to finish this PR and merge it :)

bellini666:
🧹 closing this as it is very outdated

Feel free to ping me if you want this reopened, or simply open a new PR 😊

PR #622, opened 2024-09-03
Fix calculation of `has_next_page` in `resolve_connection_from_cache`

### LINKED ISSUE #613: Optimizer window pagination
## Describe the Bug
I believe I have found multiple issues with window pagination implementation.
- `AttributeError` is always thrown in `ListConnectionWithTotalCount.resolve_connection_from_cache`
- backup solution in case of `AttributeError` prevents pagination
- order and thus pagination itself is non-deterministic when explicit order is not used
- usage of `last` returns everything and bypasses `max_results`

## System Information

 - Operating system: linux
 - Strawberry version (if applicable): 0.47.1
 - Database version: 10.11.6-MariaDB

### `ListConnectionWithTotalCount.resolve_connection_from_cache`
`AttributeError` is always thrown here:
https://github.com/strawberry-graphql/strawberry-django/blob/d6051db6f1e792604d66206734df9d6ebcfbbb40/strawberry_django/relay.py#L195
I think there should be something like this?
```python
has_next_page = result[-1]._strawberry_row_number < result[-1]._strawberry_total_count if result else False
```

### Backup solution in case of `AttributeError` prevents pagination
This _might_ be intended behavior, but:
In case of `AttributeError` pagination falls back to `resolve_connection` but (probably due to unavailable overfetch?) `pageInfo`'s `hasNextPage` is wrongly reporting as `false`

### Order and thus pagination itself is non-deterministic when explicit order is not used
This is probably the biggest issue. Maybe fixable via `pk` ordering in case of empty `order_by` in `apply_window_pagination` ?

### Usage of `last` returns everything and bypasses `max_results`
When `last` is used on it's own and execution ends up here:
https://github.com/strawberry-graphql/strawberry-django/blob/d6051db6f1e792604d66206734df9d6ebcfbbb40/strawberry_django/optimizer.py#L561-L573

`slice_metadata.start` is evaluated to 0
`slice_metadata.end` becomes `sys.maxsize`

and queryset ends up filtering out nothing

SupImDos:
## Description

The `strawberry_django.relay.ListConnectionWithTotalCount.resolve_connection_from_cache` method currently incorrectly attempts to determine `has_next_page` from the `QuerySet._result_cache`, as opposed to the cached records within it, resulting in an `AttributeError` being raised.

This PR ensures that `has_next_page` is determined via the _last_ record in the `QuerySet._result_cache` (if applicable), so that we don't get an `AttributeError`

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Partially fixes #613 

## Checklist

## Summary by Sourcery

Fix the calculation of `has_next_page` in `resolve_connection_from_cache` to prevent an `AttributeError` by using the last record in the `QuerySet._result_cache`.

Bug Fixes:
- Fix the calculation of `has_next_page` in `resolve_connection_from_cache` to prevent an `AttributeError` by correctly determining it from the last record in the `QuerySet._result_cache`.

### DIFF
--- a/strawberry_django/relay.py
+++ b/strawberry_django/relay.py
@@ -188,11 +188,15 @@ class ListConnectionWithTotalCount(relay.ListConnection[relay.NodeType]):
             for node in result
         ]
         has_previous_page = (
-            nodes[0]._strawberry_row_number > 1  # type: ignore
+            result[0]._strawberry_row_number > 1  # type: ignore
             if result
             else False
         )
-        has_next_page = result._strawberry_row_number < result if result else False
+        has_next_page = (
+            result[-1]._strawberry_row_number < result[-1]._strawberry_total_count  # type: ignore
+            if result
+            else False
+        )
 
         return cls(
             edges=edges,
--- a/tests/relay/test_nested_pagination.py
+++ b/tests/relay/test_nested_pagination.py
@@ -1,0 +1,62 @@
+import pytest
+from strawberry.relay import to_base64
+from strawberry.relay.types import PREFIX
+
+from strawberry_django.optimizer import DjangoOptimizerExtension
+from tests import utils
+from tests.projects.faker import IssueFactory, MilestoneFactory
+
+
+@pytest.mark.django_db(transaction=True)
+def test_nested_pagination(gql_client: utils.GraphQLTestClient):
+    # Nested pagination with the same arguments for the parent and child connections
+    query = """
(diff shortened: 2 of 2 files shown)


bellini666 (approved):
Ty! :)

Going to merge this even with the typing issue, as it is actually related to the newest release of pyright, and I even need to fix other unrelated code

PR #638, opened 2024-10-07
Add unit tests for partial input optional field behaviour in update mutations

SupImDos:
## Description

As noted in the module's docstring, this PR adds some unit tests for Strawberry-Django's handling of partial input optional fields in update mutations, specifically when their values are omitted or explicitly set to `null`, for different variations of model fields:

* Required fields
* Optional and nullable fields
* Optional and non-nullable fields
* Required foreign key fields
* Optional foreign key fields
* Many-to-many fields

These tests stem from the fact that the GraphQL type-system doesn't distinguish between optional and nullable. That is, type `T!` is both required and non-nullable (i.e., must be supplied and cannot be `null`), but type `T` is both optional and nullable (i.e., can be omitted and can be `null`).

These tests uncovered what I believe is a small bug in `update_m2m`, where a value of `None` wasn't being explicitly checked.

_For reference, we were going to add similar unit tests to our project, but thought they might be useful upstream._

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* N/A

## Checklist

## Summary by Sourcery

Add unit tests for handling partial input optional fields in update mutations, covering various model field types and ensuring correct behavior when values are omitted or set to null. Fix a bug in the `update_m2m` function to handle `None` values correctly.

Bug Fixes:
- Fix a bug in the `update_m2m` function where a value of `None` was not being explicitly checked, ensuring correct handling of `None` values.

Tests:
- Add comprehensive unit tests to verify the behavior of partial input optional fields in update mutations for various model field types, including required fields, optional and nullable fields, optional and non-nullable fields, required foreign key fields, optional foreign key fields, and many-to-many fields.
- Include tests for both standard and Relay GraphQL mutations to ensure consistent handling of omitted and null values across different field types.

### DIFF
--- a/strawberry_django/mutations/resolvers.py
+++ b/strawberry_django/mutations/resolvers.py
@@ -529,7 +529,7 @@ def update_m2m(
     key_attr: str | None = None,
     full_clean: bool | FullCleanOptions = True,
 ):
-    if value is UNSET:
+    if value in (None, UNSET):  # noqa: PLR6201
         return
 
     # FIXME / NOTE:  Should this be here?
--- a/tests/mutations/test_partial_updates.py
+++ b/tests/mutations/test_partial_updates.py
@@ -290,7 +290,7 @@ def test_foreign_key_required(mutation):
 
 
 def test_foreign_key_optional(mutation):
-    """Tests behaviour for a required foreign key field."""
+    """Tests behaviour for an optional foreign key field."""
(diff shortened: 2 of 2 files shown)


bellini666 (approved):
Thank you very much for this! ❤️

PR #642, opened 2024-10-13
feat: New Paginated generic to be used as a wrapped for paginated results

### LINKED ISSUE #408: Adding `totalCount` in non-relay pagination: suggestion regarding approaches
### Summary

We're using Strawberry Graphql Django in our app and we would need a `totalCount` attribute for our queries in order show it in the frontend. I had previously raised https://github.com/blb-ventures/strawberry-django-plus/issues/190 in which it was suggested to use a relay node for this purpose. 

However, considering our entire application now is built on top of non-relay nodes, we don't necessarily want to change everything to relay nodes at this point. However, we have figured out two approaches to include our custom pagination logic by overriding the `StrawberryDjangoField` class. Before committing to an approach, we wanted to get some feedback from the team as to which one seems to be more aligned with the library. 

### Problem Statement

We have create a `PaginatedConnection` class which we are using as a type for our query nodes to include the count metadata as follows.

```
@strawberry.type
class PaginationConnection(Generic[ModelReturnType]):
    page_info: PageInfo
    results: List[ModelReturnType]
    
#which is being used as
paginated_records: PaginationConnection[RecordNode] = CustomStrawberryField()
```
However, this breaks the strawberry django internal functions as it can't identify the `django_type` (and consequently `django_model`) properly anymore.

### Proposed Solutions

1. Override the `django_type field` in `CustomStrawberryField` -> Here we are overriding the property to modify the origin to be the `RecordNode` instead of `PaginationConnection`. (We saw that this approach is being followed to resolved the relay node type internally in `django_type`)

```
    # Because our return type for a paginated node is different than a regular StrawberryNode,
    # we need to figure our the actual StrawberryNode type in order to get results
    @functools.cached_property
    def django_type(self) -> type[WithStrawberryDjangoObjectDefinition] | None:
        # Storing this value so that it can be reassigned to self.type
        origin = self.type
        origin_node = None
        object_definition = get_object_definition(origin)

        if object_definition and issubclass(
            object_definition.origin, PaginationConnection
        ):
            origin_node = object_definition.type_var_map.get("ModelReturnType")

            if isinstance(origin, LazyType):
                origin_node = origin.resolve_type()

        self.type = origin_node
        result_type = super().django_type
        self.type = origin
```

2. Passing the Django model explicitly in `CustomStrawberryField` and then setting the internal `django_model` in the `__init__` function. This will directly set the `django_model` no matter what the result of `django_type is.

```
    def __init__(
        self,
        django_model: type[models.Model],
        result_formatter_class: Type[CustomResultFormatMixin],
        *args,
        **kwargs,
    ):
        self.django_model = django_model
        self.result_formatter = result_formatter_class()
        self.non_paginated_qset: QuerySet
        self.paginated_qset: QuerySet
        super().__init__(*args, **kwargs)
        
        
     #And then calling the field as 
     paginated_records: PaginationConnection[RecordNode] = CustomStrawberryField(django_model=Record)

```

As we aren't sure about any proposed future changes to the library's internal API, we wanted to be double sure we are making the correct decision in terms of customizing the `StrawberryDjangoField`. Please let us know what you folks think about it.

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry-graphql-django/issues/408">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-graphql-django/issues/408/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

bellini666:
Fix #408

- [x] Create `Paginated` generic type
- [x] Integrate `Paginated` with StrawberryDjangoField
- [x] Ensure it is working with the optimizer
- [x] Ensure it is working with permissions
- [x] Update the documentation

## Summary by Sourcery

Introduce a new Paginated generic type for handling paginated results, integrate it with StrawberryDjangoField, and update documentation and tests to support this new feature.

New Features:
- Introduce a new Paginated generic type to handle paginated results in a more structured way.

Enhancements:
- Integrate the Paginated generic with StrawberryDjangoField to support pagination in GraphQL queries.
- Ensure compatibility of the Paginated generic with existing optimizers and permission systems.

Documentation:
- Update documentation to include usage instructions and examples for the new Paginated generic type.

Tests:
- Add tests to verify the functionality of the Paginated generic type, including its integration with permissions and nested queries.

### DIFF
--- a/docs/guide/pagination.md
+++ b/docs/guide/pagination.md
@@ -12,20 +12,190 @@ An interface for limit/offset pagination can be use for basic pagination needs:
 @strawberry_django.type(models.Fruit, pagination=True)
 class Fruit:
     name: auto
+
+
+@strawberry.type
+class Query:
+    fruits: list[Fruit] = strawberry_django.field()
 ```
+
+Would produce the following schema:
+
+```graphql title="schema.graphql"
+type Fruit {
+  name: String!
+}
+
+type Query {
+  fruits(pagination: PaginationInput): [Fruit!]!
+}
+```
+
+And can be queried like:
 
 ```graphql title="schema.graphql"
 query {
   fruits(pagination: { offset: 0, limit: 2 }) {
     name
-    color
   }
 }
 ```
 
-There is not default limit defined. All elements are returned if no pagination limit is defined.
+The `pagination` argument can be given to the type, which will enforce the pagination
+argument every time the field is annotated as a list, but you can also give it directly
+to the field for more control, like:
 
-## Relay pagination
+```python title="types.py"
+@strawberry_django.type(models.Fruit)
+class Fruit:
+    name: auto
 
-For more complex scenarios, a cursor pagination would be better. For this,
-use the [relay integration](./relay.md) to define those.
+
+@strawberry.type
+class Query:
+    fruits: list[Fruit] = strawberry_django.field(pagination=True)
+```
+
+Which will produce the exact same schema.
+
+> [!NOTE]
+> There is no default limit defined. All elements are returned if no pagination limit is defined.
+
+## Paginated Generic
+
+For more complex pagination needs, you can use the `Paginated` generic, which alongside
+the `pagination` argument, will wrap the results in an object that contains the results
+and the pagination information, together with the `totalCount` of elements excluding pagination.
+
+```python title="types.py"
+from strawberry_django.pagination import Paginated
+
+
+@strawberry_django.type(models.Fruit)
+class Fruit:
+    name: auto
+
+
+@strawberry.type
+class Query:
+    fruits: Paginated[Fruit] = strawberry_django.field()
+```
+
+Would produce the following schema:
+
+```graphql title="schema.graphql"
+type Fruit {
+  name: String!
+}
+
+type PaginatedInfo {
+  limit: Int!
+  offset: Int!
+}
+
+type FruitPaginated {
+  pageInfo: PaginatedInfo!
+  totalCount: Int!
+  results: [Fruit]!
+}
+
+type Query {
+  fruits(pagination: PaginationInput): [FruitPaginated!]!
+}
+```
+
+Which can be queried like:
+
+```graphql title="schema.graphql"
+query {
+  fruits(pagination: { offset: 0, limit: 2 }) {
+    totalCount
+    pageInfo {
+      limit
+      offset
+    }
+    results {
+      name
+    }
+  }
+}
+```
+
+### Customizing the pagination
+
+Like other generics, `Paginated` can be customized to modify its behavior or to
+add extra functionality in it. For example, suppose we want to add the average
+price of the fruits in the pagination:
+
+```python title="types.py"
+from strawberry_django.pagination import Paginated
+
+
+@strawberry_django.type(models.Fruit)
+class Fruit:
+    name: auto
+    price: auto
+
+
+@strawberry.type
+class FruitPaginated(Paginated[Fruit]):
+    @strawberry_django.field
+    def average_price(self) -> Decimal:
+        if self.queryset is None:
+            return Decimal(0)
+
+        return self.queryset.aggregate(Avg("price"))["price__avg"]
+
+    @strawberry_django.field
+    def paginated_average_price(self) -> Decimal:
+        paginated_queryset = self.get_paginated_queryset()
+        if paginated_queryset is None:
+            return Decimal(0)
+
+        return paginated_queryset.aggregate(Avg("price"))["price__avg"]
+
+
+@strawberry.type
+class Query:
+    fruits: FruitPaginated = strawberry_django.field()
+```
+
+Would produce the following schema:
+
+```graphql title="schema.graphql"
+type Fruit {
+  name: String!
+}
+
+type PaginatedInfo {
+  limit: Int!
+  offset: Int!
+}
+
+type FruitPaginated {
+  pageInfo: PaginatedInfo!
+  totalCount: Int!
+  results: [Fruit]!
+  averagePrice: Decimal!
+  paginatedAveragePrice: Decimal!
+}
+
+type Query {
+  fruits(pagination: PaginationInput): [FruitPaginated!]!
+}
+```
+
+The following attributes/methods can be accessed in the `Paginated` class:
+
+- `queryset`: The queryset original queryset with any filters/ordering applied,
+  but not paginated yet
+- `pagination`: The `OffsetPaginationInput` object, with the `offset` and `limit` for pagination
+- `get_total_count()`: Returns the total count of elements in the queryset without pagination
+- `get_paginated_queryset()`: Returns the queryset with pagination applied
+
+## Cursor pagination (aka Relay style pagination)
+
+Another option for pagination is to use a
+[relay style cursor pagination](https://graphql.org/learn/pagination). For this,
+you can leverage the [relay integration](./relay.md) provided by strawberry
+to create a relay connection.
--- a/docs/guide/pagination.md
+++ b/docs/guide/pagination.md
@@ -63,8 +63,32 @@ class Query:
 
 Which will produce the exact same schema.
 
-> [!NOTE]
-> There is no default limit defined. All elements are returned if no pagination limit is defined.
+### Default limit for pagination
+
+The default limit for pagination is set to `100`. This can be changed in the
+[strawberry django settings](./settings.md) to increase or decrease that number,
+or even set to `None` to set it to unlimited.
+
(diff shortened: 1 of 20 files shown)


docs/guide/pagination.md
```diff
@@ -12,20 +12,190 @@ An interface for limit/offset pagination can be use for basic pagination needs:
...
+
+@strawberry.type
+class Query:
+    fruits: list[Fruit] = strawberry_django.field(pagination=True)
+```
+
+Which will produce the exact same schema.
+
+> [!NOTE]
+> There is no default limit defined. All elements are returned if no pagination limit is defined.
```
patrick91, line 62: should we enforce a default limit? (I'd say yes :D)

docs/guide/pagination.md
```diff
@@ -12,20 +12,190 @@ An interface for limit/offset pagination can be use for basic pagination needs:
...
+
+Would produce the following schema:
+
+```graphql title="schema.graphql"
+type Fruit {
+  name: String!
+}
+
+type Query {
+  fruits(pagination: PaginationInput): [Fruit!]!
```
patrick91, line 30: shall we show how the PaginationInput type too?

docs/guide/pagination.md
```diff
@@ -12,20 +12,190 @@ An interface for limit/offset pagination can be use for basic pagination needs:
...
+> There is no default limit defined. All elements are returned if no pagination limit is defined.
+
+## Paginated Generic
+
+For more complex pagination needs, you can use the `Paginated` generic, which alongside
+the `pagination` argument, will wrap the results in an object that contains the results
+and the pagination information, together with the `totalCount` of elements excluding pagination.
+
+```python title="types.py"
+from strawberry_django.pagination import Paginated
```
patrick91, line 71: I wonder if we should call this `OffsetPaginated` or something like that 🤔

docs/guide/pagination.md
```diff
@@ -12,20 +12,190 @@ An interface for limit/offset pagination can be use for basic pagination needs:
...
+
+        return self.queryset.aggregate(Avg("price"))["price__avg"]
+
+    @strawberry_django.field
+    def paginated_average_price(self) -> Decimal:
+        paginated_queryset = self.get_paginated_queryset()
+        if paginated_queryset is None:
+            return Decimal(0)
+
+        return paginated_queryset.aggregate(Avg("price"))["price__avg"]
```
patrick91, line 155: neat!

patrick91, line 155: neat!

docs/guide/pagination.md
```diff
@@ -12,20 +12,190 @@ An interface for limit/offset pagination can be use for basic pagination needs:
...
+
+
+@strawberry.type
+class Query:
+    fruits: FruitPaginated = strawberry_django.field()
```
patrick91, line 160: `strawberry_django.field()` is mandatory here, right?

strawberry_django/optimizer.py
```diff
@@ -574,6 +574,15 @@ def _optimize_prefetch_queryset(
             else:
                 mark_optimized = False
 
+        if isinstance(field.type, type) and issubclass(field.type, Paginated):
+            pagination = field_kwargs.get("pagination")
+            qs = apply_window_pagination(
```
patrick91, line 579: is there a way to customise the queryset? should we document it?

strawberry_django/optimizer.py
```diff
@@ -992,6 +1000,20 @@ def _get_model_hints(
             level=level,
         )
 
+    # In case this is a relay field, the selected fields are inside results selection
```
patrick91, line 1003: I think this comment is wrong (or at least the relay part is)

strawberry_django/pagination.py
```diff
@@ -59,12 +117,19 @@ def apply(
...
+            stop = start + pagination.limit
+            queryset = queryset[start:stop]
+        else:
+            queryset = queryset[start:]
 
     return queryset
 
 
+class _PaginationWindow(Window):
+    """Marker to be able to remove where clause at `get_total_count` if needed."""
```
patrick91, line 130: not sure I understand this comment :D

strawberry_django/pagination.py
```diff
@@ -116,6 +182,46 @@ def apply_window_pagination(
...
+        # the whole queryset with its original filters.
+        queryset = queryset._chain()  # type: ignore
+        queryset.query.where.children = [
+            child
+            for child in queryset.query.where.children
+            if (
+                not hasattr(child, "lhs")
+                or not isinstance(child.lhs, _PaginationWindow)
+            )
+        ]
```
patrick91, line 220: might be worth moving this to util function and test in various cases 😊

strawberry_django/pagination.py
```diff
@@ -177,8 +285,39 @@ def get_queryset(
         **kwargs,
     ) -> _QS:
         queryset = super().get_queryset(queryset, info, **kwargs)
+
+        # If this is `Paginated`, return the queryset as is as the pagination will
+        # be resolved when resolving its results.
+        if self.is_paginated:
+            return queryset
+
         return self.apply_pagination(
```
patrick91, line 294: mmh, this is a bit confusing 🤔

patrick91 (commented):
this is cool, I left some initial comments 😊

docs/guide/pagination.md
bellini666: I think we should, yes!

The idea in this PR was to add the `Paginated` only, without breaking the existing pagination we currently have, which would be a breaking change.

Do you think we should also do that here or in a following PR?

docs/guide/pagination.md
bellini666: hahaha forgot about it :)

docs/guide/pagination.md
bellini666: obs. this also works for relay connections (I've done some similar stuff in the past like this :)

docs/guide/pagination.md
bellini666: Yes it is, otherwise `@strawberry.type` wouldn't know that this needs to be set to a `StrawberryDjangoField` and not a `StrawberryField`.

something that we can maybe try to improve in the future?

strawberry_django/optimizer.py
bellini666: Hrm I haven't thought about that... good idea! Make it similar to the `Connection`

strawberry_django/pagination.py
bellini666: Let me improve this :)

docs/guide/pagination.md
bellini666: Done! :)

docs/guide/pagination.md
bellini666: Decided to go with `OffsetPaginated`, to be aligned with the `OffsetPaginationInput` that we already have :)

docs/guide/pagination.md
```diff
@@ -159,6 +159,77 @@ query {
...
+
+@strawberry_django.order(models.Fruit)
+class FruitOrder:
+    name: auto
+    price: auto
+
+
+@strawberry.type
+class Query:
+    @straberry.offset_paginated(OffsetPaginated[Fruit], order=order)
```
thclark, line 186: Typo here @bellini666 but this is really gorgeous, overall, I'm loving this feature.

docs/guide/pagination.md
```diff
@@ -159,6 +159,77 @@ query {
...
+    def fruits(self, only_available: bool = True) -> QuerySet[Fruit]:
+        queryset = models.Fruit.objects.all()
+        if only_available:
+            queryset = queryset.filter(available=True)
+
+        return queryset
+```
+
+This would produce the following schema:
```
thclark, line 195: I'm a huge fan of adding the resulting schema to the docs for each of these things, it's been a big point of frustration for me with the docs not seeing how it should come out.

bellini666: Thanks for the feedback! I'm planing on doing some improvements to the docs soon and I'll do more of this!

docs/guide/pagination.md
```diff
@@ -12,20 +12,306 @@ An interface for limit/offset pagination can be use for basic pagination needs:
...
+@strawberry.type
+class Query:
+    fruits: list[Fruit] = strawberry_django.field(pagination=True)
+```
+
+Which will produce the exact same schema.
+
+### Default limit for pagination
+
+The default limit for pagination is set to `100`. This can be changed in the
```
patrick91, line 68: 🙇‍♂️

patrick91 (approved):
Amazing stuff, the docs are 👌

Issue #635, opened 2024-10-02
`annotate` and other query optimizations fail in mutations


erwinfeser:
## Describe the Bug
Let's imagine we have two Django models: `Cage` and `Animal` where the second one has a `FK` to the first one. Then we have the following strawberry definition:
```python
@strawberry_django.type(models.Cage)
class Cage:
    id: int

    @strawberry_django.field(annotate={"total_animals": Count("animals")})
    def total_animals(
        self,
        root: models.Cage,
    ) -> int:
        return root.total_animals
```
It works great when I ask for `totalAnimals` in a query but it fails if I ask for the same field in a mutation. Examples:

It works fine
```gql
query MyQuery {
  cages {
    id
    totalAnimals
  }
}
```

It fails with the error `Error: "\'Cage\' object has no attribute \'total_animals\'"`
```gql
mutation MyMutation {
  updateCage(data: {id: 10}) {
    id
    totalAnimals
  }
}
```

## System Information

 - Operating system: GNU/Linux
 - strawberry-graphql==0.243.1
 - strawberry-graphql-django==0.48.0

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry-django/issues/635">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-django/issues/635/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry-django/issues/635/pledge.svg">
</picture>
</a>

erwinfeser:
@bellini666 a similar error raises if we use `prefetch_related` optimization with `to_attr` in the `Prefetch` definition, given its result is a list instead of a queryset.

bellini666:
I'm checking here, this is happening because we disable optimizations while doing CUD mutations to avoid unexpected issues inside it, but then we don't run the optimization again when returning the model.

The most straightforward fix would be to run the query to fetch the field again, which might lead to an extra query, but I'm not sure if we can do much better...

What do you think?

philipstarkey:
[...]

PR #674, opened 2024-12-15
fix(mutations): Refetch instances to optimize the return value

bellini666:
Refetch instances after the mutation to make sure we optimize the result.

Without this, `annotated` fields would not exist in the return value and would fail to be returned.

I'm also modifying a test to include an annotated field. Without this change applied it would fail.

Fix https://github.com/strawberry-graphql/strawberry-django/issues/635

## Summary by Sourcery

Refetch instances after mutations to ensure annotated fields are included in the return value, addressing issue #635. Introduce a refetch method to optimize the retrieval of instances post-mutation. Add a test case to verify the inclusion of the 'isDelayed' field in the mutation response.

Bug Fixes:
- Refetch instances after mutations to ensure annotated fields are included in the return value, addressing issue #635.

Enhancements:
- Introduce a refetch method to optimize the retrieval of instances post-mutation.

Tests:
- Add a test case to verify the inclusion of the 'isDelayed' field in the mutation response.

### DIFF
--- a/strawberry_django/mutations/fields.py
+++ b/strawberry_django/mutations/fields.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import inspect
-from typing import TYPE_CHECKING, Annotated, Any, Union
+from typing import TYPE_CHECKING, Annotated, Any, TypeVar, Union
 
 import strawberry
 from django.core.exceptions import (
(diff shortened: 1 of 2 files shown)


strawberry_django/mutations/fields.py
```diff
@@ -225,6 +227,32 @@ def arguments(self, value: list[StrawberryArgument]):
         args_prop = super(DjangoMutationBase, self.__class__).arguments
         return args_prop.fset(self, value)  # type: ignore
 
+    def refetch(self, resolved: _T, *, info: Info | None) -> _T:
+        if not DjangoOptimizerExtension.enabled or info is None:
```
g-as, line 231: `DjangoOptimizerExtension.enabled` is a `ContextVar`, I think `.get()` is missing here.

PR #675, opened 2024-12-15
fix(optimizer): Prevent issuing duplicated queries for certain uses of first() and get()

### LINKED ISSUE #662: Optimizer results are not used in case of one-to-many relations where just one object is used
## Describe the Bug
When you have a one-to-many relation in a model, so either a reverse lookup of a foreign key or a many to many field, the optimizer will prefetch it (good 👍). If the field in GraphQL however is declared as not returning a list or connection but just a single type, then the default resolver will use `QuerySet#first` / `QuerySet#get` to get the result, which will bypass the prefetched objects and issue new queries.

Example:
```python
class Project(models.Model):
  name = models.CharField(max_length=255)

class Issue(models.Model)
  project = models.ForeignKey(Project, related_name='issues')


@strawberry_django.type(Issue, name='Issue')
class IssueType:
  id: int


@strawberry_django.type(Project, name='Project')
class ProjectType:
  id: int
  name: str
  latest_issue: IssueType | None = strawberry_django.field(field_name='issues')


@strawberry.type
class Query:
  projects: ListConnectionWithTotalCount[ProjectType] = strawberry_django.connection()

```

This example seems a bit contrived, however it should show the bug. When querying as follows:
```graphql
query {
  projects {
    name latestIssue { id }
}
```
There will be a query for projects (👍), a prefetch-query for issues (👍) and then another query _for each project_ fetching the first issue (🙁). 

A fix could look like this, implemented as a custom `StrawberryDjangoField` subclass here:
```python
class StrawberryDjangoFieldWithFixedPrefetchInSomeCases(StrawberryDjangoField):
    def get_queryset_hook(self, info: Info, **kwargs):
        if self.is_connection or self.is_paginated or self.is_list:
            return super().get_queryset_hook(info, **kwargs)
        elif self.is_optional:
            def qs_hook(qs: QuerySet):  # type: ignore
                qs = self.get_queryset(qs, info, **kwargs)
                if is_optimized_by_prefetching(qs):
                    return next(iter(qs), None)
                else:
                    return qs.first()
            return qs_hook
        else:
            def qs_hook(qs: QuerySet):  # type: ignore
                qs = self.get_queryset(qs, info, **kwargs)
                if is_optimized_by_prefetching(qs):
                    try:
                        return next(iter(qs))
                    except StopIteration:
                        raise qs.model.DoesNotExist(
                            "%s matching query does not exist." % qs.model._meta.object_name
                        )
                else:
                    return qs.get()
            return qs_hook
```

I will submit a proper pull request if there are no objections.

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry-django/issues/662">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-django/issues/662/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

diesieben07:
For "many-type" fields (i.e. ManyToManyField or a reverse FK) when the corresponding GraphQL field has a singular result, StrawberryDjangoField will use `QuerySet#get` and `QuerySet#first` to produce the result. This prevents the QuerySet from using any prefetched data, because those methods do an implicit slice.
This PR adds a check to not call get/first in case the QuerySet is prefetched and instead mimic their behavior to prevent duplicated queries.

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

Fixes #662 

## Checklist

## Summary by Sourcery

Fix duplicated queries in StrawberryDjangoField by avoiding get() and first() on prefetched QuerySets and add tests to verify the fix.

Bug Fixes:
- Prevent duplicated queries by avoiding the use of get() and first() on prefetched QuerySets in StrawberryDjangoField.

Tests:
- Add tests to verify the behavior of prefetched QuerySets with singular GraphQL fields, ensuring no duplicated queries occur.


diesieben07:
@bellini666 I'm not sure why the type check is failing. What is "Error: Fields without default values cannot appear after fields with default values (reportGeneralTypeIssues)" referring to? Any guidance would be appreciated if you can.

bellini666 (approved):
ty! :)

I've made a small adjustment to use `qs._result_cache` instead of `len(qs)`. Although it should be ok, I think it's better to be safe than sorry 😅

diesieben07:
Thank you! :)

PR #679, opened 2024-12-22
Add new ordering method allowing ordering by multiple fields

### LINKED ISSUE #678: Improve support for ordering by multiple fields
Support a better API for ordering by multiple fields when using `@strawberry_django.order` by allowing it to be generated as a list of orderings.

## Feature Request Type

- [ ] Core functionality
- [X] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description

As stated in #677 the way `strawberry_django.order` distinguishes "order by name then birthday" vs "order by birthday then name" as follows:
```gql
order: {
  name: ASC
  birthday: ASC
}
```
vs. 
```gql
order: {
  birthday: ASC
  name: ASC
}
```

This is problematic, because usually object key order is considered to be not relevant. While object property order is now guaranteed in JavaScript and dict order is guaranteed in Python, this is still far from an ideal API, because usually object key order is considered irrelevant. Creating an object/dict with a required key order and then making sure that order is preserved when being serialized out to JSON can be challenging or might even be impossible in some languages, because the implementations might not preserve the order.
The JSON specification also states that (emphasis mine):
> An object is an _unordered_ set of name/value pairs.

This is also demonstrated by the hoops that the code needs to jump through to even _extract_ this information from the GraphQL Python implementation.

I propose adding an option to `@strawberry_django.order` named `multi`. If set to true, the `order` parameter would be generated as a _list_ instead, allowing the explicit specification of _multiple_ ordering objects.

diesieben07:
Add new `@strawberry_django.ordering()` annotation that supersedes `@strawberry_django.order()` offering better support for ordering by multiple fields.

**This is still a draft, because it is not finished, but I have added it here for feedback.**

## Description
Currently the behavior is almost exactly the same as for `@strawberry_django.order` except for the following changes:
- Top-level ordering argument is generated as a _list_ instead of an optional input. This allows better control over sorting by multiple fields (for example "sort by name, then by birthdate").
- `@strawberry_django.ordering()` exposes the `one_of` setting of `@strawberry.input()` and allows setting a default via the Strawberry Django settings. It is useful to set this to true, because sorting  by multiple fields should be given as a _list_ of objects with a single field set each. This way object property order is irrelevant for the sorting.
- `OrderSequence` type and its associated machinery is removed from `ordering`

## Still missing:
- Migration path from `order` to `ordering` in a schema-compatible way:
  This already works, you can just have both parameters, but we should validate that only one of them is set by clients and send a useful error message.
- Documentation
- Tests for custom order methods when using `ordering`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [X] New feature
- [ ] Documentation

Fixes #678.

## Checklist


strawberry_django/ordering.py
```diff
@@ -194,17 +195,102 @@ def parse_and_fill(field: ObjectValueNode, seq: dict[str, OrderSequence]):
...
+        queryset = queryset.order_by(*args)
+    return queryset
+
+
 class StrawberryDjangoFieldOrdering(StrawberryDjangoFieldBase):
-    def __init__(self, order: type | UnsetType | None = UNSET, **kwargs):
+    def __init__(
+        self,
+        order: type | UnsetType | None = UNSET,
+        ordering: type | UnsetType | None = UNSET,
```
bellini666, line 278: **suggestion:** I think it's worth to add a check/raise here in case both are defined

strawberry_django/ordering.py
```diff
@@ -214,6 +300,12 @@ def arguments(self) -> list[StrawberryArgument]:
             order = self.get_order()
             if order and order is not UNSET:
                 arguments.append(argument("order", order, is_optional=True))
+        if self.base_resolver is None:
```
bellini666, line 303: **question:** why is this `if` not the same as the above? shouldn't the logic to apply ordering be the same, only changing if we call `get_order` or `get_ordering`?

strawberry_django/ordering.py
bellini666: Also raise a `deprecation warning` when using `order`, suggesting to use `ordering` instead (same for the type definition)

strawberry_django/ordering.py
```diff
@@ -236,16 +328,82 @@ def get_order(self) -> type[WithStrawberryObjectDefinition] | None:
...
+            name=name,
+            one_of=one_of,
+            description=description,
+            directives=directives,
+        )
+
+    return wrapper
 
 
 @dataclass_transform(
```
bellini666, line 409: **todo:** worth decorating this with [@deprecated](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated) and alerting to use `ordering` instead

strawberry_django/settings.py
```diff
@@ -46,6 +46,9 @@ class StrawberryDjangoSettings(TypedDict):
     #: to set it to unlimited.
     PAGINATION_DEFAULT_LIMIT: Optional[int]
 
+    #: Whether ordering inputs are marked with oneOf directive by default.
+    ORDERING_DEFAULT_ONE_OF: bool
```
bellini666, line 50: **suggestion:** I don't think we need this option. IMO `one_of` should be `True` by default, and if the user wants it to be `False` you already added an option for that (but, in my mind, it doesn't make sense for it to not be `one_of`)

bellini666 (commented):
This is looking great, thanks for the work @diesieben07 :)

Left some suggestions. Some things I'm missing from this (probably you already have in your TODO:

- Add a test which checks the schema output of the ordering, especially to see the `oneOf` directive being used :)
- Update the documentation to reflect the new usage, and maybe keep the old one but in a `(legacy)` session?

strawberry_django/ordering.py
diesieben07: We need to be able to define both so that users can update their APIs in a backwards compatible way.

diesieben07:
@bellini666 I would appreciate another look by you!
I have added documentation, not sure if I have done it right, the content seems to be copied in the `strawberry.rocks` repo (and I tested it there so I can see the results), but I'm not sure what the workflow is for copying the content across (?). Please let me know if anything needs to be changed.

Additionally, I am not quite happy with how users now have to write `@strawberry_django.ordering.ordering` instead of just `@strawberry_django.ordering`. Unfortunately `strawberry_django.ordering` already exists as a file, so it can't _also_ be the decorator - can it? Do you have any ideas?

The failing test runs have a very strange error message which does not seem related to my changes (and earlier this evening they ran fine).

docs/guide/ordering.md
```diff
@@ -4,19 +4,22 @@ title: Ordering
 
 # Ordering
 
+`@strawberry_django.ordering.ordering` is an upgrade from the previous `@strawberry_django.order` implementation
+and allows sorting by multiple fields.
+
 ```python title="types.py"
-@strawberry_django.order(models.Color)
+@strawberry_django.ordering.ordering(models.Color)
```
bellini666, line 11: **thought:** should we expose this as `strawberry_django.ordering` only? I think it makes the API easier to use

docs/guide/legacy-ordering.md
```diff
@@ -0,0 +1,245 @@
+---
+title: Ordering
+---
+
+# Order (Legacy)
+
+The legacy `@strawberry_django.order` implementation is only provided for backwards compatibility.
+You should prefer [Ordering](ordering) instead.
```
bellini666, line 8: **thought:** Maybe have this as a warning at the beginning of the doc?

strawberry_django/ordering.py
```diff
@@ -235,16 +333,37 @@ def get_order(self) -> type[WithStrawberryObjectDefinition] | None:
...
         self,
         queryset: _QS,
         info: Info,
         *,
         order: WithStrawberryObjectDefinition | None = None,
+        ordering: list[WithStrawberryObjectDefinition] | None = None,
         **kwargs,
     ) -> _QS:
+        if order and ordering:
+            raise ValueError("Only one of ordering, order must be given.")
```
bellini666, line 361: **suggestion:** 

```suggestion
            raise ValueError("Only one of `ordering` or `order` must be given.")
```

bellini666:
> Additionally, I am not quite happy with how users now have to write @strawberry_django.ordering.ordering instead of just @strawberry_django.ordering. Unfortunately strawberry_django.ordering already exists as a file, so it can't also be the decorator - can it? Do you have any ideas?

I even commented about this =P. Yes it can, just import it on `strawberry_django.__init__.py` and it should work

> The failing test runs have a very strange error message which does not seem related to my changes (and earlier this evening they ran fine).

I see those tests failing from time to time, usually a retrigger fixes them

---

Left some suggestions, will take a look after those are fixed. Also, be sure to rebase on top of main

bellini666 (changes requested):

docs/guide/ordering.md
diesieben07: I would love to have that, but I am not sure how to make it work.
`strawberry_django.ordering` already refers to the module `strawberry_django/ordering.py`. Can `strawberry_django.ordering` be both a module _and_ a decorator? I can't see how. I tried updating `strawberry_django.__init__.py` as you suggested:
```djff
- from .ordering import Ordering, order, process_order
+ from .ordering import Ordering, order, ordering, process_order
```

But this stops any imports like this from working (which exist in strawberry_django already and I am sure exist in user code):
```python
from strawberry_django import ordering
annotation = ordering.Ordering
```

diesieben07: We could explore one of the pretty horrible hacks listed here:
https://stackoverflow.com/questions/1060796/callable-modules
What's your preference @bellini666?

diesieben07: > TBH I wouldn't mind the conflict. Like, strawbery_django.ordering would be the method, if you want something from the module you could do from strawberry_django.ordering import foo

Yes, but that's precisely what I am not sure how to accomplish. If `strawberry_django.ordering` is a function, then `from strawberry_django.ordering import foo` can't work, because `strawberry_django.ordering` is not a module, it's a function.

> Or even strawberry_django.ordering_type, as it would be more obvious what it is doing. Then we could also move strawberry_django.filter to strawberry_django.filter_type and deprecate the other one.

I do like this, as it makes it clear that you're defining a type. It would also solve the problem, as `ordering_type` would be the decorator and `ordering` would stay as the module it is currently.

If you're fine with that I would go ahead with this solution. Should I do the `filter` => `filter_type` change in this PR as well?

diesieben07:
@bellini666 Like we had discussed, I have renamed the new decorator to `order_type`. I went with `order_type` instead of `ordering_type` to align with the existing `order_field`.
This solves the naming collision between `strawberry_django.ordering` module and the decorator. Users can now import the decorator from `strawberry_django` or from `strawberry_django.ordering`.

I would appreciate another look :)

bellini666 (approved):
ty so much for this. Have been wanting something like this for a while now ❤️

PR #681, opened 2024-12-25
fix(optimizer): Fix nested pagination optimization for m2m relations

### LINKED ISSUE #650: Optimized nested pagination is broken for m2m fields
## Describe the Bug

Nested pagination through m2m fields is broken with the optimizer enabled. This appears to be because using `.prefetch_related()` with a `QuerySet` annotated with a `Window` function where `partition_by` is an m2m field causes an extra join and duplicate results.

## Reproducible Example

See a minimal reproducible example here:
* Branch: [`mre/nested-pagination-m2m`](https://github.com/BrightpathProgress/strawberry-django/tree/mre/nested-pagination-m2m)
* Commit: 5469f268abaae77697ac29bc3b62608005496761

In the unit test above, the expected result of the query is:

```json
{
  "tagConn": {
    "edges": [
      {
        "node": {
          "name": "Tag 1",
          "issues": {
            "edges": [
              {"node": {"name": "Issue 1"}},
              {"node": {"name": "Issue 2"}}
            ]
          }
        }
      },
      {
        "node": {
          "name": "Tag 2",
          "issues": {
            "edges": [
              {"node": {"name": "Issue 2"}},
              {"node": {"name": "Issue 3"}}
            ]
          }
        }
      }
    ]
  }
}
```

However, the test fails, and the actual result of the query is:

```json
{
  "tagConn": {
    "edges": [
      {
        "node": {
          "name": "Tag 1",
          "issues": {
            "edges": [
              {"node": {"name": "Issue 1"}},
              {"node": {"name": "Issue 2"}},
              {"node": {"name": "Issue 2"}}  // Duplicate!
            ]
          }
        }
      },
      {
        "node": {
          "name": "Tag 2",
          "issues": {
            "edges": [
              {"node": {"name": "Issue 2"}},
              {"node": {"name": "Issue 2"}},  // Duplicate!
              {"node": {"name": "Issue 3"}}
            ]
          }
        }
      }
    ]
  }
}
```

## System Information

 - Operating system: MacoS
 - Strawberry version: 0.247.0
 - Strawberry Django version: 0.49.1

## Additional Context

It appears that this is caused by the implementation of [`apply_window_pagination()`](https://github.com/strawberry-graphql/strawberry-django/blob/v0.49.1/strawberry_django/pagination.py#L68-L116).

As suggested above, when you annotate a `.prefetch_related()` `QuerySet` with a `Window` function and refer back to the other side of the m2m, you add another join - causing duplicate results. When Django introduced the ability to prefetch with sliced querysets (which Strawberry-Django isn't directly using) they had to do a bunch of hacks to combine calls to filter together (see [`QuerySet._next_is_sticky()`](https://github.com/django/django/blob/5.1.2/django/db/models/query.py#L1932) and [`_filter_prefetch_queryset()`](https://github.com/django/django/blob/5.1.2/django/db/models/fields/related_descriptors.py#L97)).

To demonstrate the difference directly with the Django ORM see: a6f4e31ba9692999e75d7fc1fc253892e5b3d634.

In the unit test above, Strawberry-Django produces the following SQL query for the paginated prefetch:

```sql
SELECT *
FROM
  (SELECT ("projects_issue_tags"."tag_id") AS "_prefetch_related_val_tag_id",
          "projects_issue"."name" AS "col1",
          "projects_issue"."id" AS "col2",
          ROW_NUMBER() OVER (PARTITION BY "projects_issue_tags"."tag_id"
                             ORDER BY "projects_issue"."id" ASC) AS "_strawberry_row_number",
          COUNT(1) OVER (PARTITION BY "projects_issue_tags"."tag_id") AS "_strawberry_total_count"
   FROM "projects_issue"
   LEFT OUTER JOIN "projects_issue_tags" ON ("projects_issue"."id" = "projects_issue_tags"."issue_id")
   INNER JOIN "projects_issue_tags" T4 ON ("projects_issue"."id" = T4."issue_id")
   WHERE T4."tag_id" IN (1,
                         2)
   ORDER BY "projects_issue"."id" ASC) "qualify"
WHERE "_strawberry_row_number" <= 2
ORDER BY "col2" ASC
```

Whereas Django produces this SQL query for the paginated prefetch:

```sql
SELECT "_prefetch_related_val_tag_id",
       "col1",
       "col2"
FROM
  (SELECT *
   FROM
     (SELECT ("projects_issue_tags"."tag_id") AS "_prefetch_related_val_tag_id",
             "projects_issue"."name" AS "col1",
             "projects_issue"."id" AS "col2",
             ROW_NUMBER() OVER (PARTITION BY "projects_issue_tags"."tag_id"
                                ORDER BY "projects_issue"."id" ASC) AS "qual0"
      FROM "projects_issue"
      INNER JOIN "projects_issue_tags" ON ("projects_issue"."id" = "projects_issue_tags"."issue_id")
      WHERE "projects_issue_tags"."tag_id" IN (1,
                                               2)
      ORDER BY "projects_issue"."id" ASC) "qualify"
   WHERE ("qual0" > 0
          AND "qual0" <= 2)) "qualify_mask"
ORDER BY "col2" ASC
```

**Note the extra `LEFT OUTER JOIN` which causes the issue.**

## Potential Solution

As noted in the [`apply_window_pagination()`](https://github.com/strawberry-graphql/strawberry-django/blob/v0.49.1/strawberry_django/pagination.py#L68-L116) docstring, Django 4.2+ actually [supports sliced `QuerySet`s in `.prefetch_related()` now](https://github.com/django/django/pull/15957).

The solution here may be to actually use Django's inbuilt support for `.prefetch_related()` `QuerySet` slicing. However, this means we can't annotate `_strawberry_total_count` onto the nodes anymore. It is likely that we would need to refactor that functionality onto the _parent_ records, maybe using a subquery count and `OuterRef`?

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry-django/issues/650">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry-django/issues/650/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

bellini666:
When annotating something from the relation into the prefetch queryset for a m2m relation, Django will mistakenly not reuse the existing join and end up resulting in the generation of spurious results.

There's an ongoing fix for this i this ticket: https://code.djangoproject.com/ticket/35677

This is monkey patching older versions of Django which doesn't contain the fix, and most likely won't (Django usually only backports security issues), to fix the issue.

Thanks @SupImDos for providing an MRE in the form of a test for this!

Fix #650

## Summary by Sourcery

Fix nested pagination optimization for many-to-many relations in Django 5.1 or older by monkey-patching the `_filter_prefetch_queryset` and `add_q` methods to reuse existing aliases and joins.

Bug Fixes:
- Fixed an issue where nested pagination with many-to-many relationships generated spurious results in older Django versions.

Enhancements:
- Applied a monkey patch to improve pagination performance for nested many-to-many relations in Django 5.1 or older.

Tests:
- Added a test case to verify the fix for nested pagination with many-to-many relationships.

### DIFF
--- a/strawberry_django/utils/patches.py
+++ b/strawberry_django/utils/patches.py
@@ -1,0 +1,78 @@
+import django
+from django.db import (
+    DEFAULT_DB_ALIAS,
+    NotSupportedError,
+    connections,
+)
+from django.db.models import Q, Window
+from django.db.models.fields import related_descriptors
+from django.db.models.functions import RowNumber
+from django.db.models.lookups import GreaterThan, LessThanOrEqual
+from django.db.models.sql import Query
+from django.db.models.sql.constants import INNER
+from django.db.models.sql.where import AND
+
+
+def apply_pagination_fix():
+    """Apply pagination fix for Django 5.1 or older.
+
+    This is based on the fix in this patch, which is going to be included in Django 5.2:
+    https://code.djangoproject.com/ticket/35677#comment:9
+
+    If can safely be removed when Django 5.2 is the minimum version we support
+    """
+    if django.VERSION >= (5, 2):
+        return
+
+    # This is a copy of the function, exactly as it exists on Django 4.2, 5.0 and 5.1
+    # (there are no differences in this function between these versions)
+    def _filter_prefetch_queryset(queryset, field_name, instances):
+        predicate = Q(**{f"{field_name}__in": instances})
+        db = queryset._db or DEFAULT_DB_ALIAS
+        if queryset.query.is_sliced:
+            if not connections[db].features.supports_over_clause:
+                raise NotSupportedError(
+                    "Prefetching from a limited queryset is only supported on backends "
+                    "that support window functions."
+                )
+            low_mark, high_mark = queryset.query.low_mark, queryset.query.high_mark
+            order_by = [
+                expr for expr, _ in queryset.query.get_compiler(using=db).get_order_by()
+            ]
+            window = Window(RowNumber(), partition_by=field_name, order_by=order_by)
+            predicate &= GreaterThan(window, low_mark)
+            if high_mark is not None:
+                predicate &= LessThanOrEqual(window, high_mark)
+            queryset.query.clear_limits()
+
+        # >> ORIGINAL CODE
+        # return queryset.filter(predicate)  # noqa: ERA001
+        # << ORIGINAL CODE
+        # >> PATCHED CODE
+        queryset.query.add_q(predicate, reuse_all_aliases=True)
+        return queryset
+        # << PATCHED CODE
+
+    related_descriptors._filter_prefetch_queryset = _filter_prefetch_queryset  # type: ignore
+
+    # This is a copy of the function, exactly as it exists on Django 4.2, 5.0 and 5.1
+    # (there are no differences in this function between these versions)
+    def add_q(self, q_object, reuse_all_aliases=False):
+        existing_inner = {
+            a for a in self.alias_map if self.alias_map[a].join_type == INNER
+        }
+        # >> ORIGINAL CODE
+        # clause, _ = self._add_q(q_object, self.used_aliases)  # noqa: ERA001
+        # << ORIGINAL CODE
+        # >> PATCHED CODE
+        if reuse_all_aliases:  # noqa: SIM108
+            can_reuse = set(self.alias_map)
+        else:
+            can_reuse = self.used_aliases
+        clause, _ = self._add_q(q_object, can_reuse)
+        # << PATCHED CODE
+        if clause:
+            self.where.add(clause, AND)
+        self.demote_joins(existing_inner)
+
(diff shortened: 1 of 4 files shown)


strawberry_django/utils/patches.py
```diff
@@ -0,0 +1,78 @@
...
+def apply_pagination_fix():
+    """Apply pagination fix for Django 5.1 or older.
+
+    This is based on the fix in this patch, which is going to be included in Django 5.2:
+    https://code.djangoproject.com/ticket/35677#comment:9
+
+    If can safely be removed when Django 5.2 is the minimum version we support
+    """
+    if django.VERSION >= (5, 2):
+        return
```
patrick91, line 25: nice!

patrick91 (approved):

PR #683, opened 2024-12-30
Update test scope to include django 5.1, update pre-commit hooks

kwongtn:
## Description

Update test scope & pre-commit hooks.
Replaced `django-mptt` with `django-tree-queries`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

## Summary by Sourcery

Extend support to Django 5.1 and update pre-commit hooks.

Enhancements:
- Updated Django to 5.1.
- Updated pre-commit hooks.

Tests:
- Updated tests to account for Django 5.1 changes.


.github/workflows/tests.yml
```diff
@@ -54,10 +55,26 @@ jobs:
...
-          # Django 5.0 only supports python 3.10+
+          # Django 4.2 only supports python 3.8-3.12
+          - django-version: 4.2.*
+            python-version: '3.13'
+          # Django 5.0 only supports python 3.10-3.13
           - django-version: 5.0.*
             python-version: '3.9'
+          - django-version: 5.0.*
+            python-version: '3.13'
+          # Django 5.1 only supports python 3.9+
```
bellini666, line 75: **typo:** 

```suggestion
          # Django 5.1 only supports python 3.10+
```

.github/workflows/tests.yml
```diff
@@ -54,10 +55,26 @@ jobs:
...
+            mode: std
+          - django-version: 4.2.*
+            python-version: '3.8'
+            mode: geos
         exclude:
-          # Django 5.0 only supports python 3.10+
+          # Django 4.2 only supports python 3.8-3.12
+          - django-version: 4.2.*
+            python-version: '3.13'
+          # Django 5.0 only supports python 3.10-3.13
```
bellini666, line 70: **typo:** 

```suggestion
          # Django 5.0 only supports python 3.10-3.12
```

.github/workflows/tests.yml
```diff
@@ -54,10 +55,26 @@ jobs:
...
+            python-version: '3.8'
+            mode: std
+          - django-version: 4.2.*
+            python-version: '3.8'
+            mode: geos
         exclude:
-          # Django 5.0 only supports python 3.10+
+          # Django 4.2 only supports python 3.8-3.12
+          - django-version: 4.2.*
+            python-version: '3.13'
```
bellini666, line 69: **question:** How does this works when we have both `include` and `exclude`?

pyproject.toml
```diff
@@ -177,7 +177,7 @@ max-nested-blocks = 7
 [tool.pyright]
 pythonVersion = "3.9"
 useLibraryCodeForTypes = true
-exclude = [".venv", "**/migrations", "dist", "docs"]
+exclude = [".venv", "**/migrations", "dist", "docs", "tests/relay/mptt/models.py"]
```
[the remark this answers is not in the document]
bellini666: Do we need to exclude the whole file for this? Can we just `# type: ignore` or `cast` in the lines that need that?

tests/relay/mptt/models.py
```diff
@@ -1,9 +1,132 @@
+import django
 from django.db import models
 from mptt.fields import TreeForeignKey
 from mptt.models import MPTTModel
+from mptt.models import MPTTModelBase as _MPTTModelBase
 
+if django.VERSION >= (5, 1):
```
bellini666, line 7: **question:** why is this needed?

Can you add a comment with the explanation, and also pointing out where this code came from?

bellini666 (commented):
Ty for this! :)

Left a couple of suggestions/questions

.github/workflows/tests.yml
kwongtn: As far as I know, they are not mutually exclusive. And according to [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude)

> All include combinations are processed after exclude. This allows you to use include to add back combinations that were previously excluded.

---

Exclude works by excluding anything that matches the declared params.
For example as of the above, it excludes Django 4.2 and Python 3.13, no matter if it is geo or standard.

As for include, I added it in as it'll be redundant to add Python 3.8 to the matrix, then exclude all except Django 4.2 😅

tests/relay/mptt/models.py
kwongtn: Turns out its not needed anymore 🥳
According to [django mptt repo](https://github.com/django-mptt/django-mptt), `django-mptt` is now unmaintained. It recommends using [django-tree-queries](https://github.com/feincms/django-tree-queries) instead.

Hence I tried using it, and guess what all tests pass 😁
Will be deprecating `django-mptt` and removing references in an upcoming PR

bellini666 (approved):
Ty! :)

PR #687, opened 2025-01-06
fix: skip empty choice value

fabien-michel:
Seems they are not supported by GraphQL

Having an empty choice make generate enum from choice feature crash, trying to use None as a string.

```
class Choice(models.TextChoices):
    A = "a", "A description"
    B = "b", "B description"
    __empty__ =  "Empty"
```

## Description

The fix, for now, is to skip empty value, so the behavior is the same as `TextChoiceField`. 
It make the schema building pass, but sadly, data with an empty value will still fail with `Enum 'MyAutoGeneratedEnum' cannot represent value: ''` and I'm note sure what to do with that.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

## Summary by Sourcery

Bug Fixes:
- Prevent crashes when generating enums from choices with empty values by skipping them.

### DIFF
--- a/strawberry_django/fields/types.py
+++ b/strawberry_django/fields/types.py
@@ -466,7 +466,11 @@ def resolve_model_field_type(
             meta = model_field.model._meta
 
             enum_choices = {}
-            for c in cast("Iterable[tuple[str, str]]", model_field.choices):
+            for c in cast("Iterable[tuple[str | None, str]]", model_field.choices):
+                # Skip empty choice (__empty__)
+                if not c[0]:
+                    continue
+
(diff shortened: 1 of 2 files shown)


fabien-michel:
Hmm, seems the problem is more than that.
Having value set to empty string, make GraphQL fail with `Enum 'MyAutoGeneratedEnum' cannot represent value: ''`

I'm not sure what to do here.

bellini666 (approved):

bellini666:
Ty! :)

PR #701, opened 2025-02-07
chore: Update mutations.py to expose the full_clean parameter

keithhackbarth:
This is a draft PR. Just looking for feedback.

We would like to control if clean is called. Correctly, it looks like the only option on the field but the mutation wrapper is missing the parameter. This change simply adds the parameter.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

## Summary by Sourcery

Enhancements:
- Allow controlling whether the clean method is called on mutations.


bellini666:
Hey @keithhackbarth ,

> We would like to control if clean is called. Correctly, it looks like the only option on the field but the mutation wrapper is missing the parameter. This change simply adds the parameter.

I think this change is ok! TBH, not sure why we were not exposing those in first place.

keithhackbarth:
@bellini666 — Thanks so much for following up here. I thought I had pushed the linting follow-up, but it looks like it never went through and fell off my radar. Really appreciate you closing this out and getting it merged!

bellini666:
> @bellini666 — Thanks so much for following up here. I thought I had pushed the linting follow-up, but it looks like it never went through and fell off my radar. Really appreciate you closing this out and getting it merged!

It happens :)

Thank you for the contribution ❤️

Issue #707, opened 2025-02-14
Add GraphQL type for "GeometryField" defined in GeoDjango


shmoon-kr:
## Feature Request Type

- [V] Alteration (enhancement/optimization) of existing feature(s)

## Description

Currently, strawberry_graphql supports GeoDjango. It supports several types such as Polygon and MultiPolygon, but it does not support Geometry type. Geometry type is an arbitrary spatial type. It can be a Polygon, or it can be a MultiPolygon. Adding Geometry type would provide a lot of convenience for users.

The next part of the code. /strawberry_django/fields/fields.py

```
    MultiLineString = strawberry.scalar(
        NewType("MultiLineString", tuple[LineString]),
        serialize=lambda v: v.tuple if isinstance(v, geos.MultiLineString) else v,
        parse_value=lambda v: geos.MultiLineString(*[geos.LineString(x) for x in v]),
        description="A geographical object that contains multiple line strings.",
    )

    MultiPolygon = strawberry.scalar(
        NewType("MultiPolygon", tuple[Polygon]),
        serialize=lambda v: v.tuple if isinstance(v, geos.MultiPolygon) else v,
        parse_value=lambda v: geos.MultiPolygon(
            *[geos.Polygon(*list(x)) for x in v],
        ),
        description="A geographical object that contains multiple polygons.",
    )

    field_type_map.update(
        {
            geos_fields.PointField: Point,
            geos_fields.LineStringField: LineString,
            geos_fields.PolygonField: Polygon,
            geos_fields.MultiPointField: MultiPoint,
            geos_fields.MultiLineStringField: MultiLineString,
            geos_fields.MultiPolygonField: MultiPolygon,
        },
    )
```

PR #709, opened 2025-02-14
Add support for the the general `Geometry` type

shmoon-kr:
## Description

Add Geometry type

## Types of Changes

- [V ] New feature

Fix https://github.com/strawberry-graphql/strawberry-django/issues/707

## Checklist

## Summary by Sourcery

Add the `Geometry` scalar type.

New Features:
- Add support for the `Geometry` type to represent arbitrary geographical objects.  This includes `Point`, `LineString`, `LinearRing`, `Polygon`, `MultiPoint`, `MultiLineString`, and `MultiPolygon` types.

Tests:
- Add tests for the `Geometry` type.


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Implement Geometry type for django.contrib.gis.db.models.GeometryField
```
bellini666, line 3: **issue:** unfortunately we don't have auto publishing in here :( (at least not yet). You can remove this file

strawberry_django/fields/types.py
```diff
@@ -344,6 +345,37 @@ def __hash__(self):
...
+        NewType(
+            "Geometry",
+            Union[
+                Point,
+                LineString,
+                LinearRing,
+                Polygon,
+                MultiPoint,
+                MultiLineString,
+                MultiPolygon,
```
bellini666, line 358: **question:** isn't this the same as just using `GEOSGeometry`? Which is a base of all of those

strawberry_django/fields/types.py
```diff
@@ -344,6 +345,37 @@ def __hash__(self):
...
+        if isinstance(
+            v,
+            (
+                geos.Point,
+                geos.LineString,
+                geos.LinearRing,
+                geos.Polygon,
+                geos.MultiPoint,
+                geos.MultiLineString,
+                geos.MultiPolygon,
```
bellini666, line 371: **thought:** ditto here

bellini666 (commented):
Looks good in general, left a couple of coments

RELEASE.md
Evan-Dabeeo: I suggest you to add CONTRIBUTING.md file to the project. I tried to find a contributor's guide but I couldn't find it. So I just followed the guide of the strawberry graphql project. I'll remove this file.

strawberry_django/fields/types.py
Evan-Dabeeo: I'm not exactly sure about this, but if you look at the existing code above, you'll see that it defined MultiPolygon as a tuple[Polygon], so I just defined it using subtypes in a similar way.

strawberry_django/fields/types.py
Evan-Dabeeo: Oh I just misunderstood your comment. I think your comment is correct since if isinstance(v, geos.Point) is true then isinstance(v, geos.Geometry) is also true. I'll update the PR.

bellini666 (approved):
Ty! :)

RELEASE.md
bellini666: Thanks for the suggestion! Yeah, we indeed need one here T_T

Issue #708, opened 2025-02-14
Add Geospatial Lookup filter.


shmoon-kr:
## Feature Request Type

- [V] Alteration (enhancement/optimization) of existing feature(s)

## Description

Currently, strawberry_graphql supports the following types provided by GeoDjango: Point, Polygon, and MultiPolygon.
However, it does not support a number of geospatial query filters based on these types.
For example, filters like intersects. Adding these filters would be very useful for GeoDjango users.

shmoon-kr:
I tried to add a filter for Geometry type in `strawberry_django/fields/filter_types.py` file like following.

```
try:
    from django.contrib.gis import geos
    from django.contrib.gis.db import models as geos_fields

except ImproperlyConfigured:
    # If gdal is not available, skip.
    pass
else:
    @strawberry.input
    class GeometryFilterLookup(Generic[T]):
        intersects: Optional[Annotated["Geometry", strawberry.lazy(".types")]] = UNSET

    type_filter_map.update(
        {
            geos_fields.PointField: GeometryFilterLookup,
            geos_fields.LineStringField: GeometryFilterLookup,
            geos_fields.PolygonField: GeometryFilterLookup,
            geos_fields.MultiPointField: GeometryFilterLookup,
            geos_fields.MultiLineStringField: GeometryFilterLookup,
            geos_fields.MultiPolygonField: GeometryFilterLookup,
            geos_fields.GeometryField: GeometryFilterLookup,
        },
    )
```

But when I run Graphql server and see the GeometryFilterLookup type, it's just look like the FilterLookupType and I cannot find intersects filter that I defined.

```
GeometryFilterLookup

Fields

[exact]: [Geometry]
Exact match. Filter will be skipped on null value
[isNull]: [Boolean]
Assignment test. Filter will be skipped on null value
[inList]: [[Geometry]
Exact match of items in a given list. Filter will be skipped on null value
[iExact]: [Geometry]
Case-insensitive exact match. Filter will be skipped on null value
[contains]: [Geometry]
Case-sensitive containment test. Filter will be skipped on null value
[iContains]: [Geometry]
Case-insensitive containment test. Filter will be skipped on null value
[startsWith]: [Geometry]
Case-sensitive starts-with. Filter will be skipped on null value
[iStartsWith]: [Geometry]
Case-insensitive starts-with. Filter will be skipped on null value
[endsWith]: [Geometry]
Case-sensitive ends-with. Filter will be skipped on null value
[...]

PR #710, opened 2025-02-18
Feature: Add geometry filters

shmoon-kr:
## Description

Added GeometryLookupFilter type so that spatial filters available in GeoDjango can be used in strawberry_django.

## Types of Changes

- [O] New feature

* https://github.com/strawberry-graphql/strawberry-django/issues/708

## Checklist

- [V] My code follows the code style of this project.
- [V] My change requires a change to the documentation.
- [V] I have updated the documentation accordingly.
- [V] I have read the CONTRIBUTING document.
- [V] I have tested the changes and verified that they work and don't break anything (as well as I can manage).


thclark:
I didn't even think  about this possibility - great work @shmoon-kr

strawberry_django/__init__.py
```diff
@@ -71,3 +72,13 @@
...
+
+from django.core.exceptions import ImproperlyConfigured
+
+try:
+    from .fields.filter_types import GeometryFilterLookup
+
+    __all__.append("GeometryFilterLookup")
+except ImproperlyConfigured:
+    # If gdal is not available, skip.
+    pass
```
bellini666, line 84: **suggestion:** Maybe we should always expose this in `__all__`, but make it `None` inside `filter_typers` for import errors

strawberry_django/fields/filter_types.py
```diff
@@ -123,3 +129,57 @@ class DatetimeFilterLookup(DateFilterLookup[T], TimeFilterLookup[T]):
     str: FilterLookup,
     uuid.UUID: FilterLookup,
 }
+
+
+try:
+    pass
```
bellini666, line 135: **todo:** hrm, I think this is missing something =P

strawberry_django/fields/types.py
```diff
@@ -557,6 +558,19 @@ def resolve_model_field_type(
     ):
         if using_old_filters:
             field_type = filters.FilterLookup[field_type]
+        elif type(
+            field_type
+        ) is ScalarWrapper and field_type._scalar_definition.name in (
```
bellini666, line 563: **suggestion:** might need a `# type: ignore` here

bellini666 (changes requested):
So far so good! :)

Waiting for some tests

strawberry_django/fields/types.py
Evan-Dabeeo: OK.

strawberry_django/fields/filter_types.py
Evan-Dabeeo: It seems that it was removed by auto fixes from pre-commit.com hooks. Originally there was a code to test whether geodjango is available. `from django.contrib.gis.geos import GEOSGeometry` And I think that it's a better idea to define a configuration variable that indicates geodjango is available or not. But I don't know where it should be located.

Evan-Dabeeo:
```
try:
    from django.contrib.gis.db import models as geos_fields

    GEOS_IMPORTED = True

    class GeosFieldsModel(models.Model):
        point = geos_fields.PointField(null=True, blank=True)
        line_string = geos_fields.LineStringField(null=True, blank=True)
        polygon = geos_fields.PolygonField(null=True, blank=True)
        multi_point = geos_fields.MultiPointField(null=True, blank=True)
        multi_line_string = geos_fields.MultiLineStringField(null=True, blank=True)
        multi_polygon = geos_fields.MultiPolygonField(null=True, blank=True)
        geometry = geos_fields.GeometryField(null=True, blank=True)

except ImproperlyConfigured:
    GEOS_IMPORTED = False
```
I found this in tests/models.py file. But I think it should be located in somewhere else.

shmoon-kr:
Hi. I found the following error when I was trying to run a test code. `poetry run pytest`
It seems something wrong when you release the most recent version. Please check.

  File "/Users/evan/PycharmProjects/strawberry-django/tests/polymorphism_inheritancemanager/models.py", line 2, in <module>
    from model_utils.managers import InheritanceManager
ModuleNotFoundError: No module named 'model_utils'

And one more question: I tried to write a test code which is run only if GeoDjango is available. But in a testing environment, GeoDjango seems not available so I cannot test geometry operations. Any suggestions?

bellini666:
> Hi. I found the following error when I was trying to run a test code. `poetry run pytest` It seems something wrong when you release the most recent version. Please check.
> 
> File "/Users/evan/PycharmProjects/strawberry-django/tests/polymorphism_inheritancemanager/models.py", line 2, in from model_utils.managers import InheritanceManager ModuleNotFoundError: No module named 'model_utils'
> 

That's a new dev dependency that got merged from a PR yesterday. Running `poetry install` should fix it for you =P

> And one more question: I tried to write a test code which is run only if GeoDjango is available. But in a testing environment, GeoDjango seems not available so I cannot test geometry operations. Any suggestions?

You just need to make sure that the geospatial libraries are available in your system. This should help: https://docs.djangoproject.com/en/5.1/ref/contrib/gis/install/geolibs/

From your reply, I can see that you are on MacOS, which is slightly more complicated than linux (e.g. on debian/ubuntu you just need to `sudo apt-get install binutils libproj-dev gdal-bin`)

shmoon-kr:
@bellini666 I revised the PR as you instructed. I have one more question. Can I test whether GeoDjango is available or not with the following code from now? Then the code may get simpler.
```
if strawberry_django.GeometryFilterLookup is not None:
    # GeoDjango is available here
```

bellini666:
This PR got really outdated, so I'm closing it
Feel free to reopen it in case you would like to rebase on top of `main`

PR #715, opened 2025-02-27
Order unordered querysets by primary key

### LINKED ISSUE #714: Models without a default `Meta.ordering` can cause non-deterministic results
## Describe the Bug

GraphQL types for models with no default `Meta.ordering` can result in non-deterministic results, in particular when paginating.

The reason for this is obvious, but what's not clear is _how_ / _where_ this should be handled.

## System Information

 - Operating system: MacOS
 - Strawberry version (if applicable): 0.260.2

## Additional Context

I understand that one approach to fixing this would be to just add `Meta.ordering = [...]` to the relevant models, but it seems legitimate that outside of a GraphQL context some models shouldn't need to have a default ordering.

So instead, it seems like Strawberry-Django could handle this, by either:

1. Allowing a default `order_by` to be provided in a type's definition, and require it be provided for models with no `Meta.ordering`
2. Adding a default `.order_by("pk")` for any models without `Meta.ordering` (when appropriate)

I've opened a draft PR for (2) above, but I'm still unsure about whether this being done in the right spot for two reasons:

1. Ordering is a database operation that has some cost associated, so we only want to add ordering when we absolutely have to.
2. We don't want to override any ordering that may be provided by a type's `get_queryset`, field extensions, etc.

PR: https://github.com/strawberry-graphql/strawberry-django/pull/715

Regardless, given the risks associated with non-deterministic results, it feels like we should do _something_ here to stop it from happening.

---

For reference, our current workaround for this is adding the following to our types for models with no `Meta.ordering` (but this seems repetitive / pointless):

```python
@classmethod
def get_queryset(cls, queryset, info, **kwargs):
    return queryset.order_by("pk")
```

SupImDos:
## Description

As per #714, types for models with no default `Meta.ordering` can result in non-deterministic results. This PR adds a `.order_by("pk")` to unordered QuerySets for any list, connection, paginated or optional type. It also adds some extra fields to the `tests/projects/schema.py` GraphQL schema to enable the testing of more combinations of unordered models / types / fields.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

Fix #714

## Checklist

## Summary by Sourcery

Orders unordered QuerySets by primary key to ensure deterministic results for list, connection, paginated or optional types.

Bug Fixes:
- Fixes a bug where models with no default `Meta.ordering` can result in non-deterministic results.

Tests:
- Adds tests to verify that queries on list fields and nested list fields with no ordering have an `ORDER BY` clause.

## Summary by Sourcery

Ensures deterministic ordering of query results by ordering unordered QuerySets by their primary key. This change affects list, connection, paginated, and optional types, guaranteeing consistent results when no explicit ordering is defined in the model's Meta class.

Bug Fixes:
- Fixes non-deterministic results for models without default ordering in Meta, ensuring consistent query results.

Enhancements:
- Adds primary key ordering to unordered QuerySets for list, connection, paginated, or optional types.

Tests:
- Adds tests to verify that queries on list fields and nested list fields with no ordering have an `ORDER BY` clause.

### DIFF
--- a/strawberry_django/pagination.py
+++ b/strawberry_django/pagination.py
@@ -342,6 +342,14 @@ class StrawberryDjangoPagination(StrawberryDjangoFieldBase):
     ) -> _QS:
         queryset = super().get_queryset(queryset, info, **kwargs)
 
+        # If the queryset is not ordered, and this field is either going to return
+        # multiple records, or call `.first()`, then order by the primary key to ensure
+        # deterministic results.
+        if not queryset.ordered and (
+            self.is_list or self.is_paginated or self.is_connection or self.is_optional
+        ):
+            queryset = queryset.order_by("pk")
+
         # This is counter intuitive, but in case we are returning a `Paginated`
         # result, we want to set the original queryset _as is_ as it will apply
         # the pagination later on when resolving its `.results` field.
@@ -356,9 +364,6 @@ class StrawberryDjangoPagination(StrawberryDjangoFieldBase):
             self.is_list or self.is_paginated or self.is_connection
         ):
             if self.is_optional:
-                # first() applies order by pk if not ordered
-                if not queryset.ordered:
-                    queryset = queryset.order_by("pk")
                 pagination = OffsetPaginationInput(offset=0, limit=1)
             else:
                 pagination = OffsetPaginationInput(offset=0, limit=MAX_GET_RESULTS)
--- a/tests/projects/schema.py
+++ b/tests/projects/schema.py
@@ -100,8 +100,7 @@ class ProjectFilter:
 @strawberry_django.type(Project, filters=ProjectFilter, pagination=True)
 class ProjectType(relay.Node, Named):
     due_date: strawberry.auto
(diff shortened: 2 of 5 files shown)


bellini666:
Hi @SupImDos ,

I like the solution :)

tip: rebase this on top of `main` as I fixed the current typing issues which broke in this PR

SupImDos:
> Hi @SupImDos ,
> 
> I like the solution :)
> 
> tip: rebase this on top of `main` as I fixed the current typing issues which broke in this PR

Thanks @bellini666! I've rebased and fleshed out the unit tests, but it looks like there are still some typing errors for some reason :shrug:

bellini666:
> > Hi @SupImDos ,
> > I like the solution :)
> > tip: rebase this on top of `main` as I fixed the current typing issues which broke in this PR
> 
> Thanks @bellini666! I've rebased and fleshed out the unit tests, but it looks like there are still some typing errors for some reason 🤷

It is because of a revert from pyright lol

Will merge this and fix the typing issues in a following commit

bellini666 (approved):

PR #719, opened 2025-03-17
Fix unnecessary window pagination being used

diesieben07:
This was broken by me in #675, I apologize.

## Description

When prefetching, `strawberry-django` currently always uses window pagination, even though it is not necessary in most situations. This was broken in #675. 
In theory there are situation where using Window pagination here makes sense, but I am not sure it is worth it to try and detect those, so I am just disabling the automatic pagination from #675 for prefetches.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

## Summary by Sourcery

Bug Fixes:
- Fixes unnecessary window pagination being used when prefetching data.

### DIFF
--- a/strawberry_django/pagination.py
+++ b/strawberry_django/pagination.py
@@ -360,8 +360,10 @@ class StrawberryDjangoPagination(StrawberryDjangoFieldBase):
         # Add implicit pagination if this field is not a list
         # that way when first() / get() is called on the QuerySet it does not cause extra queries
         # and we don't prefetch more than necessary
-        if not pagination and not (
-            self.is_list or self.is_paginated or self.is_connection
+        if (
+            not pagination
+            and not (self.is_list or self.is_paginated or self.is_connection)
+            and not _strawberry_related_field_id
         ):
             if self.is_optional:
                 pagination = OffsetPaginationInput(offset=0, limit=1)
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -3,7 +3,9 @@ from typing import Any, cast
 
 import pytest
 import strawberry
+from django.db import DEFAULT_DB_ALIAS, connections
 from django.db.models import Prefetch
+from django.test.utils import CaptureQueriesContext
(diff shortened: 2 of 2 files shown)


strawberry_django/pagination.py
```diff
@@ -360,8 +360,10 @@ def get_queryset(
         # Add implicit pagination if this field is not a list
         # that way when first() / get() is called on the QuerySet it does not cause extra queries
         # and we don't prefetch more than necessary
-        if not pagination and not (
-            self.is_list or self.is_paginated or self.is_connection
+        if (
+            not pagination
+            and not (self.is_list or self.is_paginated or self.is_connection)
+            and not _strawberry_related_field_id
```
aprams, line 366: When is `_strawberry_related_field_id` actually set? Is this only during nested field prefetches? 
If I read this correctly, we would now not have pagination whenever this is set, which would only happen when prefetching nested fields. Is that correct? If yes, then this makes sense to me.

diesieben07: Yes, `_strawberry_related_field_id` is set by the optimizer when it gets the queryset for prefetching a foreign key.

For example, if you're querying a list of milestones and for each one you want to get the project then the optimizer will turn that into `select_related` or `prefetch_related`. `prefetch_related` is used if the Project GraphQL type has a custom `get_queryset` for example.
When determining the queryset to use for prefetching the projects in this case, the `_strawberry_related_field_id` would then be `milestones` (`milestones` is the field pointing back to the milestones list from the project). 

The use-case for this is mainly for nested pagination. For example say a milestone could have _multiple_ projects attached to it. The user would query all milestones and then for each milestone get the first 10 projects. To prefetch all those projects in one query, the optimizer uses SQL window functions to get the first 10 projects for every milestone.

The "automatic pagination" that I added in #675 is unrelated to this and only does bad things when used here.

aprams (approved):

bellini666 (approved):
Ty!! :)

PR #720, opened 2025-03-18
Improve handling of polymorphism in the optimizer

diesieben07:
This improves handling of polymorphic types in the optimizer, when using Django Polymorphic, InheritanceManager from django-model-utils or a custom solution.

## Description

Following improvements are made:
- For Django-Polymorphic models the optimizer automatically adds the `polymorphic_ctype` field to `only`. This field is used by django-polymorphic and would otherwise cause extra queries.
- A bug is fixed where the optimizer did not honor the `prefix` when adding the primary key to `only`. This manifested when using Django Polymorphic.
- Support for django-model-utils' `InheritanceManager` is added. When used, the optimizer will use it to support polymorphic models, just like with django-polymorphic.
- The logic for extracting possible subtypes (interface implementations and union members) was extracted into a separate function. Previously this logic was only used at the top level and not for optimizing connections or OffsetPaginated. This resulted in the optimizer missing fields from subtypes when using connections or OffsetPaginated.
- The logic for extracting possible subtypes has been made aware of Django Polymorphic and InheritanceManager. In case they are used it will not just return _supertypes_ of the currently optimized function but also subtypes (speaking in the Model world, not in the GraphQL world). This makes it so the optimizer can now see fields in subtypes of Django Polymorphic models and optimize them.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [X] New feature
- [X] Documentation

## Checklist

- [X] I have updated the documentation accordingly.

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -61,9 +61,8 @@ from .utils.inspect import (
     get_model_fields,
     get_possible_concrete_types,
     get_possible_type_definitions,
-    is_polymorphic_model,
     is_inheritance_manager_or_qs,
-    get_inheritance_prefix,
+    is_polymorphic_model,
 )
 from .utils.typing import (
     AnnotateCallable,
@@ -841,7 +840,11 @@ def _get_hints_from_django_relation(
         remote_model, schema, f_type
     ):
         django_definition = get_django_definition(concrete_field_type.origin)
-        if django_definition and django_definition.model != remote_model and issubclass(django_definition.model, remote_model):
+        if (
+            django_definition
+            and django_definition.model != remote_model
+            and issubclass(django_definition.model, remote_model)
+        ):
             subclasses.append(django_definition.model)
         concrete_store = _get_model_hints(
             remote_model,
@@ -1069,8 +1072,13 @@ def _get_model_hints(
                     prefix=f"{prefix}{dj_definition.model._meta.app_label}__{dj_definition.model._meta.model_name}___",
                 )
             if is_inheritance_manager_or_qs(model._default_manager) and (
-                prefix := get_inheritance_prefix(dj_definition.model, model)
+                path_to_parent := dj_definition.model._meta.get_path_to_parent(model)
             ):
+                prefix = LOOKUP_SEP.join(
+                    p.join_field.remote_field.get_accessor_name()
+                    for p in path_to_parent
+                )
+                prefix += LOOKUP_SEP
                 return _get_model_hints(
                     dj_definition.model,
                     schema,
@@ -1407,7 +1415,11 @@ def optimize(
     ):
         if inheritance_qs:
             django_definition = get_django_definition(inner_object_definition.origin)
-            if django_definition and django_definition.model != qs.model and issubclass(django_definition.model, qs.model):
+            if (
+                django_definition
+                and django_definition.model != qs.model
+                and issubclass(django_definition.model, qs.model)
+            ):
                 subclasses.append(django_definition.model)
         parent_type = _get_gql_definition(schema, inner_object_definition)
         new_store = _get_model_hints(
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -1081,11 +1081,12 @@ def _get_model_hints(
                     prefix=f"{prefix}{dj_definition.model._meta.app_label}__{dj_definition.model._meta.model_name}___",
                 )
             if is_inheritance_manager(model._default_manager) and (
-                path_to_parent := dj_definition.model._meta.get_path_to_parent(model)
+                path_from_parent := dj_definition.model._meta.get_path_from_parent(
+                    model
+                )
             ):
                 prefix = LOOKUP_SEP.join(
-                    p.join_field.remote_field.get_accessor_name()
-                    for p in reversed(path_to_parent)
+                    p.join_field.get_accessor_name() for p in path_from_parent
                 )
                 prefix += LOOKUP_SEP
                 return _get_model_hints(
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -818,7 +818,9 @@ def _get_hints_from_django_relation(
     remote_model = remote_field.model
     field_store = None
     f_type = f_types[0]
-    for concrete_field_type in get_possible_concrete_types(remote_model, schema, f_type):
+    for concrete_field_type in get_possible_concrete_types(
+        remote_model, schema, f_type
+    ):
         concrete_store = _get_model_hints(
             remote_model,
             schema,
@@ -830,7 +832,9 @@ def _get_hints_from_django_relation(
             level=level + 1,
         )
         if concrete_store is not None:
-            field_store = concrete_store if field_store is None else field_store | concrete_store
+            field_store = (
+                concrete_store if field_store is None else field_store | concrete_store
+            )
     if field_store is None:
         return store
 
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -1017,6 +1017,7 @@ def _get_model_hints(
     prefix: str = "",
     cache: dict[type[models.Model], list[tuple[int, OptimizerStore]]] | None = None,
     level: int = 0,
+    subclass_collection: set[type[models.Model]] | None = None,
 ) -> OptimizerStore | None:
     cache = cache or {}
 
@@ -1032,6 +1033,7 @@ def _get_model_hints(
             prefix=prefix,
             cache=cache,
             level=level,
+            subclass_collection=subclass_collection,
         )
 
     # In case this is a Paginated field, the selected fields are inside results selection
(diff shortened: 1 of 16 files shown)


strawberry_django/optimizer.py
```diff
@@ -1234,39 +1265,78 @@ def _get_model_hints_from_paginated(
...
+                cache=cache,
+                level=level,
+            )
+            if concrete_store is not None:
+                store = concrete_store if store is None else store | concrete_store
 
     return store
 
 
+def _get_possible_concrete_types(
```
bellini666, line 1308: **todo:** this should go to `utils/inspect.py`

strawberry_django/optimizer.py
```diff
@@ -1234,39 +1265,78 @@ def _get_model_hints_from_paginated(
...
+    model: type[models.Model],
+    schema: Schema,
+    strawberry_type: StrawberryObjectDefinition | StrawberryType,
+) -> Iterable[StrawberryObjectDefinition]:
+    for object_definition in get_possible_type_definitions(strawberry_type):
+        if object_definition.is_interface:
```
bellini666, line 1314: **nitpick:** we can reduce the indentation here by doing:

```python
if not object_definition.is_interface:
    yield object_definition
    continue

interface_definitions = ...
...
```

bellini666 (changes requested):
LGTM in general 😊

Just 2 small asks and we can merge/release this

diesieben07:
I have converted this PR to a draft, because I just noticed a flaw with the approach I have taken. It does not work for nested relations (e.g. "get all projects (in a polymorphic way) for a company").
I will look into fixing that first.

Apologies for not catching that earlier.

diesieben07:
I have fixed the problem with relations not being optimized and added even more tests.

diesieben07:
As discussed on Discord, I have expanded the scope of this pull request to also include support for `InheritanceManager`.
In addition, I have added a section in the documentation that covers the optimizer's polymorphic abilities.

stygmate:
@diesieben07 Also Thank you :) You are resolving a lot of headaches with this PR 👍

stygmate:
@diesieben07 ultimately, have you tested the optimzer with a polymorphic relation inside a polymorphic subtype (with inheritance manager) ? (i have the case 😅 )

diesieben07:
@stygmate I'm looking into it!

diesieben07:
Okay, I am not sure how to make this work. Here are the models that I used for testing: https://gist.github.com/diesieben07/7360ebf934d174abcf865c23856bef97

Assume we want to get all notifications (polymorphic) and then for every `CommunicationNotification` we also want to get its corresponding `Communication` object (again, polymorphic).

The top level "Notification" query will use `select_subclasses()` to get a polymorphic queryset.
To now have `CommunicationNotification#communication` be polymorphic, we can't use `select_related('communicationnotification__communication')`, because that's not polymorphic. So we use `prefetch_related` instead and make sure to use an `InheritanceQuerySet`. But now we have both `select_related('communicationnotifaction')` (from `select_subclasses`) _and_ `prefetch_related('communicationnotification__communication')` and this doesn't work, because you cannot prefetch_related into an object that was fetched by select_related.

@stygmate Can you let me know how you would write this query manually, without the optimizer?

stygmate:
@diesieben07 I will check all of that tomorrow.

diesieben07:
@stygmate See https://github.com/jazzband/django-model-utils/pull/639 :)

stygmate:
@diesieben07 

I'm not sure we're talking about the same thing.
I wrote a project (without Strawberry) to illustrate my case.
It is available here: https://github.com/stygmate/test_django_inheritance.

Create the virtual environment with Poetry, then run migrations, then the management command `populate_db`, and finally execute my test using another management command called `test_query`.

What I would like is for the optimizer to add this:
https://github.com/stygmate/test_django_inheritance/blob/763d58d081e2923f317883672ec923ce641b8bf8/myapp/management/commands/test_query.py#L56

This is a different issue, but in this case, the foreign key returns the base model, whereas it should return a subtype so that strawberry-django can select the correct GraphQL type.

However, with this approach, everything fits into a single SQL query. The query is displayed on the screen during the execution of the `test_query` command.

diesieben07:
@stygmate On first glance the first problem is that you're calling select_related _and_ select_subclasses, which overwrite each other unless you do extra work.
I'll take a closer look this evening

diesieben07:
@stygmate I looked into it further, and as it stands that will never work with `InheritanceManager` and there's not much the optimizer can do about that.
The reason is that the logic for which subclass to return is implemented entirely within the iteration of the QuerySet. Without `InheritanceIterable` (which `InheritanceQuerySet` returns) there are no subclasses. All `select_subclasses` does is translate the models you give it into `select_related`. But since there's no queryset involved for the return value of a ForeignKey, you won't get subclasses there.
It can be made work with `prefetch_related` however, because that supports custom querysets. But `prefetch_related` has problems with `InheritanceQuerySet` as well - which I have fixed in the PR linked above.

stygmate:
@diesieben07 I'll try using your model-utils version combined with this PR and run some tests.
Thanks for your time and effort! 🙂

strawberry_django/utils/inspect.py
```diff
@@ -143,6 +152,83 @@ def get_possible_type_definitions(
...
+    try:
+        from polymorphic.models import PolymorphicModel
+    except ImportError:
+        return False
+    return issubclass(v, PolymorphicModel)
+
+
+def is_inheritance_manager(
+    v: Any,
+) -> bool:
```
bellini666, line 165: **question:** why is this not a `TypeGuard` / `TypeIs` as well?

strawberry_django/utils/inspect.py
```diff
@@ -143,6 +152,83 @@ def get_possible_type_definitions(
...
+    try:
+        from model_utils.managers import InheritanceManagerMixin
+    except ImportError:
+        return False
+    return isinstance(v, InheritanceManagerMixin)
+
+
+def is_inheritance_qs(
+    v: Any,
+) -> TypeIs[InheritanceQuerySetMixin]:
```
bellini666, line 175: **question:** This is using `TypeIs`, as opposed to `TypeGuard` above. Maybe try to stick with one? Or is there a reason for the difference?

(I have to say, I'm still not 100% certain about the differences between both)

strawberry_django/utils/inspect.py
```diff
@@ -143,6 +152,83 @@ def get_possible_type_definitions(
             yield t.__strawberry_definition__
 
 
+def is_polymorphic_model(v: type) -> TypeGuard[type[PolymorphicModel]]:
+    try:
+        from polymorphic.models import PolymorphicModel
+    except ImportError:
+        return False
```
bellini666, line 159: **thought:** those are used a lot in the optimizer, and trying to import it every time in the case where it doesn't exist has a huge performance impact:

Screenshot 2025-03-30 at 13 19 24

Just checking if something is not `None` is almost 900 times faster.

So maybe try to import it at the module level and define it as `None` on `ImportError`? Then just return `False` if it is `None`.

Or even define this function in a try/except ImportError, where the except case would be a function that just returns `False`?

obs. the same applies for the functions below

bellini666:
@diesieben07 thank you so much for this ❤️

Really well done and properly tested.

Just some small things to adjust and it should be good to merge.

Btw, regarding your discussion with @stygmate , is there anything else that you want to implement here? Or is it good to merge after the adjustments I'm asking?

diesieben07:
@bellini666 Thank you! Not sure if I missed it, but I can't see any requested changes from you.

As for the discussion, in my opinion the still present issues with relations in polymorphic models are ones that the optimizer can't fix and that need to be fixed in Django-Model-Utils (see the PR I linked above) and Django-Polymorphic.

bellini666 (changes requested):
Oh, my bad! Forgot to apply the comments 😅

strawberry_django/utils/inspect.py
diesieben07: No reason except that the code that calls it doesn't need type narrowing. But I've added it!

strawberry_django/utils/inspect.py
diesieben07: You're right. `TypeIs` is the more modern one and the more intuitive one. `TypeIs` is necessary so that when `is_inheritance_qs` is called on a `QuerySet[T]` the result becomes _both_ `QuerySet[T]` _and` `InheritanceQuerySetMixin`, so we can still use it as a normal `QuerySet`. `TypeGuard` doesn't do that.
I had to switch over `is_inheritance_qs` to make Pyright happy for this reason and just forgot to change the other one over. I've switched it so both are using `TypeIs`.

strawberry_django/utils/inspect.py
diesieben07: Right, I meant to do that still and forgot.
It is a bit tricky for `PolymorphicModel`, because importing it requires the Django apps to be ready. I have settled for `import polymorphic` to check if the library is installed and then using the `polymorphic_model_marker` attribute for the type check instead.

bellini666 (approved):
Ty so much for this! ❤️

bellini666 (approved):
Ty so much for this! ❤️

PR #730, opened 2025-04-09
Add support for "true" cursor based pagination in connections

diesieben07:
This pull request adds a new connection class `DjangoCursorConnection` which supports efficient cursor-based pagination through any Django `QuerySet` without relying on offset-slicing.

## Description

`ListConnection` uses slicing to achieve pagination. This works for the general case, but can be inefficient for large datasets, because large page numbers result in a large `OFFSET` in SQL. An alternative to limit/offset pagination is cursor based pagination, which replaces `OFFSET` by range queries such as `Q(due_date__gt=...)`. `DjangoCursorConnection` implements this approach.

## How it works
`DjangoCursorConnection` inspects the `QuerySet` and extracts its ordering parameters. It then uses those parameters to construct the cursors. For example, for `order_by("due_date", "pk")` the ordering parameters would be `due_date` and `pk`. If the ordering parameter is an expression or not a direct field on the model (e.g. `order_by(Upper("name"))` or `order_by("project__name")`, a new annotation will be added to the queryset, mirroring the ordering expression, so that the value can be extracted later.
The extracted values are then encoded into a cursor.
When paginating, the cursor is deconstructed into its parts again and those parts are then used to build a pagination filter.
For example, when ordering by `"due_date", "pk"`, the cursor might contain the parts `"2025-03-01"` and `,"3"`. If that cursor is passed for `after`, the following filter would be constructed:
`Q(due_date__gt="2025-03-01") | (Q(due_date="2025-03-01") & Q(pk__gt="3"))`

Serializing and deserializing the field values to strings is dedicated to the model field implementation, ensuring maximal compatibility. 

## Other
- During the implementation I discovered a bug in `get_queryset_config`. It is used in the code as if it did a "get or create" operation, setting the config on the QuerySet if not already present. However it did not actually do so.
This is likely the cause for why [this hack](https://github.com/strawberry-graphql/strawberry-django/blob/8898175e53c3bc49059ce3f2a7f16084022ea9e5/strawberry_django/optimizer.py#L1457-L1467) was implemented.

- While writing tests I have noticed that `strawberry_django.field(disable_optimization=True)` had no effect when used on a top level field. I have fixed this.

- Currently, the code lives in a separate `relay_cursor.py` file. I have chosen to do this _for now_ to make the diff of this PR easier to parse. However I think the relay code should be refactored so that `relay.py` is removed and we have `relay/__init__.py` instead. Then the code can be split up into multiple files and still offer the same imports. What do you think?

I have fixed `get_queryset_config`, added specific tests for it and removed the (now unnecessary) hack in `is_optimized_by_prefetching`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [X] New feature
- [X] Documentation

## Checklist

- [X] I have updated the documentation accordingly.


diesieben07:
I don't know why the Typing test fails. `pyright` reports no problems for me locally and the reported file here is a file that I haven't touched.

strawberry_django/optimizer.py
```diff
@@ -1451,20 +1470,17 @@ def optimize(
...
 
 def mark_optimized_by_prefetching(qs: QuerySet[_M]) -> QuerySet[_M]:
-    # This is a bit of a hack, but there is no easy way to mark a related manager
-    # as optimized at this phase, so we just add a mark to the queryset that
-    # we can check leater on using is_optimized_by_prefetching
-    return qs.annotate(**{
-        NESTED_PREFETCH_MARK: models.Value(True),
-    })
+    get_queryset_config(qs).optimized_by_prefetching = True
+    return qs
```
bellini666, line 1479: **question:** is this change safe? If I remember correctly, the reason we would do this was to mark a prefetch query as optimized, because that config there could be lost

Is that not an issue anymore with the changes here? I sincerely don't remember how it would be lost and even if we have a test that would break without this (hopefully we do)

strawberry_django/relay_cursor.py
```diff
@@ -0,0 +1,461 @@
...
+def _get_order_by(qs: QuerySet) -> list[OrderBy]:
+    return [
+        expr
+        for expr, _ in qs.query.get_compiler(
+            using=qs._db or DEFAULT_DB_ALIAS  # type: ignore
+        ).get_order_by()
+    ]
+
+
+class OrderingDescriptor(NamedTuple):
```
bellini666, line 39: **nitpick:** instead of a `NamedTuple`, this can be a `dataclass`

strawberry_django/relay_cursor.py
```diff
@@ -0,0 +1,461 @@
...
+        if current is None:
+            current = cmp
+        else:
+            eq = descriptor.get_eq(value_expr)
+            current = cmp | (eq & current) if cmp is not None else eq & current
+    return current if current is not None else Q()
+
+
+class AttrHelper:
+    pass
```
bellini666, line 161: **question:** Why do we have this?

strawberry_django/relay_cursor.py
```diff
@@ -0,0 +1,461 @@
...
+        )
+    elif slice_ is not None:
+        qs = qs[slice_]
+
+    get_queryset_config(qs).ordering_descriptors = ordering_descriptors
+
+    return qs, ordering_descriptors
+
+
+class OrderedCollectionCursor(NamedTuple):
```
bellini666, line 280: **nitpick:** Ditto

strawberry_django/relay_cursor.py
```diff
@@ -0,0 +1,461 @@
...
+    field_values: list[Any]
+
+    @staticmethod
```
bellini666, line 283: **suggestion:** if you make this a classmethod, you can `return cls(field_values=values)` below

strawberry_django/relay_cursor.py
```diff
@@ -0,0 +1,461 @@
...
+        values = [
+            _extract_expression_value(
+                model, descriptor.order_by.expression, descriptor.attname
+            )
+            for descriptor in descriptors
+        ]
+        return OrderedCollectionCursor(field_values=values)
+
+    @staticmethod
+    def from_cursor(
```
bellini666, line 296: **suggestion:** ditto

bellini666 (changes requested):
Good job here, really appreciate your contributions ❤️

Left a couple of comments/suggestions

strawberry_django/optimizer.py
diesieben07: From my understanding it is safe, yes. If I change `is_optimized_by_prefetching` to just `return False` then various tests fail, most importantly this one: https://github.com/strawberry-graphql/strawberry-django/blob/048c57d83e8c3ab34a0d05cf6318a2d0c1ab9832/tests/relay/test_nested_pagination.py#L11 but also various others that work with nested connections.

My understanding of why this code was written as it was before my change is as follows:
1. `get_queryset_config` was broken in that it never set the configuration on the QuerySet. This wasn't caught because
    1. There were no tests for it.
    2. If it fails, that is only catastrophic for `is_optimized_by_prefetching`. The other things in the `QuerySet` config can cause the optimizer to optimize a `QuerySet` twice or they can cause a type's `get_queryset` to be called twice. Both of these are usually idempotent so don't cause any issues except for wasted CPU cycles.
2. The above caused the `get_queryset_config` bug to never surface _until_ `is_optimized_by_prefetching` was introduced and at that point the config disappearing was wrongly attributed to being related to `prefetch_related` when in reality `get_queryset_config` _never worked_.

strawberry_django/relay_cursor.py
diesieben07: Sure! I think I used a `NamedTuple` originally for performance reasons, but if you think that's moot than I am fine changing it to a dataclass.

strawberry_django/relay_cursor.py
diesieben07: We need to serialize the values of all `order_by` columns into the cursor, which is just a string. The fields could be of any type so we don't want to just `str()` them. Django model fields already have built-in serialization to and from strings, which Django uses for serializing and deserializing model instances (mainly for fixtures).

The API however is unfortunately somewhat clunky. Instead of `Field.value_to_string(value)` we get `Field.value_to_string(obj)` with `obj` being the model to look up the field on. So you could do something like
```python
field_value_str = some_model_field.value_to_string(my_model_instance)
```

This works for fields which are present on the model, but users can order by arbitrary annotations:
```py
qs.alias(name_up=Upper('name')).order_by('name_up')
```
Now we need to somehow turn the value of that annotation into a string via `value_to_string` - but there's no actual field for it. Hence we construct this `AttrHelper` and set the attribute that the field is looking for on it. This happens here:
https://github.com/diesieben07/strawberry-graphql-django/blob/39dcdd6e09aaf20c831707d6e01e4bfdacb24028/strawberry_django/relay_cursor.py#L179-L184

We can't use a plain `object()`, because you can't set arbitrary attributes on them.

Django's own `ArrayField` for PostgreSQL uses a similar technique:
https://github.com/django/django/blob/ab1b9cc1b38e8735979c817f42e8f54a5795c4f7/django/contrib/postgres/fields/array.py#L166-L167

strawberry_django/relay_cursor.py
diesieben07: Right. I changed this because at one point I had this class as a generic class and didn't manage to figure out how to make `Self` work with generic arguments.
But I can change it back now.

diesieben07:
See https://github.com/strawberry-graphql/strawberry-django/pull/736#issuecomment-2833415584 for the typing issue.

@bellini666 As I said on Discord I have updated this for the changes in Strawberry Core, feel free to review this again or merge it if you think it's good to go!

bellini666 (changes requested):
I have one last ask, and then it is an automerge.

Can we have all of this under `strawberry_django.relay`?

We can do that by creating a `relay` directory, rename the current `relay.py` to `utils.py`, move the `ListConnectionWithTotalCount` in it to a `connections.py` module, and put everything you have on `relay_cursor` inside it. Then we export the important stuff on `relay/__init__.py`. Wdyt?

We would end up with something like

```
strawberry_django/
  relay/
    __init__.py
    connections.py
    utils.py
```

diesieben07:
> Can we have all of this under strawberry_django.relay?

Yes, 100%! That was my plan all along, I did it this way for now to make the diff easier to parse and limit the changes to what I am actually changing in this PR. I'll refactor it as you suggested.

diesieben07:
@bellini666  I've applied your suggestion, DjangoCursorConnection can now be imported from `strawberry_django.relay`.
I've kept `cursor_connection.py` and `list_connection.py` as separate files.
I had to move the actual relay implementation to `relay_impl` and introduce `relay` as an "interface package" to resolve complex import cycles that arose from just having the implementation in `relay`. See this video (with timestamp) for info: https://www.youtube.com/watch?v=UnKa_t-M_kM&t=329s

bellini666 (approved):
Ty very much! ❤️

PR #736, opened 2025-04-27
Fix duplicate LOOKUP_SEP being used when field hints are used with polymorphic queries

### LINKED ISSUE #733: Polymorphic models with fields with optimizer hints aren't resolved correctly
## Describe the Bug

When using an annotated field with any optimizer hints, the optimizer attempts to load the fields with a prefix `app_label__ModelName___` ([here](https://github.com/strawberry-graphql/strawberry-django/blob/5d217529ce407d72c94bbfa18bd561d70f627e54/strawberry_django/optimizer.py#L1070)). This works perfectly for related field resolution, however it does not work for the hints. They are resolved like this:

`app_label__ModelName_____someField` (note that there are 5 underscores)

This causes django to try to prefetch an empty field and crashes.

## System Information

 - Operating system: Any
 - Strawberry django version: 0.58.0

## Additional Context

I've created a small reproduction [here](https://github.com/ManiacMaxo/polymorphic-strawberry/tree/main) with the file in question: [graphql.py](https://github.com/ManiacMaxo/polymorphic-strawberry/blob/main/core/graphql.py)

I'll continue trying to find the cause and might open a PR with a fix. Not sure if I'll be able to though...

diesieben07:
When an optimizer `only` hint was used in a polymorphic model, the optimizer used an incorrect prefix.

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #733

## Checklist

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -965,7 +965,8 @@ def _get_hints_from_django_field(
     if (model_field := get_model_field(model, model_fieldname)) is None:
         return None
 
-    path = f"{prefix}{model_fieldname}"
+    lookup_prefix = prefix + LOOKUP_SEP if prefix else ""
+    path = f"{lookup_prefix}{model_fieldname}"
 
     if isinstance(model_field, (models.ForeignKey, OneToOneRel)):
         store = _get_hints_from_django_foreign_key(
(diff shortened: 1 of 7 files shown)


diesieben07:
I don't know why the typing check failed. Pyright reports no errors for me locally and it does not seem related to this change at all.

bellini666 (approved):

PR #738, opened 2025-04-30
Fix a minor typing issue

diesieben07:
This fixes a minor typing issue. It was previous just ignored with type: ignore, but a Pyright update apparently changed how those are applied, so the type issue is no longer ignored.
I have fixed the typing issue itself, so the type: ignore is no longer necessary.

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

## Summary by Sourcery

Fix a typing issue in the Django optimizer by removing type ignore and using a type cast

Bug Fixes:
- Resolved a typing issue with deferred loading in the Django optimizer

Enhancements:
- Replaced type: ignore with a safe type cast
- Updated deferred loading initialization to use a set instead of a list

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -333,8 +333,9 @@ class OptimizerStore:
 
         # Abort only optimization if one prefetch related was made for everything
         for ao in abort_only:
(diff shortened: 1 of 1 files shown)


bellini666 (approved):
Ty! ❤️

PR #742, opened 2025-05-05
Optimizer: Pass accurate "info" parameter to PrefetchCallable and AnnotateCallable

### LINKED ISSUE #697: Nested prefetch_related resolver optimizations are seemingly ignored
## Describe the Bug

The schema that I'm working with often has nested field types with `prefetch_related` optimizer hints. I noticed that if one field resolver has a `prefetch_related` defined, nested field resolvers would not get optimized, which resulted in N+1s.

Example schema:

```python
@strawberry_django.type(Project, filters=ProjectFilter, pagination=True)
class ProjectType(relay.Node, Named):
    @strawberry_django.field(
        pagination=True,
        prefetch_related=["milestones"],
    )
    def milestones(self) -> list["MilestoneType"]:
        # This resolver is contrived for this example. In actuality more would
        # be happening other than returning .all(), such as extra filtering, etc.
        return self.milestones.all()

@strawberry_django.type(
    Milestone, filters=MilestoneFilter, order=MilestoneOrder, pagination=True
)
class MilestoneType(relay.Node, Named):
    issues: list["IssueType"] = strawberry_django.field(
        filters=IssueFilter,
        order=IssueOrder,
        pagination=True,
    )

@strawberry_django.type(Issue)
class IssueType(relay.Node, Named):
    priority: strawberry.auto
```

Test query:
```graphql
query TestQuery ($node_id: GlobalID!) {
  project (id: $node_id) {
    milestones {
      issues {
        priority
      }
    }
  }
}
```

Resulting DB queries from test:

```sql
AssertionError: 8 queries executed, 3 expected
Captured queries were:
1. SELECT "projects_project"."id" FROM "projects_project" WHERE "projects_project"."id" = 1 ORDER BY "projects_project"."id" ASC LIMIT 1
2. SELECT "projects_milestone"."name", "projects_milestone"."id", "projects_milestone"."due_date", "projects_milestone"."project_id" FROM "projects_milestone" WHERE "projects_milestone"."project_id" IN (1)
3. SELECT "projects_issue"."id", "projects_issue"."priority" FROM "projects_issue" WHERE "projects_issue"."milestone_id" = 1 ORDER BY "projects_issue"."id" ASC
4. SELECT "projects_issue"."id", "projects_issue"."milestone_id" FROM "projects_issue" WHERE "projects_issue"."id" = 1 LIMIT 21
5. SELECT "projects_issue"."id", "projects_issue"."milestone_id" FROM "projects_issue" WHERE "projects_issue"."id" = 2 LIMIT 21
6. SELECT "projects_issue"."id", "projects_issue"."priority" FROM "projects_issue" WHERE "projects_issue"."milestone_id" = 2 ORDER BY "projects_issue"."id" ASC
7. SELECT "projects_issue"."id", "projects_issue"."milestone_id" FROM "projects_issue" WHERE "projects_issue"."id" = 3 LIMIT 21
8. SELECT "projects_issue"."id", "projects_issue"."milestone_id" FROM "projects_issue" WHERE "projects_issue"."id" = 4 LIMIT 21
```

`milestones` is prefetched appropriately, but then the optimizer doesn't continue with optimizing the `issues` queries. Changing the `milestones` field resolver back to `milestones: list["MilestoneType"] = strawberry_django.field(pagination=True)` resulted in these expected db queries with issues prefetched:

```sql
1. SELECT "projects_project"."id" FROM "projects_project" WHERE "projects_project"."id" = 1 ORDER BY "projects_project"."id" ASC LIMIT 1
2. SELECT "projects_milestone"."id", "projects_milestone"."project_id", 1 AS "_strawberry_nested_prefetch_optimized" FROM "projects_milestone" WHERE "projects_milestone"."project_id" IN (1)
3. SELECT "projects_issue"."id", "projects_issue"."priority", "projects_issue"."milestone_id", 1 AS "_strawberry_nested_prefetch_optimized" FROM "projects_issue" WHERE "projects_issue"."milestone_id" IN (1, 2) ORDER BY "projects_issue"."id" ASC
```

## Additional Context

I'm unsure if this is an actual bug or expected behavior based on the schema setup. I know there were two different changes that skip optimizations early:

1. https://github.com/strawberry-graphql/strawberry-django/pull/583/files - skipping optimizations if a field has a base resolver
2. https://github.com/strawberry-graphql/strawberry-django/pull/582/files - skipping optimizations if `prefetch_related` is in the field's store   
   * This check targets prefetches with filters/annotations, but I _think_ it's also having the side effect of skipping nested optimizations if prefetches are querysets without filters/annotations

So I believe nested fields aren't reaching `_get_hints_from_django_relation` / `_get_model_hints` due to those early returns. I tried removing these early return blocks locally out of curiosity, and that reduced the number of issue queries back to one (though I don't believe that's the right fix, it was purely for testing).

If this is an issue due to schema setup, is there general guidance on how to set up resolvers to ensure that all `prefetch_related` optimizer hints are successfully handled?

Thanks in advance!

## System Information

 - Operating system: macOS
 - Strawberry version (if applicable): 0.55.0

diesieben07:
The optimizer allows dynamic annotations or prefetches by passing a callable. Currently this callable does not always receive an accurate `GraphQLResolveInfo`.

## Description

The optimizer is not consistent in when it resolves dynamic prefetch_related or annotate callables. Sometimes they are resolved immediately, with the correctly scoped info object, but sometimes they are only resolved late, when the optimizer store is applied. In the latter case, the top level Info object is passed, which is incorrect.

This change updates the opimizer to always resolve the callables immediately when a field is looked at. This allows them to receive the correctly scoped info object. This is achieved by a new method `with_resolved_callables` in `OptimizerStore`, which resolves any callables.

To write correct tests for this feature, I required access to the SQL queries issued by the optimizer. This is done by `CaptureQueriesContext`, however this feature did not work in an async context. I have fixed this, so `assert_num_queries` now also works in async mode.

Additionally, I found that Strawberry Django will add `filter` and `ordering` arguments to fields that point to a `@model_property`. This is not useful, because the model_property effectively acts as a custom resolver. I have added an additional check to not add these arguments if the field points to a model_property.

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #697 

## Checklist

### DIFF
--- a/strawberry_django/fields/base.py
+++ b/strawberry_django/fields/base.py
@@ -20,6 +20,7 @@ from strawberry.types.field import UNRESOLVED, StrawberryField
 from strawberry.types.union import StrawberryUnion
 from strawberry.utils.inspect import get_specialized_type_var_map
 
+from strawberry_django.descriptors import ModelProperty
 from strawberry_django.resolvers import django_resolver
 from strawberry_django.utils.typing import (
     WithStrawberryDjangoObjectDefinition,
@@ -138,6 +139,13 @@ class StrawberryDjangoFieldBase(StrawberryField):
             django_type.__strawberry_django_definition__.model
             if django_type is not None
             else None
+        )
+
+    @functools.cached_property
+    def has_model_property(self) -> bool:
+        django_definition = self.origin_django_type
+        return django_definition is not None and isinstance(
+            getattr(django_definition.model, self.python_name, None), ModelProperty
         )
 
     @functools.cached_property
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -209,6 +209,23 @@ class OptimizerStore:
             ),
         )
 
+    def with_resolved_callables(self, info: GraphQLResolveInfo):
+        if not any(isinstance(p, Callable) for p in self.prefetch_related):
+            return self
+
+        prefetch_related = []
+        for p in self.prefetch_related:
+            if isinstance(p, Callable):
+                prefetch_related.append(p(info))
+            else:
+                prefetch_related.append(p)
+        return self.__class__(
+            only=self.only,
+            select_related=self.select_related,
+            prefetch_related=prefetch_related,
+            annotate=self.annotate,
+        )
+
     def with_prefix(self, prefix: str, *, info: GraphQLResolveInfo):
         """Create a copy of this store with the given prefix.
 
@@ -668,6 +685,8 @@ def _get_hints_from_field(
     if not (field_store := getattr(field, "store", None)):
         return None
 
+    field_store: OptimizerStore
+
     if len(field_store.annotate) == 1 and _annotate_placeholder in field_store.annotate:
         # This is a special case where we need to update the field name,
         # because when field_store was created on __init__,
@@ -680,7 +699,7 @@ def _get_hints_from_field(
             field.name: field_store.annotate[_annotate_placeholder],
         }
 
-    return field_store.with_prefix(prefix, info=f_info) if prefix else field_store
+    return field_store.with_prefix(prefix, info=f_info) if prefix else field_store.with_resolved_callables(f_info)
 
 
 def _get_hints_from_model_property(
@@ -697,7 +716,7 @@ def _get_hints_from_model_property(
         and model_attr.store
     ):
         attr_store = model_attr.store
-        store = attr_store.with_prefix(prefix, info=f_info) if prefix else attr_store
+        store = attr_store.with_prefix(prefix, info=f_info) if prefix else attr_store.with_resolved_callables(f_info)
     else:
         store = None
 
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -10,6 +10,7 @@ from collections.abc import Callable
 from typing import (
     TYPE_CHECKING,
     Any,
+    Optional,
     TypeVar,
     cast,
 )
@@ -697,10 +698,10 @@ def _get_hints_from_field(
     f_info: GraphQLResolveInfo,
     prefix: str = "",
 ) -> OptimizerStore | None:
-    if not (field_store := getattr(field, "store", None)):
+    if not (
+        field_store := cast("Optional[OptimizerStore]", getattr(field, "store", None))
+    ):
         return None
-
-    field_store: OptimizerStore
 
     if len(field_store.annotate) == 1 and _annotate_placeholder in field_store.annotate:
         # This is a special case where we need to update the field name,
@@ -714,7 +715,11 @@ def _get_hints_from_field(
             field.name: field_store.annotate[_annotate_placeholder],
         }
 
-    return field_store.with_prefix(prefix, info=f_info) if prefix else field_store.with_resolved_callables(f_info)
+    return (
+        field_store.with_prefix(prefix, info=f_info)
+        if prefix
+        else field_store.with_resolved_callables(f_info)
+    )
 
 
 def _get_hints_from_model_property(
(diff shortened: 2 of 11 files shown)


strawberry_django/optimizer.py
```diff
@@ -208,6 +209,35 @@ def with_hints(
...
+            isinstance(p, Callable) for p in self.annotate.values()
+        ):
+            return self
+
+        prefetch_related = []
+        for p in self.prefetch_related:
+            if isinstance(p, Callable):
+                prefetch_related.append(p(info))
+            else:
+                prefetch_related.append(p)
```
bellini666, line 227: **nitpick:** 

```suggestion
        prefetch_related = [
            p(info) if callable(p) else p for p in self.prefetch_related
        ]
```

strawberry_django/optimizer.py
```diff
@@ -208,6 +209,35 @@ def with_hints(
...
+        annotate = {}
+        for label, annotation in self.annotate.items():
+            if isinstance(annotation, Callable):
+                annotate[label] = annotation(info)
+            else:
+                annotate[label] = annotation
```
bellini666, line 233: **nitpick:** 

```suggestion
        annotate = {
            label: annotation(info) if callable(annotation) else annotation
            for label, annotation in self.annotate.items()
        }
```

strawberry_django/optimizer.py
```diff
@@ -208,6 +209,35 @@ def with_hints(
             ),
         )
 
+    def with_resolved_callables(self, info: GraphQLResolveInfo):
+        """Resolve any prefetch/annotate callables using the provided info and return a new store.
+
+        This is used to resolve callables using the correct info object, scoped to their respective fields.
+        """
+        if not any(isinstance(p, Callable) for p in self.prefetch_related) and not any(
+            isinstance(p, Callable) for p in self.annotate.values()
```
bellini666, line 218: **nitpick:** 

```suggestion
        if not callable(p) for p in self.prefetch_related) and not any(
            callable(a) for a in self.annotate.values()
```

strawberry_django/optimizer.py
```diff
@@ -695,7 +729,11 @@ def _get_hints_from_field(
             field.name: field_store.annotate[_annotate_placeholder],
         }
 
-    return field_store.with_prefix(prefix, info=f_info) if prefix else field_store
+    return (
+        field_store.with_prefix(prefix, info=f_info)
```
bellini666, line 733: **question:** doesn't this also needs to call `.with_resolved_callables` after?

strawberry_django/optimizer.py
```diff
@@ -712,7 +750,11 @@ def _get_hints_from_model_property(
         and model_attr.store
     ):
         attr_store = model_attr.store
-        store = attr_store.with_prefix(prefix, info=f_info) if prefix else attr_store
+        store = (
+            attr_store.with_prefix(prefix, info=f_info)
```
bellini666, line 754: **question:** same question

tests/utils.py
```diff
@@ -86,19 +92,40 @@ def deep_tuple_to_list(data: tuple) -> list:
     return return_list
 
 
+class AsyncCaptureQueriesContext:
```
bellini666, line 95: **praise:** OMG, thanks for this ❤️

strawberry_django/fields/base.py
```diff
@@ -140,6 +141,13 @@ def django_model(self) -> type[models.Model] | None:
             else None
         )
 
+    @functools.cached_property
+    def has_model_property(self) -> bool:
```
bellini666, line 145: **suggestion:** how about `is_model_property`?

bellini666 (changes requested):
Mostly nitpicks, but the solution in general looks good! 😊

strawberry_django/optimizer.py
diesieben07: No, because `with_prefix` also resolves them. It has to do so to add the prefix. I'll add this to the docs for `with_prefix` though and add a comment here.

strawberry_django/fields/base.py
diesieben07: I picked "has" over "is", because the field itself and the model property are different. The field _uses_ the model property to resolve its value. But I am not hell-bent on it either way.

strawberry_django/optimizer.py
diesieben07: I always forget about `callable` :D Thank you!

bellini666 (approved):
Ty ❤️

PR #777, opened 2025-07-12
Relay pagination optimizations

Kitefiko:
## Description
Goal is to optimize relay nested and non-nested pagination when paginating backwards (via `last`) without `before`. Current behavior is to fetch entire table for this case.

Additionally added `total_count` optimization for non-nested pagination.

I am not very confident with the solutions for `total_count` in  `strawberry_django/relay/list_connection.py`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry-django/issues/613

## Checklist

## Summary by Sourcery

Improve Relay pagination performance by enabling efficient backward pagination for last-only queries and optimizing totalCount resolution via SQL window functions to avoid full-table scans and extra COUNT queries.

Enhancements:
- Automatically inject before cursors for queries using last without a before argument and apply reversed window pagination to limit QuerySets.
- Annotate QuerySets with a strawberry-specific windowed total count (_strawberry_total_count) to serve totalCount from existing query annotations.
- Extend apply_window_pagination to support reversed ordering for backward pagination scenarios.

Tests:
- Add tests for nested backward pagination and non-nested totalCount optimizations.
- Update expected database query counts in existing pagination and optimizer tests to reflect the new optimizations.

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -37,7 +37,7 @@ from graphql import (
 from graphql.execution.collect_fields import collect_sub_fields
 from graphql.language.ast import OperationType
 from graphql.type.definition import GraphQLResolveInfo, get_named_type
-from strawberry import relay
+from strawberry import UNSET, relay
 from strawberry.extensions import SchemaExtension
 from strawberry.relay.utils import SliceMetadata
 from strawberry.schema.schema import Schema
@@ -82,9 +82,12 @@ if TYPE_CHECKING:
     from collections.abc import Generator
 
     from django.contrib.contenttypes.fields import GenericRelation
+    from strawberry.relay import Edge
     from strawberry.types.execution import ExecutionContext
     from strawberry.types.field import StrawberryField
     from strawberry.utils.await_maybe import AwaitableOrValue
+
+    from strawberry_django.pagination import OffsetPaginationInfo
 
 
 __all__ = [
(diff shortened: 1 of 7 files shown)


strawberry_django/optimizer.py
```diff
@@ -582,19 +585,34 @@ def _optimize_prefetch_queryset(
                 connection_type is relay.ListConnection
                 or connection_type is DjangoListConnection
             ):
+                field_def_ = connection_type_def.get_field("edges")
+                assert field_def_
+                field_ = field_def_.resolve_type(type_definition=connection_type_def)
+                while isinstance(field_, StrawberryContainer):
+                    field_ = field_.of_type
```
bellini666: You can use `unwrap_type` from `utils/typing.py` in here

strawberry_django/relay/list_connection.py
```diff
@@ -69,6 +113,58 @@ def resolve_connection(
...
+            if (last or 0) > 0 and before is None:
+                # The idea here is to prevent fetching entire table by getting
+                # total_count and using it ad before
+                type_def = get_object_definition(cls)
+                assert type_def
+                field_def = type_def.get_field("edges")
+                assert field_def
+                field = field_def.resolve_type(type_definition=type_def)
+                while isinstance(field, StrawberryContainer):
+                    field = field.of_type
```
bellini666: Ditto

strawberry_django/relay/list_connection.py
```diff
@@ -69,6 +113,58 @@ def resolve_connection(
...
+                edge_class = cast("relay.Edge[relay.NodeType]", field)
+
+                if in_async_context():
+
+                    async def inject_before():
+                        total_count = await nodes.acount()
```
bellini666: I don't remember if `.acount` exists in the minimum django version we support? If not, we might need to change this to `await sync_to_async(nodes.count)()`

strawberry_django/relay/list_connection.py
```diff
@@ -25,6 +63,11 @@ class DjangoListConnection(relay.ListConnection[relay.NodeType]):
     def total_count(self) -> Optional[int]:
         assert self.nodes is not None
 
+        try:
+            return self.edges[0].node._strawberry_total_count  # type: ignore
+        except (IndexError, AttributeError):
+            pass
```
bellini666, line 70: Isn't this what `get_total_count` below is doing? It has an extra step of trying to get from the cache, to avoid triggering a DB query in case the nodes have not been resolved yet

Or is this here for a specific reason?

strawberry_django/relay/list_connection.py
```diff
@@ -69,6 +113,58 @@ def resolve_connection(
                 conn.nodes = nodes
                 return conn
 
+        if queryset_config and queryset_config.optimized:
+            assert isinstance(nodes, models.QuerySet)
```
bellini666: Maybe move this to a separate `resolve_optimized_connection`, like the `if` above?

And then rename the existing `resolve_connection_from_cache` to `resolve_optimized_connection_by_prefetch`

bellini666 (changes requested):
Left some suggestions, but in general looks good! Will take another look after the adjustments

> I am not very confident with the solutions for total_count in strawberry_django/relay/list_connection.py.

I've commented there. Is it related or something else?

strawberry_django/relay/list_connection.py
Kitefiko: 4.2 is lowest supported version and `acount()` was [supported](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.acount)
PR tests would have catch this, no?

strawberry_django/relay/list_connection.py
Kitefiko: `get_total_count` tries firstly to get a value from `_result_cache`, but that exists only for nested queries (prefetched relations), for "root" queries `_strawberry_total_count` is annotated and available only on row data 
Coincidentally this approach works for for nested queries `total_count` too.

bellini666 (approved):
Thank you for this! ❤️

PR #795, opened 2025-10-12
feat: support for python 3.14

bellini666:
* Add support for Python 3.14
* Drop support for the now EOLed Python 3.9
* Bump minimum strawberry version to 0.283.2
* Replace psycopg2 with psycopg (at version 3), as the former doesn't build on Python 3.14

## Summary by Sourcery

Enable support for Python 3.14 by updating project metadata, raising the minimum Python requirement to 3.10, upgrading dependencies, and aligning linting and pre-commit configurations.

New Features:
- Add Python 3.14 classifier to project metadata

Enhancements:
- Raise minimum supported Python version to >=3.10,<4.0
- Upgrade strawberry-graphql, django-types, pytest-cov, ruff, psycopg, and other dependencies to their latest compatible versions
- Replace psycopg2 packages with psycopg v3.2.10
- Update Ruff target-version to py310 and Pyright pythonVersion to 3.10

CI:
- Bump the Ruff pre-commit hook revision to v0.14.0


strawberry_django/fields/filter_order.py
```diff
@@ -95,7 +95,7 @@ def type_annotation(self) -> StrawberryAnnotation | None:
 
                 annotation = ordering.Ordering
 
-            return StrawberryAnnotation(Optional[annotation])
+            return StrawberryAnnotation(annotation | None)
```
g-as, line 98: This now raises:
`TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'` (I'm on `3.12`).

bellini666: Strange that this have not broken in CI

Is that breaking in any place where you are using `__future__` annotations? Or something that can help me track it down?

g-as: Ha, maybe, let me check.

g-as: Yes indeed.

g-as: I can confirm that there's no issue when removing `__future__` annotations and using str annotations.

bellini666: This should fix it: https://github.com/strawberry-graphql/strawberry-django/pull/800

Releasing it right now

g-as: It works for most of my filters, but not the ones containing a `@strawberry_django.filter_field`, which yields the same:
```python
/.venv/lib/python3.12/site-packages/strawberry_django/fields/filter_order.py:98: in type_annotation
    return StrawberryAnnotation(annotation | None)
                                ^^^^^^^^^^^^^^^^^
E   TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

I'll file a proper issue with min repro case if you cannot reproduce easily.

And in any case, thanks ;)

bellini666: Oh, my bad here, I checked all places but somehow missed that one :(

Releasing 0.66.2 right now

g-as: All good. Thanks a lot!

Issue #792, opened 2025-10-03
Wrong connection total_count when use DISTINCT


iamcrookedman:
When there is a filter by m-2-m or 1-2-m relation and DISTINCT=true, sometimes total_count is calculated wrong.
Example response:
```
{
  "data": {
    "standConnection": {
      "edges": [
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjA=",
          "node": {
            "id": "U3RhbmQ6NTY2NQ=="
          }
        }
      ],
      "totalCount": 2
    }
  }
}
```

bellini666:
Hi @iamcrookedman ,

Could you provide an example and/or MRE for this issue?

iamcrookedman:
In my case, it looks like this:

```
class A(models.Model):
    ...

class B(models.Model):
    a = models.ForeignKey(A, related_name="rel_to_b", on_delete=models.CASCADE)
    name = models.TextField()

@strawberry_django.filter_type(model=A, name="AFilter" lookups=True)
class AGqlFilter:
    rel_to_b: BGqlFilter | None

@strawberry_django.filter_type(B, lookups=True)
class BGqlFilter:
    name: auto
    a: AGqlFilter | None
```

Query:
```
query ($first: Int, $after: String, $filters: AFilter!) {
  aConnection(
    first: $first
    after: $after
    filters: $filters
  ) {
    edges {
      cursor
      node {
        id
      }
    }
    totalCount
  }
}
```

Variables:
```
{
    "filters": {
      "relToB": {
        "name": { "inList": ["a", "b"] }
      },
      "DISTINCT": true
    },
    "first": 100
}
```

Issue #788, opened 2025-09-30
N+1 queries due to queryset.count() if cache is empty


mronian:
## Describe the Bug

In a nested `ListConnectionWithTotalCount`, if the child has `total_count` in the query, if there is a parent node with no related child nodes, there is a `queryset.count()` triggered. This leads to N+1 queries like scenario if very few parent nodes have child nodes.  

https://github.com/strawberry-graphql/strawberry-django/blob/269954c582dfa38dc93af95abd2a275888f50103/strawberry_django/pagination.py#L269-L298

Instead, if the queryset is optimized (`is_optimized_by_prefetching` is `True`) then it should just return 0 assuming no nodes were found, calling the `count()` method only for unoptimized querysets.

bellini666:
Hi @mronian ,

I tried here adding a:

```python
elif results is not None:
    return 0
```

And an issue can be seen in this line: https://github.com/strawberry-graphql/strawberry-django/blob/fix-count/tests/test_optimizer.py#L1622-L1622

If we find a way to prevent that (i.e. that test to not fail), then we can improve this

Issue #772, opened 2025-07-10
N+1 when using a Prefetch without explicit ordering


taobojlen:
## Describe the Bug

When using `prefetch_related` on a resolver, unless you explicitly define an ordering, there's an N+1.

```python
@strawberry_django.type(MyModel)
class MyModelType(strawberry.relay.Node):
    # this causes an N+1 on MyModel.foo_set when it's resolved
    @strawberry.django.field(prefetch_related=[Prefetch("foo_set", queryset=Foo.objects.filter(bar="baz")])
    def my_resolver(self) -> list[Foo]:
        # do something
```

We have a number of resolvers with custom querysets in `prefetch_related`. When I updated from strawberry-django 0.54.0 to 0.62.0, I started getting N+1 errors (via [django-zeal](https://github.com/taobojlen/django-zeal), a library to detect N+1s). I also saw some of my tests failing because the DB query count was higher than expected.

Using git bisect, I tracked this down to a specific change in strawberry-django: https://github.com/strawberry-graphql/strawberry-django/pull/715

Since then, if using custom querysets in your prefetches, it seems you have to add an explicit `.order_by()` clause to avoid N+1s.

## System Information

 - Operating system: MacOS 15.5
 - Python version: 3.12.8
 - Strawberry version (if applicable): 0.275.5

bellini666:
Hrm, I've seen other reports like this and am wondering if we should revert #175 

@SupImDos as you added it, do you know what might be going on? And what side-effects we need to workaround, or inform as a "breaking change", in case we need to revert it?

Or maybe we can fix it differently? I'll also try to take a look at this

SupImDos:
> @SupImDos as you added it, do you know what might be going on? And what side-effects we need to workaround, or inform as a "breaking change", in case we need to revert it?

#715 was added so that we never return multiple records from an unordered QuerySet, as it would be non-deterministic - which still seems a sensible assumption to me. If we were to revert this, the results from querying models with no default `Meta.ordering` would become non-deterministic again, and end users would either need to:

a) Add a default `Meta.ordering` to their model
b) Add a `get_queryset` method to their Strawberry type that applies `.order_by(...)`

I would need a more complete test case / reproducible example to understand the cause of the N+1s though. Maybe the `.order_by("pk")` is being applied too late in the process - and the ordering is somehow breaking the prefetch cache causing an N+1?

I can look into this too 🙂 

---

[...]

Issue #771, opened 2025-07-09
N+1 if query with a fragment and custom connection class (introduced in 0.59.0)


rcybulski1122012:
GraphQL query cause N+1 database queries when a fragment is used with a custom implementation of Connection

## Describe the Bug

Repository to clone, that reproduces the bug: https://github.com/rcybulski1122012/strawberry-django-bug-reproduction

Here's the minimal setup that allows to reproduce it:

models.py

```python
from django.db import models

class ParentModel(models.Model):
    name = models.CharField(blank=True, default='', max_length=255)
    child = models.ForeignKey('ChildModel', on_delete=models.CASCADE, null=True, blank=True)


class ChildModel(models.Model):
    name = models.CharField(blank=True, default='', max_length=255)
    grandchild = models.ForeignKey('GrandChildModel', on_delete=models.CASCADE, null=True, blank=True)


class GrandChildModel(models.Model):
    name = models.CharField(blank=True, default='', max_length=255)
```

schema.py
```python

from typing import Optional, Any, Self

import strawberry
import strawberry_django
from django.db.models import QuerySet
from strawberry import relay, auto, Info
from strawberry_django.optimizer import DjangoOptimizerExtension
from strawberry_django.relay import DjangoListConnection

from bug_test.models import ParentModel, GrandChildModel, ChildModel


@strawberry_django.type(GrandChildModel)
class GrandChildType(relay.Node):
    name: auto


@strawberry_django.type(ChildModel)
class ChildType(relay.Node):
    name: auto
    grandchild: GrandChildType


@strawberry_django.type(ParentModel)
class ParentType(relay.Node):
    name: auto
    child: ChildType


@strawberry.type()
class CustomConnection(relay.Connection[ParentType]):
    @classmethod
    async def resolve_connection(
            cls,
            nodes: QuerySet[ParentModel],
            *,
            info: Info,
            before: Optional[str] = None,
            after: Optional[str] = None,
            first: Optional[int] = None,
            last: Optional[int] = None,
            **kwargs: Any,
    ) -> Self:
        return await DjangoListConnection.resolve_connection(
            nodes=nodes,
            info=info,
            before=before,
            after=after,
            first=first,
            last=last,
        )


def resolver() -> list[ParentType]:
    return ParentModel.objects.all()


@strawberry.type
class Query:
    some_types = strawberry_django.connection(
        graphql_type=CustomConnection,
        resolver=resolver,
    )

schema = strawberry.Schema(
[...]

PR #808, opened 2025-11-05
perf: Improve optimizer for polymorphism and connections

stygmate:
**Summary**
This PR substantially improves the Django optimizer to better support polymorphic models, nested reverse relations, and Relay connections. It introduces a post-fetch prefetching pass that batches reverse relations after queryset evaluation and injects them into Django’s `_prefetched_objects_cache`. It also adds dedicated handling for `InheritanceManager` (django-polymorphic) and enhances Relay connection resolution to avoid extra queries (including `totalCount`).

**Key changes**
- Add post-fetch prefetching
  - New module `strawberry_django/postfetch.py` and integration into `default_qs_hook`.
  - Defers certain reverse lookups to a postfetch phase, batching by FK and filling nested caches.
  - Supports both child-level and parent-level postfetch branches collected during optimization.
- Polymorphism and InheritanceManager
  - Rewrite prefetch paths after `select_subclasses` so subclass instances get correct lookups and avoid redundant queries.
  - Collect and merge postfetch hints across interface/union selections and concrete types.
- Relay/pagination
  - If a queryset is optimized-by-prefetching, use the specialized connection path, otherwise compute `totalCount` via a window `Count(1)` in the main query when requested, avoiding a separate `count()` query.
  - Materialize querysets that carry postfetch hints before connection construction to reuse caches in nested connections.
- Field and fetch hooks
  - Connection/list resolvers can reuse an existing `_prefetched_objects_cache` on the source instance.
  - Single-object fields avoid `.get()` when a prefetched result cache is present to prevent LIMIT queries discarding prefetched data.
- Tests and docs
  - Extensive new tests for polymorphism, Relay, and InheritanceManager, including postfetch branches and query behavior.
  - Documentation updates in `docs/guide/optimizer.md`.

**Configuration**
- `OptimizerConfig` retains existing flags (`enable_only`, `enable_select_related`, `enable_prefetch_related`, `enable_annotate`, `enable_nested_relations_prefetch`) and adds support pathways for `postfetch_prefetch` and parent-branch hints carried via `OptimizerStore`.
- No breaking changes to the public API; the postfetch mechanism is internal and triggered automatically by the optimizer.

**Performance and behavior**
- Reduces N+1s on nested reverse relations (particularly under polymorphic models and nested connections) by batching and caching after evaluation.
- Avoids redundant or path-mismatched prefetches when using `InheritanceManager.select_subclasses` by rewriting prefetch paths relative to the subclass.
- Prevents extra `count()` queries for `totalCount` by using SQL window functions when the field is selected.

**Related issues**
- Likely fixes or improves: #593, #772, #788.
- Likely improves generic N+1 patterns around connections/fragments: #771.
- Possibly affected (needs validation): #792 (DISTINCT counts), #743/#549 (nested annotations).

**Notes**
- The PR keeps the optimizer conservative on already-evaluated querysets and lists (no re-optimization, respects `result_cache`).
- The new logic is exercised by the added test suites, especially under `tests/polymorphism*`, `tests/polymorphism_relay*`, and `tests/polymorphism_inheritancemanager*`.

**Review note**
This one’s a heavyweight and touches the optimizer’s core. It needs a reviewer who knows the library well and can take the time for a slow, careful pass @bellini666 ☕🙂. Also, if anyone wants to experiment with this version in their projects, please do—real-world mileage reports are pure gold.


stygmate:
I realized that under certain conditions, evaluating a queryset was loading all the records from the tables. I’ll look for fixes.

stygmate:
I fixed the bug I detected earlier today. I also did quite a bit of refactoring to meet Pyright and Ruff requirements and to make my changes clearer.

stygmate:
I have code branches that aren’t covered by tests. Given the complexity of handling optimizations related to polymorphism, I worked iteratively (hence the large number of commits), and some parts of the code may have become unnecessary. That said, I’m seeing very promising results.

tests/polymorphism_relay/test_excessive_materialization.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def find_projects_in_query(sql: str) -> bool:
+        return project_table in sql
+
+    projects_sql = [q["sql"] for q in ctx if find_projects_in_query(q["sql"])]
+
+    # If a projects query exists, ensure it does NOT batch across multiple company ids.
+    # It's acceptable that no projects query is executed if data was served from cache
+    # after page-level postfetch populated it.
+    if projects_sql:
```
bellini666, line 117: **question:** does this `if` makes sense in a test context? I mean, this should be deterministic, unless this was a pytest parametrized value

tests/polymorphism_inheritancemanager_relay/test_parent_postfetch.py
```diff
@@ -0,0 +1,84 @@
...
+            for n_edge in art_notes_conn.get("edges", []):
+                note_node = (n_edge or {}).get("node") or {}
+                details_conn = note_node.get("details") or {}
+                for d_edge in details_conn.get("edges", []):
+                    d_node = (d_edge or {}).get("node") or {}
+                    text = d_node.get("text")
+                    if text:
+                        details_texts.add(text)
+
+    assert {"d11", "d12", "d21"}.issubset(details_texts)
```
bellini666, line 84: **suggestion:** maybe we should just do a simpler `assert result.data == {...}`? Seems easier to visualize the output, and the code is less complex

tests/polymorphism_inheritancemanager_relay/test_excessive_materialization.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def find_projects_in_query(sql: str) -> bool:
+        return project_table in sql
+
+    projects_sql = [q["sql"] for q in ctx if find_projects_in_query(q["sql"])]
+
+    # If a projects query exists, ensure it does NOT batch across multiple company ids.
+    # It's acceptable that no projects query is executed if data was served from cache
+    # after page-level postfetch populated it.
+    if projects_sql:
```
bellini666, line 117: **suggestion:** I started reviews from bottom to top. Ditto from same suggestion in the other similar test

tests/polymorphism_inheritancemanager/test_parent_postfetch.py
```diff
@@ -0,0 +1,66 @@
...
+    art_projects = [
+        p for p in companies[0]["projects"] if p["__typename"] == "ArtProjectType"
+    ]
+    details_texts = {
+        d["text"]
+        for p in art_projects
+        for n in p.get("artNotes", [])
+        for d in n.get("details", [])
+    }
+    assert {"d11", "d12", "d21"}.issubset(details_texts)
```
bellini666, line 66: **suggestion:** I commented about this below, I think just doing `result.data == {...}` is better

tests/polymorphism_inheritancemanager/test_excessive_materialization.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def find_projects_in_query(sql: str) -> bool:
+        return project_table in sql
+
+    projects_sql = [q["sql"] for q in ctx if find_projects_in_query(q["sql"])]
+
+    # If a projects query exists, ensure it does NOT batch across multiple company ids.
+    # It's acceptable that no projects query is executed if data was served from cache
+    # after page-level postfetch populated it.
+    if projects_sql:
```
bellini666, line 91: **issue:** ditto below

strawberry_django/resolvers.py
```diff
@@ -12,6 +12,9 @@
 from strawberry.utils.inspect import in_async_context
 from typing_extensions import ParamSpec
 
+# Post-fetch utilities used by default_qs_hook
```
bellini666, line 15: **nitpick:** we don't need this comment

strawberry_django/fields/field.py
```diff
@@ -286,13 +301,54 @@ async def async_resolver():
...
+                # bypass pagination because the queryset already carries LIMIT/OFFSET.
+                from strawberry_django.queryset import (
+                    get_queryset_config as _get_qs_cfg,
+                )
+
+                cfg = _get_qs_cfg(qs2)
+                if getattr(cfg, "parent_postfetch_branches", None):
+                    from strawberry_django.resolvers import default_qs_hook as _dqsh
+
+                    qs2 = _dqsh(qs2)
```
bellini666, line 330: **question:** why the aliases here? I think you can just import and use their names directly

strawberry_django/fields/field.py
```diff
@@ -286,13 +301,54 @@ async def async_resolver():
...
+                        # Only short-circuit to the cache if the queryset does NOT carry
+                        # postfetch hints. If it does, we must run the default_qs_hook so
+                        # nested postfetch can execute on this queryset.
+                        from strawberry_django.queryset import (
+                            get_queryset_config as _get_qs_cfg,
+                        )
+
+                        cfg = _get_qs_cfg(qs)
+                        if not getattr(cfg, "postfetch_prefetch", None):
+                            return cache[use_cache_key]
```
bellini666, line 351: **nitpick:** Ditto

Also, can't we just import those at module-level? (don't remember if they would cause import loops)

strawberry_django/fields/field.py
```diff
@@ -464,13 +520,75 @@ def resolve(
         assert self.connection_type is not None
         nodes = cast("Iterable[relay.Node]", next_(source, info, **kwargs))
 
+        # Helper to apply early SQL pagination for simple forward root connections
+        def _apply_early_pagination(qs: models.QuerySet):
+            import contextlib
```
bellini666, line 525: **nitpick:** this should be imported at module level

strawberry_django/fields/field.py
```diff
@@ -464,13 +520,75 @@ def resolve(
...
+            page_limit = first
+            if isinstance(self.max_results, int) and page_limit is not None:
+                page_limit = min(page_limit, self.max_results)
+
+            # Ensure deterministic ordering
+            if not qs.ordered:
+                qs = qs.order_by("pk")
+
+            # Build ORDER BY expressions as Django does for window pagination
+            from django.db import DEFAULT_DB_ALIAS
```
bellini666, line 549: **nitpick:** ditto

strawberry_django/fields/field.py
```diff
@@ -464,13 +520,75 @@ def resolve(
...
+
+            # Use the public `db` property instead of the private `_db` attribute
+            with contextlib.suppress(Exception):
+                compiler = qs.query.get_compiler(using=(qs.db or DEFAULT_DB_ALIAS))
+                order_by_exprs = [expr for expr, _ in compiler.get_order_by()]
+
+                # Annotate row number and total count (global partition)
+                from django.db.models import Count, Window
+                from django.db.models.functions import RowNumber
```
bellini666, line 558: **nitpick:** ditto

strawberry_django/fields/field.py
```diff
@@ -464,13 +520,75 @@ def resolve(
...
+                    _strawberry_row_number=Window(
+                        RowNumber(),
+                        partition_by=None,
+                        order_by=order_by_exprs,
+                    ),
+                    _strawberry_total_count=Window(
+                        Count(1),
+                        partition_by=None,
+                    ),
+                )
```
bellini666, line 570: **thought:** this is duplicating some code that we have on pagination module. Also, maybe this whole function should be there, as it isnot related to the base field itself

bellini666 (changes requested):
Left some initial comments, but still need to take another look after.

Also, I hate doing this, but would it be possible to split this into multiple PRs? We have so much going on here that it is really hard to understand what is going on in all the changes

stygmate:
@bellini666 I have a huge amount of work at the moment. I’ll try to come back to this topic in a week. In the meantime, beyond the review, if you have a bit of time it would be interesting to test it on a real project (ideally one that uses polymorphism).

PR #825, opened 2025-12-10
feat: Add string-based lookups for UUID fields

Akay7:
## Description

Fix filters for UUID fields. 

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Unable to filter by part of UUID, appears exception that UUID is expected when when filter `contains` or `regex`

## Checklist

## Summary by Sourcery

Allow string-based lookups on UUID fields and add coverage for UUID filtering behavior.

Bug Fixes:
- Fix UUID field filtering to accept string values for partial and regex-based lookups instead of raising type errors.

Enhancements:
- Adjust generic filter lookup definitions so containment and regex-style operators use string inputs, improving compatibility with non-string field types like UUIDs.

Tests:
- Add a UUID-backed test model and GraphQL query type to verify partial-string filtering on UUID fields.
- Tidy existing GraphQL filter tests by removing a redundant error assertion.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
       - id: check-xml
       - id: check-symlinks
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.7
+    rev: v0.14.9
     hooks:
       - id: ruff-format
       - id: ruff


strawberry_django/fields/filter_types.py
```diff
@@ -56,28 +56,28 @@ class FilterLookup(BaseFilterLookup[T]):
     i_exact: T | None = filter_field(
         description=f"Case-insensitive exact match. {_SKIP_MSG}"
     )
-    contains: T | None = filter_field(
```
Akay7, line 59: @bellini666 for now `FilterLookup` only used for str and UUID, so there's should be no issues to set str for filters. But if there are some users using FilterLookup for non str based datatypes there could be issues. But those users will also have issues with `regex` filters, because regex should be `str` based.

But if you think that we should have one more type of filter additionally to `FilterLookup` I can add it.

bellini666: what worries me here is other people using this for custom stuff, and that can end up being a major breaking change.

Maybe we should create a `class StrFilterLookup(FilterLookup[str])`? Wdyt?

Akay7: Yeah, let's be on the safe side. I'll add a separate type.

But I think a separate filter is needed for UUID, something like `UUIDFilterLookup`, for str `FilterLookup[str]` works well.

strawberry_django/fields/filter_types.py
```diff
@@ -74,10 +74,41 @@ class FilterLookup(BaseFilterLookup[T]):
...
         description=f"Case-sensitive regular expression match. {_SKIP_MSG}"
     )
-    i_regex: T | None = filter_field(
+    i_regex: str | None = filter_field(
+        description=f"Case-insensitive regular expression match. {_SKIP_MSG}"
+    )
+
+
+@strawberry.input
+class UUIDFilterLookup(BaseFilterLookup[T]):
```
Akay7, line 86: So I made another filter. For now it's used only for UUID filtering, but actually could be used for other types. 

I have tried to reuse `FilterLookup[str]` but got issues with that, so I think for now it's better to keep it simple. 
1. Used that only for filtering UUID and named it as the place where it used. If there will be other cases where it can be used, then it can be renamed accordingly. (Well, I haven’t yet been able to think of anything else besides filtering as text for UUID fields)
2. To have some code duplication, but have correct annotations. (without monkey patching, overrides at related packages, extra branches in the code, dark magic)

But @bellini666 if you know the say how to make it more eloquent feel free to share and I will be happy to refactor.

strawberry_django/fields/filter_types.py
```diff
@@ -74,10 +74,41 @@ class FilterLookup(BaseFilterLookup[T]):
     i_ends_with: T | None = filter_field(
         description=f"Case-insensitive ends-with. {_SKIP_MSG}"
     )
-    regex: T | None = filter_field(
+    regex: str | None = filter_field(
```
Akay7, line 77: 100% sure that there was mistake, and `regex` and `i_regex` couldn't be any other type beside str

strawberry_django/fields/filter_types.py
Akay7: @bellini666 
Also I opened alternative realization https://github.com/strawberry-graphql/strawberry-django/pull/829.

bellini666:
Superseded by https://github.com/strawberry-graphql/strawberry-django/pull/829

PR #829, opened 2025-12-19
feat: Add string-based lookups for UUID fields

Akay7:
## Description

Fix filters for UUID fields. 

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Unable to filter by part of UUID, appears exception that UUID is expected when when filter `contains` or `regex`

## Checklist

## Summary by Sourcery

Allow string-based lookups for UUID fields and extend filter support for partial and regex matching.

New Features:
- Introduce GraphQL filter type and object type for a UUID-backed model to expose UUID fields via the API.

Bug Fixes:
- Fix UUID field filters so that string-based partial lookups and regex queries no longer raise UUID parsing errors.

Enhancements:
- Add a dedicated string filter lookup input type and reuse it for both string and UUID scalar filters to support case-(in)sensitive partial and regex operations.

Tests:
- Add model, schema fields, and parametrized tests to verify string-based and partial lookup behavior for UUID filters in the GraphQL API.

### DIFF
--- a/strawberry_django/fields/filter_types.py
+++ b/strawberry_django/fields/filter_types.py
@@ -82,7 +82,7 @@ class FilterLookup(BaseFilterLookup[T]):
     )
 
 
-@strawberry.input
+@strawberry.input(name="FilterLookup")
 class StrFilterLookup(BaseFilterLookup[T]):
     i_exact: T | None = filter_field(
         description=f"Case-insensitive exact match. {_SKIP_MSG}"


strawberry_django/fields/filter_types.py
```diff
@@ -82,6 +82,37 @@ class FilterLookup(BaseFilterLookup[T]):
     )
 
 
+@strawberry.input
+class StrFilterLookup(BaseFilterLookup[T]):
```
Akay7, line 86: https://github.com/strawberry-graphql/strawberry-django/pull/825

 So there specific filter for `StrFilterLookup` that used for UUID and str. But `FilterLookup[str]` is just for backward compatibility in case if someone implement something with it.

bellini666:
This looks good, but I just noticed the `StrStrFilterLookup`, which I don't like too much

You can probably do `@strawberry.input(name="FilterLookup")` and it should fix it

Akay7:
> This looks good, but I just noticed the `StrStrFilterLookup`, which I don't like too much
> 
> You can probably do `@strawberry.input(name="FilterLookup")` and it should fix it

We could also use `TextFilterLookup` similar to naming of other filters in that file (ie `DateFilterLookup`, `DatetimeFilterLookup`, `ComparisonFilterLookup`)

---

Made changes `@strawberry.input(name="FilterLookup")` and that fix double str prefix.

PR #871, opened 2026-02-28
fix: optimizer skips optimization for aliased fields

### LINKED ISSUE #867: Optimizer does not work if a query uses aliases
## Describe the Bug

Recently, I opened an issue regarding unused arguments when defining a complex `prefetch_related` on a field ([issue link](https://github.com/strawberry-graphql/strawberry-django/issues/863#issuecomment-3932526763)). I have since realized that my provided example fails when using an alias because of how to_attr is handled. I tested it and the error I got surprised me:

"'User' object has no attribute 'prefetched_some_related_objects'"

I expected a Django ORM error regarding duplicate prefetches or overridden values. I attempted to make the `to_attr` value unique, but discovered that the callable provided via prefetch_related is only invoked once, regardless of the number of aliases. In my opinion, it should be called once for every alias to ensure distinct prefetching.

```graphql
query {
  users {
    edges {
      node {
        id
        a: someRelatedObjects {
          id
        }
        b: someRelatedObjects {
          id
        }
      }
    }
  }
}
```

Additionally, I tested whether aliases cause issues when no custom resolver is defined. While the query doesn't crash, it results in N+1 database queries.

## System Information

 - Operating system: macOs 14.8 (23J21)
 - Python version: 3.12
 - Strawberry version (if applicable): 0.288.2

bellini666:
## Summary

- Merge aliased selections with identical arguments into a single optimization pass instead of skipping optimization entirely
- Aliases with different arguments are still skipped (preserving existing safety behavior)
- Replace `Counter`-based deduplication with alias-aware grouping logic in `_get_model_hints()`

Closes #867

## Test plan

- [x] `test_query_prefetch_with_aliases_same_field` — two plain aliases for the same field, verifies optimization works (2 queries vs 3 without)
- [x] `test_query_prefetch_with_aliases_different_subselections` — two aliases with different sub-field selections, verifies merged optimization covers both
- [x] `test_query_nested_connection_with_filter_and_alias` — existing test for aliases with different filter arguments still passes (optimization correctly skipped)
- [x] Full optimizer test suite (150 tests) passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Improve the Django optimizer’s handling of aliased GraphQL field selections to allow safe optimization when aliases share the same arguments while preserving existing safety behavior for differing arguments.

New Features:
- Enable optimization of aliased GraphQL fields when the aliases target the same underlying field with identical arguments by merging their selections.

Bug Fixes:
- Prevent the optimizer from unnecessarily skipping optimization for identical aliased field selections, reducing redundant database queries while keeping behavior correct for aliases with differing arguments.

Enhancements:
- Replace Counter-based duplicate field detection with alias-aware grouping and argument comparison when deriving model hints from query selections.

Tests:
- Add regression tests covering optimization behavior for aliased fields with identical subselections and with differing subselections to validate query count and response structure.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,3 +3,5 @@ release type: patch
 ---
 
 Fix optimizer skipping optimization entirely for aliased fields. When a GraphQL query uses aliases for the same field (e.g., `a: milestones { id }` and `b: milestones { id }`), the optimizer now merges them into a single prefetch instead of skipping optimization, preventing N+1 queries.
+
+Aliases with different arguments (e.g., `a: issues(filters: {search: "Foo"})` and `b: issues(filters: {search: "Bar"})`) are still skipped, since a single prefetch cannot satisfy both filter sets and optimizing one would produce wrong results for the other.


botberry:

---

Fix optimizer skipping optimization entirely for aliased fields. When a GraphQL query uses aliases for the same field (e.g., `a: milestones { id }` and `b: milestones { id }`), the optimizer now merges them into a single prefetch instead of skipping optimization, preventing N+1 queries.

Aliases with different arguments (e.g., `a: issues(filters: {search: "Foo"})` and `b: issues(filters: {search: "Bar"})`) are still skipped, since a single prefetch cannot satisfy both filter sets and optimizing one would produce wrong results for the other.

This release was contributed by @bellini666 in https://github.com/strawberry-graphql/strawberry-django/pull/871

botberry:
This PR was published as [0.76.1](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.76.1). Thank you for contributing!

PR #872, opened 2026-02-28
Pass `Info` instead of `GraphQLResolveInfo` to optimizer callables

rcybulski1122012:
The documentation says that `Info` is passed, but this was no the case.

## Description

In some optimizing functions either an instance of `StrawberryField` as well as `GraphQLResolveInfo` are available. The PR adjust the code so when callable is passed to `prefetch_related` or `annotated` it is executed with `Info`.

This could probably be taken further and create `Info` instance earlier (schema extension?).

That's probably a breaking change.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry-django/issues/864

## Checklist

## Summary by Sourcery

Pass Strawberry `Info` objects instead of `GraphQLResolveInfo` to optimizer callables used by `strawberry_django.field` prefetch/annotate configuration, aligning runtime behavior with documented expectations.

Enhancements:
- Update optimizer internals to construct and propagate Strawberry `Info` instances when resolving `prefetch_related` and `annotate` callables.
- Adjust optimizer hint resolution to derive model properties from the `Info` context rather than directly from the field definition.
- Change prefetch-related typing aliases to expect Strawberry `Info` instead of `GraphQLResolveInfo`.
- Add release metadata marking this behavior adjustment as a patch release.

Tests:
- Add an integration test verifying that `Info` is passed to `prefetch_related` and `annotate` callables when using the query optimizer.

Chores:
- Introduce a top-level release note file describing the change in info type passed to `prefetch_related` and `annotate` callables.

### DIFF
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -514,6 +514,7 @@ class OptimizerStore:
 def _create_strawberry_info(raw_info: GraphQLResolveInfo) -> Info:
     schema: Schema = raw_info.schema._strawberry_schema  # type: ignore
     field = schema.get_field_for_type(raw_info.field_name, raw_info.parent_type.name)
+    assert field
     return schema.config.info_class(raw_info, field)
 
 


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Pass `Info` instead of `GraphQLResolveInfo` to callables provided in `prefetch_related` and `annotate` arguments of `strawberry_django.field`.
```
bellini666: This is "kinda" of a breaking change

Lets make it a minor (not major because we don't have 1.0 yet)

Also, worth adding a note on why this is kinda of a breaking change, but not much, as strawberry info has properties that proxies to the actual resolver, and it is also accepted on places where you would want to pass it inside the prefetch function, like `optimize`, etc

bellini666:
Also, rebase on top of latest `main` to fix autopub workflow :)

botberry:

---

Pass `Info` instead of `GraphQLResolveInfo` to callables provided in `prefetch_related` and `annotate` arguments of `strawberry_django.field`.

This is technically a breaking change because the argument type passed to these callables has changed. However, `Info` acts as a proxy for `GraphQLResolveInfo` and is compatible with the utilities typically used within prefetch or annotate functions, such as `optimize`.

This release was contributed by @rcybulski1122012 in https://github.com/strawberry-graphql/strawberry-django/pull/872

rcybulski1122012:
The type check ~~fails~~ failed because I changed the annotation, which is used in other places in this file. `OptimizerStore.apply` also calls the callables (idk when).

I don't have direct access to the `StrawberryField` instance there , but I can get the field from the `info.schema._strawberry_schema` using `get_field_for_type`. I'm wondering what the point is of `Info` taking` _field` as an argument when it can just get it from `_raw_info` (except performance) 🤔 

BTW, should I change it so all functions associated with the optimizer uses `Info` from strawberry?

bellini666 (approved):
Nice! :)

botberry:
This PR was published as [0.79.0](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.79.0). Thank you for contributing!

PR #873, opened 2026-02-28
fix: auto-add FK field to .only() in prefetch with optimize()

### LINKED ISSUE #862: Prefetch with optimize and only optimization enabled causes N+1
## Describe the Bug

As the title says - if have the only optimization enabled and I optimize a queryset that is prefetched, it is going to cause N+1, problem because the id field that is used to matched the prefetch objects is excluded from the queryset. 

Not sure if it's a bug or expected behaviour but I spend some time debugging it 😅. I solved the problem by disabling the only optimization, by passing the config instance to the optimize.

If it's not a bug I'd appreciate a more specific documentation on optimizer 🙏 

```py
prefetched_fields: list[SomeFieldType] = strawberry_django.field(
    prefetch_related=lambda info: Prefetch(
      "some_fields",
      queryset=optimize(SomeField.objects.filter(...)),
      to_attr="prefetched_fields"
    )
)
```

## System Information

 - Operating system: MacOs 14.8 (23J21)
 - Python version: 3.12

bellini666:
## Summary

- Fix path-walking loop in optimizer that auto-adds FK fields to `.only()` — it was skipping `info.path.typename` (the parent type) and never finding the FK field, causing N+1 queries when using `optimize()` inside `Prefetch`
- Remove manual `OptimizerStore.with_hints(only="project_id")` workarounds from test fixtures (now automatic)
- Add regression test for the fix

Closes #862

## Test plan

- [x] Existing custom prefetch tests pass without the workaround (8 tests)
- [x] New `test_custom_prefetch_optimize_auto_selects_fk` asserts 2 queries, `project_id` in SQL, `due_date` excluded
- [x] Full optimizer test suite passes (152 tests)
- [x] pyright clean on `strawberry_django/optimizer.py`

## Summary by Sourcery

Fix N+1 query issues when using optimize() inside Django Prefetch by ensuring foreign key fields are automatically included in .only() selections and document this as a patch release.

Bug Fixes:
- Ensure the optimizer correctly walks the GraphQL path to auto-include the parent foreign key field in .only() when optimize() is used within Prefetch, preventing N+1 queries.

Enhancements:
- Simplify custom prefetch test fixtures by removing the need for manual OptimizerStore.only() hints now handled automatically by the optimizer.

Documentation:
- Add a RELEASE.md entry describing the patch release and the N+1 fix for optimize() used with Prefetch and .only().

Tests:
- Add a regression test verifying that optimize() inside Prefetch auto-selects the foreign key, executes only two queries, and excludes unrelated fields from the SQL.

### DIFF
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -2185,8 +2185,9 @@ def test_custom_prefetch_optimize_auto_selects_fk(gql_client):
         ]
     }
 
-    prefetch_sql = ctx.captured_queries[1]["sql"]
-    assert Milestone._meta.db_table in prefetch_sql
+    prefetch_sql = next(
+        q["sql"] for q in ctx.captured_queries if Milestone._meta.db_table in q["sql"]
+    )
     assert "project_id" in prefetch_sql
     assert Milestone._meta.get_field("due_date").name not in prefetch_sql
 


botberry:

---

Fix N+1 queries when using `optimize()` inside a `Prefetch` object with `.only()` optimization. The optimizer now correctly auto-adds the FK field needed by Django to match prefetched objects back to their parent.

This release was contributed by @bellini666 in https://github.com/strawberry-graphql/strawberry-django/pull/873

botberry:
This PR was published as [0.76.2](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.76.2). Thank you for contributing!

PR #876, opened 2026-02-28
feat: add skip_queryset_filter to filter_field() for virtual fields

bellini666:
## Summary

- Adds `skip_queryset_filter: bool = False` parameter to `filter_field()` that prevents the field from being applied as a Q object during filter processing
- The field remains in the GraphQL input type and is accessible via `self.<field>` in custom filter methods
- Prevents `_process_type()` from converting `skip_filter` fields into `StrawberryDjangoField`, avoiding model field resolution errors

Closes strawberry-graphql/strawberry#4252

## Example

```python
@strawberry_django.filter_type(MyModel)
class MyFilter:
    min_similarity: float | None = strawberry_django.filter_field(
        default=0.3, skip_filter=True
    )

    @strawberry_django.filter_field
    def filter(self, queryset: QuerySet, prefix: str):
        if self.min_similarity is not None:
            queryset = queryset.filter(
                similarity__gte=self.min_similarity
            )
        return queryset, Q()
```

## Test plan

- [x] New test `test_skip_filter_field` verifies the field is skipped in `process_filters()` and accessible via `self`
- [x] Full test suite passes (1156 passed)

## Summary by Sourcery

Add support for marking filter fields that should be excluded from automatic queryset filtering while remaining available to custom filter logic.

New Features:
- Introduce a skip_filter option on filter_field to keep fields in the GraphQL filter input but omit them from automatic Q-based filtering.

Bug Fixes:
- Prevent skip_filter filter fields from being converted into StrawberryDjangoField instances, avoiding model field resolution errors during type processing.

Tests:
- Add a regression test ensuring skip_filter fields are ignored by process_filters yet remain accessible on the filter instance.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,10 +14,10 @@ class FruitFilter:
     )
 
     @strawberry_django.filter_field
-    def filter(self, queryset: QuerySet, prefix: str):
+    def search(self, info: Info, queryset: QuerySet[models.Fruit], value: str, prefix: str):
         if self.min_similarity is not None:
             queryset = queryset.annotate(
-                similarity=TrigramSimilarity(f"{prefix}name", self.search)
+                similarity=TrigramSimilarity(f"{prefix}name", value)
             ).filter(similarity__gte=self.min_similarity)
         return queryset, Q()
 ```


botberry:

---

Add `skip_queryset_filter` parameter to `filter_field()` for declaring virtual (non-filtering) fields on filter types.

Fields marked with `skip_queryset_filter=True` appear in the GraphQL input type but are not applied as database filters. They are accessible via `self.<field>` in custom filter methods, making them useful for passing parameters like thresholds or configuration values.

```python
@strawberry_django.filter_type(models.Fruit)
class FruitFilter:
    min_similarity: float | None = strawberry_django.filter_field(
        default=0.3, skip_queryset_filter=True
    )

    @strawberry_django.filter_field
    def search(self, info: Info, queryset: QuerySet[models.Fruit], value: str, prefix: str):
        if self.min_similarity is not None:
            queryset = queryset.annotate(
                similarity=TrigramSimilarity(f"{prefix}name", value)
            ).filter(similarity__gte=self.min_similarity)
        return queryset, Q()
```

This release was contributed by @bellini666 in https://github.com/strawberry-graphql/strawberry-django/pull/876

botberry:
This PR was published as [0.78.0](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.78.0). Thank you for contributing!

PR #910, opened 2026-05-23
fix: Date/Time/Datetime filter lookups without type parameter

### LINKED ISSUE #907: DatetimeFilterLookup is missing type
## Describe the Bug

 When using `DatetimeFilterLookup`.
Now I get the the following error:

> Expected type arguments for generic class "DatetimeFilterLookup"

Solution:
Using `DatetimeFilterLookup[datetime]` solved the issue.

## System Information

 - Operating system:
 - Python version:
 - Strawberry version (if applicable): 0.82.1

## Additional Context

Pretty much a dupe of https://github.com/strawberry-graphql/strawberry-django/issues/885 which is resolved but that was for `Str` instead of `datetime`. Should also apply to `DateFilterLookup`

bellini666:
Closes #907.

`DateFilterLookup`, `TimeFilterLookup` and `DatetimeFilterLookup` were declared `Generic[T]` but always used with a fixed type, so referencing them without `[T]` raised `TypeError: ... is generic, but no type has been passed`. Made them concrete (mirroring the `StrFilterLookup` fix from #891) and refactored the auto-resolution path at `fields/types.py` so concrete lookups aren't subscripted. Also dropped the now-unneeded `__class_getitem__` hack from `StrFilterLookup`.

## Breaking change

Generated GraphQL input names lose their type prefix: `DateDateFilterLookup` becomes `DateFilterLookup`. Clients referencing these inputs by name need to update. Bumping as `minor` for this reason.

## Test plan

- [x] Added repro tests for Date/Time/Datetime lookups without type param
- [x] All 1171 tests pass, pyright and ruff clean
- [x] Schema snapshot updated to reflect the rename

## Summary by Sourcery

Make date, time, and datetime filter lookups concrete types and adjust filter type resolution to avoid subscripting non-generic lookups.

Bug Fixes:
- Allow DateFilterLookup, TimeFilterLookup, and DatetimeFilterLookup to be referenced without a type parameter in filters and schema generation.

Enhancements:
- Update DatetimeFilterLookup to define its own date/time component fields instead of inheriting from generic date/time lookups.
- Adjust filter type auto-resolution to only parametrize genuinely generic lookup types, aligning behavior across string and temporal lookups.

Documentation:
- Add RELEASE.md entry documenting the change in filter lookup typing and the resulting GraphQL input name changes.

Tests:
- Add regression tests ensuring Date/Time/Datetime filter lookups work without explicit type parameters and update schema snapshot expectations accordingly.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -9,3 +9,9 @@ release type: minor
 class ProjectFilter:
     due_date: strawberry_django.DateFilterLookup | None
 ```
+
+Migrating:
+
+- Drop the type argument from `StrFilterLookup[str]`, `DateFilterLookup[datetime.date]`, etc. The bare lookup now works; using the bracket form raises `TypeError` at import time.
+- `DatetimeFilterLookup.date` and `.time` now accept `Date` / `Time` values (previously typed as `Int`, which never matched Django's `__date` / `__time` transforms).
+- `TimeFilterLookup.date` and `.time` were removed. Django's `__date` / `__time` transforms don't apply to `TimeField`.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -12,6 +12,6 @@ class ProjectFilter:
 
 Migrating:
 
-- Drop the type argument from `StrFilterLookup[str]`, `DateFilterLookup[datetime.date]`, etc. The bare lookup now works; using the bracket form raises `TypeError` at import time.
+- Drop the type argument from `StrFilterLookup[str]`, `DateFilterLookup[datetime.date]`, etc. The bare lookup now works; the bracket form still resolves to the same class but emits a `DeprecationWarning`.
 - `DatetimeFilterLookup.date` and `.time` now accept `Date` / `Time` values (previously typed as `Int`, which never matched Django's `__date` / `__time` transforms).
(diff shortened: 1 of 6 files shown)


botberry:

---

`DateFilterLookup`, `TimeFilterLookup` and `DatetimeFilterLookup` no longer require a type parameter, matching `StrFilterLookup`. The generated GraphQL input names also lose their type prefix (e.g. `DateDateFilterLookup` becomes `DateFilterLookup`).

```python
@strawberry_django.filter_type(models.Project)
class ProjectFilter:
    due_date: strawberry_django.DateFilterLookup | None
```

Migrating:

- Drop the type argument from `StrFilterLookup[str]`, `DateFilterLookup[datetime.date]`, etc. The bare lookup now works; the bracket form still resolves to the same class but emits a `DeprecationWarning`.
- `DatetimeFilterLookup.date` and `.time` now accept `Date` / `Time` values (previously typed as `Int`, which never matched Django's `__date` / `__time` transforms).
- `TimeFilterLookup.date` and `.time` were removed. Django's `__date` / `__time` transforms don't apply to `TimeField`.

This release was contributed by @bellini666 in https://github.com/strawberry-graphql/strawberry-django/pull/910

botberry:
This PR was published as [0.86.0](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.86.0). Thank you for contributing!

PR #921, opened 2026-06-17
fix: avoid sync_to_async thread hop when resolving node id

### LINKED ISSUE #913: Resolving an id of a Node is very slow
I profiled my application to see what is slowing my query (which returns a lot of data) and it turned out that `Node.resolve_id` was the cause.

## Describe the Bug

I prepared a reproduction with benchmarks - https://github.com/rcybulski1122012/strawberry-django-slow-node-reproduction

Copied from readme, generated by an agent:

When @strawberry_django.type processes a class that inherits from relay.Node, it replaces both resolve_id and resolve_id_attr on the class with django_resolver(resolve_model_id) — but only if the class hasn't defined its own override (strawberry_django/type.py:210–221).

django_resolver wraps functions with @sync_to_async, so in an async context every call goes through:

id field resolver
  → django_resolver(resolve_model_id)
    → async_resolver          # thread hop + contextvars copy
      → sync_resolver
        → resolve_model_id
          → resolve_id_attr   # cheap cache hit

## System Information

 - Operating system: MacOS 26.4.1 (25E253)
 - Python version: 3.14.0
 - Strawberry version (if applicable): -

bellini666:
Fixes #913.

## Problem

Profiling in #913 traced slow `id` resolution to `Node.resolve_id` going through `sync_to_async` on every call.

When `@strawberry_django.type` processes a `relay.Node` subclass, it wraps all four default relay resolvers (`resolve_id`, `resolve_id_attr`, `resolve_node`, `resolve_nodes`) in `django_resolver` and binds them to the class. In an async context, `django_resolver` routes through `@sync_to_async` (`thread_sensitive=True`), which costs a thread hop into asgiref's shared executor plus a contextvars copy on every call.

That cost is warranted for `resolve_node`/`resolve_nodes`, which query the database, but not for `resolve_id`/`resolve_id_attr`:

- `resolve_model_id_attr` is a pure annotation scan with no I/O.
- `resolve_model_id` reads the pk off the in-memory instance via `root.__dict__[id_attr]`. It only touches the DB in the deferred-field fallback, where it calls `django_getattr`, which is itself `django_resolver`-wrapped and already bridges sync/async on its own.

## Fix

Apply `django_resolver` only to `resolve_node`/`resolve_nodes`, and bind `resolve_id`/`resolve_id_attr` directly. `resolve_model_id` needs no change: it already returns the correct `AwaitableOrValue[str]`, a plain `str` on the fast path and an awaitable from `django_getattr` on the deferred path.

## Test

Added a regression test asserting that, inside a running event loop, `resolve_id` returns a plain `str` rather than a coroutine, for both default-pk and explicit-`NodeID` types. Against the old code it fails with `result` being a `SyncToAsync.__call__` coroutine.

## Summary by Sourcery

Avoid routing Relay node ID resolution through async thread hops and document the change as a patch release.

Bug Fixes:
- Prevent Relay node ID resolution from incurring unnecessary sync_to_async thread hops in async contexts.

Enhancements:
- Clarify relay ID resolver docstrings to describe synchronous in-memory ID access and deferred-field behavior.

Documentation:
- Add RELEASE notes describing the performance improvement to Relay node ID resolution.

Tests:
- Add a regression test ensuring resolve_id returns a plain string and not an awaitable when called inside an async context.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,8 @@
+---
+release type: patch
+---
+
+Resolving a Relay node's `id` no longer goes through `sync_to_async` on every call.
+`resolve_id`/`resolve_id_attr` now read the primary key directly off the in-memory
+instance, removing an unnecessary thread hop (and contextvars copy) in async contexts.
+The deferred-field fallback still bridges database access safely.
--- a/strawberry_django/relay/utils.py
+++ b/strawberry_django/relay/utils.py
@@ -283,7 +283,9 @@ def resolve_model_node(
 
 
 def resolve_model_id_attr(source: type) -> str:
-    """Resolve the model id, ensuring it is retrieved in a sync context.
+    """Resolve the name of the model attribute holding the node id.
+
+    This is a pure annotation scan, so it never touches the database.
 
     Args:
     ----
@@ -309,7 +311,11 @@ def resolve_model_id(
     *,
     info: Info | None = None,
 ) -> AwaitableOrValue[str]:
-    """Resolve the model id, ensuring it is retrieved in a sync context.
+    """Resolve the model id, reading it off the in-memory instance.
+
+    The pk is read directly from the instance's ``__dict__``, so the common case
+    stays synchronous. Only a deferred id field falls back to ``django_getattr``,
+    which bridges the resulting database read to the current sync/async context.
 
     Args:
     ----
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -213,17 +213,21 @@ def _process_type(
     # Default querying methods for relay
     if issubclass(cls, relay.Node):
         for attr, func in [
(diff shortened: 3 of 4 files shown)


botberry:

---

Resolving a Relay node's `id` no longer goes through `sync_to_async` on every call.
`resolve_id`/`resolve_id_attr` now read the primary key directly off the in-memory
instance, removing an unnecessary thread hop (and contextvars copy) in async contexts.
The deferred-field fallback still bridges database access safely.

This release was contributed by @bellini666 in https://github.com/strawberry-graphql/strawberry-django/pull/921

botberry:
This PR was published as [0.86.2](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.86.2). Thank you for contributing!

PR #922, opened 2026-06-17
fix: call is_type_of methods in superclasses of Django GraphQL types

### LINKED ISSUE #919: Custom `is_type_of` is overridden when inherited from a parent class
A `is_type_of` method on a `@strawberry_django.interface` cannot be inherited to its subclasses, which use `@strawberry_django.type`.

## Describe the Bug

When defining a `@strawberry_django.interface` you may need to define a `is_type_of` function, to, for example, achieve polymorphism in the GraphQL type system even though in the Django ORM there is just one model (with a type tag field on it). 

However, Strawberry Django puts its own `is_type_of` implementation into every `@strawberry_django.type`, regardless of whether a superclass already defines `is_type_of` (only `cls.__dict__` is checked):
https://github.com/strawberry-graphql/strawberry-django/blob/386922b0b7ecb9462365f799bf393e30f7364f0f/strawberry_django/type.py#L203-L211

## System Information

 - Operating system: Linux
 - Python version: 3.14
 - Strawberry version (if applicable): n/a

## Additional Context

Here is some code of what I would like to achieve:

```py
class Command(models.Model):
  type = models.CharField(choices=...) # this defines the type of command
  # other fields here...

@strawberry_django.interface(Command)
class CommandType:
  matched_type = None

  @classmethod
  def is_type_of(cls, obj, info):
    # BUG: this will never be called, because Strawberry Django puts a is_type_of inside FooCommandType and BarCommandType
    return isinstance(obj, Command) and obj.type == cls.matched_type

  @classmethod
  def get_queryset(cls, qs, info):
    matched_type = cls.matched_type
    if matched_type is not None:
      qs = qs.filter(type=matched_type) # only return the appropriate objects for resolves with a concrete subtype
    return qs

@strawberry_django.type(Command)
class FooCommandType(CommandType):
  matched_type = "foo"

@strawberry_django.type(Command)
class BarCommandType(CommandType):
  matched_type = "bar"
```

Currently I have to define dummy is_type_of methods in every subclass that just call super.

diesieben07:
## Description

Previously, any `strawberry_django.type` would unconditionally add an `is_type_of` method, which replaced any custom logic in parent classes.
This pull request improves the added `is_type_of` method to also call super().is_type_of.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #919

## Checklist

## Summary by Sourcery

Ensure Django Strawberry types respect custom and inherited is_type_of logic while still handling model-based type resolution.

Bug Fixes:
- Preserve and invoke is_type_of implementations from superclasses for strawberry_django.type-generated GraphQL types instead of overwriting them.

Tests:
- Add integration-style tests covering custom, mixin-based, and inherited is_type_of implementations, including behavior with strawberry type casting and schema execution for Fruit-related types.

### DIFF
--- a/strawberry_django/type.py
+++ b/strawberry_django/type.py
@@ -203,10 +203,14 @@ def _process_type(
     # Make sure model is also considered a "virtual subclass" of cls
     if "is_type_of" not in cls.__dict__:
 
-        def is_type_of(obj, info):
+        @classmethod
+        def is_type_of(virtual_cls, obj, info):
             if (type_cast := get_strawberry_type_cast(obj)) is not None:
                 return type_cast is cls
-            return isinstance(obj, (cls, model))
+            super_func = getattr(super(cls, virtual_cls), "is_type_of", None)
+            return (super_func is None or super_func(obj, info)) and isinstance(
+                obj, (cls, model)
+            )
 
         cls.is_type_of = is_type_of
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+---
+release type: patch
+---
+
+`@strawberry_django.type` types no longer overwrite `is_type_of` methods in superclasses.
+Instead, the superclass' result will be taken into account as well.
--- a/tests/test_custom_is_type_of.py
+++ b/tests/test_custom_is_type_of.py
@@ -1,0 +1,144 @@
+import typing
+from typing import Any, cast
+
+import pytest
+import strawberry
+from graphql import GraphQLResolveInfo
+from strawberry.types.base import StrawberryObjectDefinition
+from strawberry.types.cast import cast as strawberry_cast
+
+import strawberry_django
+from tests.models import Fruit
+
+
+@strawberry_django.interface(Fruit)
+class FruitType:
+    name: str
+    sweetness: int
+
+    @classmethod
+    def is_type_of(cls, obj: Any, _info: GraphQLResolveInfo) -> bool:
+        if not isinstance(obj, Fruit):
+            return False
+        is_sweet = obj.sweetness >= 5
+        return (is_sweet and cls == SweetFruitType) or (
+            not is_sweet and cls == SourFruitType
+        )
+
+
+@strawberry_django.type(Fruit)
+class SweetFruitType(FruitType):
+    pass
+
+
+@strawberry_django.type(Fruit)
+class SourFruitType(FruitType):
+    pass
+
+
+@strawberry_django.interface(Fruit)
+class PlainFruitType:
+    name: str
+    sweetness: int
+
+
+@strawberry_django.type(Fruit)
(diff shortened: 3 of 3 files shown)


botberry:

---

`@strawberry_django.type` types no longer overwrite `is_type_of` methods in superclasses.
Instead, the superclass' result will be taken into account as well.

This release was contributed by @diesieben07 in https://github.com/strawberry-graphql/strawberry-django/pull/922

diesieben07:
Note: The GitHub workflows for tests don't seem to use the latest version of Django 6.
This is because `uv pip install 'django==6.0.*'` will install Django 6.0.0 (it seems). `uv pip install --upgrade 'django==6.0.*'` will work fine.
One of my tests used `bulk_create` which hit this regression in Django 6: https://code.djangoproject.com/ticket/36818
I have switched the test to just use plain create, because I didn't want to touch the test workflows in this PR.

strawberry_django/type.py
```diff
@@ -203,10 +203,14 @@ def _process_type(
...
+        def is_type_of(virtual_cls, obj, info):
             if (type_cast := get_strawberry_type_cast(obj)) is not None:
                 return type_cast is cls
-            return isinstance(obj, (cls, model))
+            super_func = getattr(super(cls, virtual_cls), "is_type_of", None)
+            return (super_func is None or super_func(obj, info)) and isinstance(
+                obj, (cls, model)
+            )
 
         cls.is_type_of = is_type_of
```
bellini666, line 215: Should we not decorate with `@classmethod` and instead, do `cls.is_type_of = types.MethodType(is_type_of, cls)`, similarly to how we do below?

bellini666 (commented):
Looks good, thank you! :)

Left one suggestion, let me know what you think and if it makes sense

strawberry_django/type.py
diesieben07: I must admit I have never used `types.MethodType` before.
I used classmethod so that I can have the actual class (`virtual_cls`) In my code that the function is being called on, so that `subclass.is_type_of(...)` will call the right super method.

diesieben07: I think if we're not gonna use `classmethod` (which would enable calling `is_type_of` on a subclass of a `strawberry_django.type` which itself is _not_ a `strawberry_django.type`, I am not sure how useful that is), then we don't need `types.MethodType`, because all that is going to do is pass `cls` in as the first parameter, and that is already captured from the outer `_process_type` scope. If you prefer that I can do that.

bellini666: Ok, I agree, lets go with this :)

bellini666 (approved):

botberry:
This PR was published as [0.86.4](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.86.4). Thank you for contributing!

PR #923, opened 2026-06-19
Expose input extension support

ayushin:
## Summary

Exposes input extension support through the Django integration:

- adds `extend=True` to `strawberry_django.input(...)`
- adds `extend=True` to `strawberry_django.partial(...)`
- adds coverage for rendering `extend input ...` SDL and reading input extension fields in resolvers
- adds coverage that existing `strawberry_django.type(..., extend=True)` works with Strawberry core object extension support

## Dependency

Depends on strawberry-graphql/strawberry#4464. The PR no longer includes local `uv` source overrides; the execution tests were validated locally by putting the sibling Strawberry core checkout on `PYTHONPATH`.

## Related

- Depends on strawberry-graphql/strawberry#4464
- Tracks strawberry-graphql/strawberry#4463

## Validation

- `uv run ruff check strawberry_django/type.py tests/test_type.py pyproject.toml`
- `PYTHONPATH=<strawberry-django>:<strawberry-core> uv run pytest tests/test_type.py::test_input_can_extend_existing_input_type tests/test_type.py::test_partial_can_extend_existing_input_type tests/test_type.py::test_type_can_extend_existing_type -q`

## Summary by Sourcery

Expose Django input and partial decorators to support extending existing GraphQL input types using Strawberry core input extension functionality.

New Features:
- Allow `strawberry_django.input()` to declare input types as extensions via an `extend` flag.
- Allow `strawberry_django.partial()` to declare input types as extensions via an `extend` flag.

Documentation:
- Add release notes documenting exposure of the `extend=True` option on Django input and partial decorators.

Tests:
- Add tests validating that Django input types can extend existing input types and are correctly rendered in SDL and resolvers.
- Add tests ensuring partial input types and object types properly participate in Strawberry core extension support.


botberry:

---

Expose `extend=True` on `strawberry_django.input()` and
`strawberry_django.partial()` so Django input types can extend existing GraphQL
input object types when used with Strawberry core input extension support.

This release was contributed by @ayushin in https://github.com/strawberry-graphql/strawberry-django/pull/923

patrick91:
@ayushin can you give me a real use case for this? I'm struggling to understand how this can be used in practice, given you still need to import all the types in one centralized place?

ayushin:
@patrick91 answered over at strawberry-graphql/strawberry#4464 (comment) — and closing this one, because your question exposed a real problem with how I filed all of this.

I framed the whole contribution around **input** types: the issue, the branch name, and this PR. But our actual use case is **object** types, and we don't use the input side at all. So this PR — the one you looked at — genuinely has no use case behind it. You were right to push back.

What's left here is only consistency: `strawberry_django.type()` already forwards `extend=True` to core, so once core supports it, `input()` and `partial()` would be the odd ones out. That's worth doing eventually, but it shouldn't be an open PR competing for your attention, and it can't land before core does.

Reopening this later if strawberry-graphql/strawberry#4464 lands.

PR #928, opened 2026-06-23
fix: honor DEFAULT_PK_FIELD_NAME in mutation instance lookups

### LINKED ISSUE #927: Mutations ignore `DEFAULT_PK_FIELD_NAME`: `_parse_pk` looks up by `pk=` instead of the configured key
## Describe the bug

`DEFAULT_PK_FIELD_NAME` lets a project use a field other than the Django primary
key as the GraphQL identifier (e.g. a `uuid`/`slug`/external id, or a virtual id
field such as `django-sqids`). **Filters honor it** (`filters.apply` resolves the
configured field), but **mutations do not**.

In `strawberry_django/mutations/resolvers.py`, `_parse_pk` reads the input value
under `key_attr` (defaulting to `DEFAULT_PK_FIELD_NAME`) but then looks the
instance up by **`pk=`**:

```python
# strawberry_django/mutations/resolvers.py — _parse_pk
if isinstance(value, dict):
    if key_attr is None:
        settings = strawberry_django_settings()
        key_attr = settings["DEFAULT_PK_FIELD_NAME"]

    if key_attr in value:
        obj_pk = value[key_attr]
        if obj_pk is not strawberry.UNSET:
            return model._default_manager.get(pk=obj_pk), value   # <-- looks up by pk, not key_attr
        return None, value

    return None, value

return model._default_manager.get(pk=value), None                 # <-- same for a bare id
```

So when `DEFAULT_PK_FIELD_NAME` is **not** the primary key, `update` (and relation
inputs that reference an existing object by that field) execute
`get(pk=<value-of-the-configured-field>)` — which raises `DoesNotExist` or matches
the wrong row, because the value is not the primary key.

## Expected

The lookup should use the configured field when it is not the primary key —
`get(**{key_attr: obj_pk})`, falling back to `pk=` when `key_attr` is the primary
key (name/attname/`"pk"`). This makes mutations consistent with filters, which
already honor `DEFAULT_PK_FIELD_NAME`.

## Reproduction

1. A model with a unique non-pk identifier, e.g. `code = models.CharField(max_length=20, unique=True)`.
2. `STRAWBERRY_DJANGO = {"DEFAULT_PK_FIELD_NAME": "code"}`.
3. An `update` mutation with input `{ code: "ABC", name: "…" }`.

Actual: resolves `model._default_manager.get(pk="ABC")` → `DoesNotExist` (or wrong
row). Expected: `get(code="ABC")`. The same applies to relation inputs that
reference an existing related object by the configured field.

## Related

- #446 added the `DEFAULT_PK_FIELD_NAME` setting (read/filter side) — the mutation
  input lookup in `_parse_pk` was not updated to honor it.
- #348 ("locate model on uuid instead of id") is the user-facing use case this
  blocks for mutations.

## Proposed fix

In `_parse_pk`, when `key_attr` resolves to a non-pk field, look up by that field
instead of `pk` (and likewise for the bare-id branch). Happy to open a PR with a
regression test — the default-pk path stays unchanged.

## Environment

Reproduced against `main` (the `_parse_pk` code above is current on `main`).

ayushin:
Fixes #927.

`_parse_pk` resolved mutation input instances with
`model._default_manager.get(pk=…)` even when `key_attr` (defaulting to the
`DEFAULT_PK_FIELD_NAME` setting) named a non-pk field. Filters already honor the
setting, but mutations did not — so relation inputs (and the bare-id path) that
reference an existing object by the configured field looked the row up by `pk`
with the wrong value, raising `DoesNotExist` / `ValueError`.

This adds a `_pk_lookup` helper that builds the lookup from `key_attr`:
`{key_attr: value}` for a non-pk field, falling back to `{"pk": value}` when
`key_attr` is `None` / `"pk"` / the pk's `name` / `attname` — so the default path
is byte-for-byte equivalent. Applied at both `get()` sites in `_parse_pk`.

### Tests
- New `tests/mutations/test_parse_pk.py` (regression): non-pk-key lookups resolve
  by the configured field; the default-pk path is unchanged. Verified it fails on
  unfixed code (`get(pk='red')` → `ValueError`) and passes with the fix.
- Full suite: **1194 passed, 11 skipped, 0 failed**. `ruff` lint + format clean.

## Summary by Sourcery

Honor the configured primary key field name when resolving instances in mutation input parsing.

Bug Fixes:
- Ensure `_parse_pk` resolves mutation inputs using the configured key attribute (e.g. `DEFAULT_PK_FIELD_NAME`) instead of always querying by the model `pk`.

Tests:
- Add regression tests verifying `_parse_pk` looks up instances by non-pk key attributes and that the default pk-based behavior remains unchanged.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,11 @@
+---
+release type: patch
+---
+
+Honor the `DEFAULT_PK_FIELD_NAME` setting (and per-field `key_attr`) when
+resolving existing instances in mutation inputs. Previously `_parse_pk` read the
+input value under `key_attr` but always looked the instance up by `pk=`, so
+relation inputs and bare-id updates that reference an object by a configured
+non-pk field raised `DoesNotExist` or matched the wrong row. Filters already
+honored the setting; mutations now do too. The default `pk`-based path is
+unchanged.
--- a/strawberry_django/mutations/resolvers.py
+++ b/strawberry_django/mutations/resolvers.py
@@ -56,6 +56,20 @@ _T = TypeVar("_T")
 _M = TypeVar("_M", bound=Model)
 
 
+def _pk_lookup(model: type[Model], key_attr: str | None, value: Any) -> dict[str, Any]:
+    """Build the `get()` lookup that resolves an instance by its key.
+
+    Honors `DEFAULT_PK_FIELD_NAME` the same way filters do: when `key_attr`
+    names a non-pk field, look the instance up by that field; otherwise fall
+    back to `pk` so the default path stays byte-for-byte equivalent.
+    """
+    pk = model._meta.pk
+    assert pk is not None
+    if key_attr in (None, "pk", pk.name, pk.attname):  # noqa: PLR6201
+        return {"pk": value}
+    return {key_attr: value}
+
+
 def _parse_pk(
     value: ParsedObject | strawberry.ID | _M | None,
     model: type[_M],
@@ -79,11 +93,14 @@ def _parse_pk(
         if key_attr in value:
             obj_pk = value[key_attr]
             if obj_pk is not strawberry.UNSET:
-                return model._default_manager.get(pk=obj_pk), value
+                return (
+                    model._default_manager.get(**_pk_lookup(model, key_attr, obj_pk)),
+                    value,
+                )
 
         return None, value
 
-    return model._default_manager.get(pk=value), None
+    return model._default_manager.get(**_pk_lookup(model, key_attr, value)), None
 
 
 def _parse_data(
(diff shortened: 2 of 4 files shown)


botberry:

---

Honor the `DEFAULT_PK_FIELD_NAME` setting (and per-field `key_attr`) when
resolving existing instances in mutation inputs. Previously `_parse_pk` read the
input value under `key_attr` but always looked the instance up by `pk=`, so
relation inputs and bare-id updates that reference an object by a configured
non-pk field raised `DoesNotExist` or matched the wrong row. Filters already
honored the setting; mutations now do too. The default `pk`-based path is
unchanged.

This release was contributed by @ayushin in https://github.com/strawberry-graphql/strawberry-django/pull/928

bellini666 (approved):
Thanks! :)

botberry:
This PR was published as [0.86.7](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.86.7). Thank you for contributing!

PR #930, opened 2026-07-03
feat: support aliasing fields with optimizer hint callables

rcybulski1122012:
## Description

Builds on https://github.com/strawberry-graphql/strawberry-django/pull/882, which taught the optimizer to handle auto-generated relation fields selected multiple times via aliases with different arguments.

This PR extends that to **custom fields with optimizer hint callables** (`annotate=lambda info: ...`, `prefetch_related=lambda info: ...`). Previously, aliasing such a field with different arguments broke: every alias resolved its callable into the same annotation label (or the same prefetch attribute), so the values clashed and one alias silently won.

Now, when a field with callable hints is selected under multiple response keys, the optimizer processes each alias separately: the callable is resolved once per alias with an `info` scoped to that selection, and callable annotations are automatically stored under an alias-scoped label.

Since annotation/`to_attr` names must be unique per alias, the hint callable and the resolver need a shared, deterministic way to derive that name. Three new public helpers provide it:

- `optimizer_hint_key(info)` — returns a unique attribute name for the current selection, derived from its response key (the alias, if aliased). Crucially, it returns the **same value inside a hint callable and inside the resolver**, so it can name a `Prefetch(to_attr=...)` in the hint and locate the value in the resolver. (Internally, the synthetic resolve info built during hint collection now uses the response key as its path key — that's what makes the two phases agree.)
- `get_hint_value(source, info, default_attr=None, *, default=...)` — reads the hint-produced value back: alias-scoped attribute first, then `default_attr` (the plain annotation label used when the field isn't aliased), then `default` (handy when the optimizer is turned off), otherwise raises `AttributeError`.
- `get_field_arguments(info)` — resolves the current selection's argument values (variables included), usable inside hint callables where resolver kwargs aren't available.

Example:

```python
@strawberry_django.type(Milestone)
class MilestoneType:
    @strawberry_django.field(
        annotate=lambda info: Count(
            "issue",
            filter=Q(issue__name__contains=get_field_arguments(info)["nameContains"]),
        ),
    )
    def issues_count_filtered(self, root, info, name_contains: str) -> int:
        return get_hint_value(root, info, "issues_count_filtered")

{
  milestoneList {
    foo: issuesCountFiltered(nameContains: "foo")   # annotated as _strawberry_alias_foo
    bar: issuesCountFiltered(nameContains: "bar")   # annotated as _strawberry_alias_bar
  }
}
```

Notes:

- Behavior with the optimizer turned off is unchanged; resolvers can opt into a fallback via get_hint_value(..., default=...).
- Non-aliased fields keep the plain field.name annotation label, so existing resolvers are unaffected.
- Fields with callable hints are handled once per response key even when the alias arguments are identical (correctness over deduplication).
- Known limitation (documented): dict-form annotations with custom labels (annotate={"_my_label": lambda info: ...}) keep their static label and therefore don't support different-args aliasing — use the single-expression form for that.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

*

## Checklist

## Summary by Sourcery

Enable the Django optimizer to safely handle aliased field selections with argument-dependent optimization hints.

New Features:
- Support alias-specific optimization for custom fields using callable annotation and prefetch hints.
- Expose helpers for deriving selection-scoped hint keys, reading optimized values, and resolving field arguments.

Bug Fixes:
- Prevent aliased selections with different arguments from sharing annotation or prefetch results and silently returning the wrong value.
- Handle aliased relation selections with differing filters, ordering, pagination, and nested selections without conflicting prefetched data.

Enhancements:
- Process callable hints independently for each response key, including aliases with identical arguments, while preserving existing labels for non-aliased fields.
- Resolve field arguments with variables and maintain consistent hint naming between optimization and resolver execution.

Documentation:
- Document usage and limitations of alias-scoped callable optimizer hints and the new public helpers.

Tests:
- Add coverage for aliased annotations, dictionary annotations, callable prefetches, variables, mixed aliased and unaliased selections, and relation argument combinations.

Chores:
- Add minor release notes describing alias-aware optimizer behavior and the new public API.


botberry:

---

Adjust the optimizer to handle when the same field is selected multiple times via aliases with different arguments and the field doesn't have a custom resolver.

Also add support for aliasing custom fields with `annotate`/`prefetch_related` optimizer hint callables, together with three new public helpers (see the [optimizer guide](https://strawberry-graphql.github.io/strawberry-django/guide/optimizer/) for examples):

- `optimizer_hint_key(info)`: a unique, deterministic attribute name for the current field selection (based on the alias), stable between the hint callable and the resolver.
- `get_hint_value(source, info, default_attr=None, *, default=...)`: reads the value produced by an optimizer hint for the current selection.
- `get_field_arguments(info)`: resolves the coerced argument values (including variables and schema defaults) of the current field selection, for use inside hint callables.

Note: inside an optimizer hint callable (and any `get_queryset` reached through the optimizer's prefetch path), `info.path.key` now reports the selection's response key — the alias when the field is aliased — rather than always the field name. This is what makes the per-alias hints deterministic, but code that keys cache or permission logic on `info.path` inside a hint callable should account for it.

This release was contributed by @rcybulski1122012 in https://github.com/strawberry-graphql/strawberry-django/pull/930

strawberry_django/optimizer.py
```diff
@@ -94,14 +96,19 @@
...
     "optimize",
+    "optimizer_hint_key",
 ]
 
 _M = TypeVar("_M", bound=models.Model)
 
 _sentinel = object()
 _annotate_placeholder = "__annotated_placeholder__"
 
+ALIAS_PREFIX = "_strawberry_alias_"
```
bellini666: Lets keep it similar to how `_annotate_placeholder` is defined

```suggestion
_alias_prefix = "_strawberry_alias_"
```

strawberry_django/optimizer.py
```diff
@@ -1347,19 +1508,48 @@ def _get_model_hints(
...
+                None,
+            ))
         else:
-            first_args = _get_field_arguments(groups[0][0])
-            if all(_get_field_arguments(g[0]) == first_args for g in groups[1:]):
-                merged_node_lists.append([node for group in groups for node in group])
+            # Different args - each alias gets its own to_attr
+            for group in groups:
+                alias = group[0].alias
+                to_attr = f"{ALIAS_PREFIX}{alias.value}" if alias else None
```
bellini666: I'm worried here that this can break. Django's `Prefetch.get_current_to_attr` split by `LOOKUP_SEP`, and might treat this as a multi-level lookup in case the alias starts with `_` (e.g. `_foo` would become `_strawberry_alias__foo`

bellini666: Claude raised this concern to me:

**Blocker.** `to_attr = f"{ALIAS_PREFIX}{alias.value}"` breaks whenever the alias contains `__` or starts with `_`. `_strawberry_alias_` + `_a` gives `_strawberry_alias__a`, and Django's `Prefetch.get_current_to_attr()` splits `prefetch_to` on `LOOKUP_SEP`, so it reads that as a multi-level lookup and the query fails.

Reproduced on this branch, works on the merge base:

```graphql
projectsPaginated { results { id
  _a: milestones(filters:{name:{contains:"a"}}) { id }
  _b: milestones(filters:{name:{contains:"b"}}) { id } } }
```

-> `GraphQLError: 'Project' object has no attribute '_strawberry_alias'`, data `None`. Aliases `a__x`/`b__y` fail the same way. Any alias outside `[A-Za-z0-9][A-Za-z0-9_]*`-without-`__` is fatal.

Annotation labels are fine here (Django's `check_alias` accepts `__` in `annotate()`), so only the prefetch path breaks. Sanitizing or hashing the key would fix it, or a separator that can't collide with `LOOKUP_SEP`. Same construction at line 1530.

strawberry_django/optimizer.py
```diff
@@ -1347,19 +1508,48 @@ def _get_model_hints(
...
+    # list of (field_nodes, to_attr)
+    merged_node_lists: list[tuple[list[FieldNode], str | None]] = []
+    for name, groups in field_name_groups.items():
         if len(groups) == 1:
-            merged_node_lists.append(groups[0])
+            merged_node_lists.append((groups[0], None))
+            continue
+
+        s_field = _find_strawberry_field(object_definition, schema, name)
+        if s_field is not None and _field_has_callable_hints(s_field):
```
bellini666: Claude raised this concern to me:

**Blocker.** A field with a callable `prefetch_related` hint selected under two response keys now raises at query time. `_field_has_callable_hints` forces the per-response-key split even when the aliases are argument-free and identical, so the callable resolves once per alias into two `Prefetch` objects with the same `prefetch_to` and no `_optimizer_sentinel` -- `PrefetchInspector.merge` then refuses. `_scoped_label` below only rewrites `annotate` labels; `prefetch_related` passes through unchanged, so nothing disambiguates them.

Reproduced against this PR's own test schema:

```
{ milestoneList { a: myIssues { name } b: myIssues { name } } }
main: 2 queries, {'a': [], 'b': []}
PR:   ValueError: Tried to prefetch 2 queries with different filters
      to the same attribute. Use `to_attr` in this case...
```

Same for `mixedPrefetchAnnotated`.

Related: this branch runs before the same-args comparison, so identical-arg aliases are never merged at all. Confirmed on the annotate side too -- `SELECT ..., 1 AS "_strawberry_alias_a", 1 AS "_strawberry_alias_b"` for two aliases with the same arguments. Gating the split on the coerced arguments actually differing would fix both the ValueError and the duplicate work.

strawberry_django/optimizer.py
```diff
@@ -797,6 +922,31 @@ def _get_hints_from_field(
...
+    if hint_key and field_store.annotate:
+        # The field is selected multiple times via aliases with its callables
+        # resolved once per alias, so each callable annotation gets an
+        # alias-scoped label to avoid clashing with the other aliases:
+        # the single-expression form (labeled with the field's name) becomes
+        # the hint key itself, while custom dict labels get prefixed by it.
+        # Resolvers can read the values back with `get_hint_value`.
+        # Static expressions can't depend on the alias arguments, so they keep
+        # their shared label and are annotated only once.
+        def _scoped_label(label: str, value: AnnotateType) -> str:
```
bellini666, line 1011: Claude raised this concern to me:

**Blocker.** Dict-form callable `annotate` hints get renamed out from under existing resolvers. When the field is selected twice the label becomes `_strawberry_alias_<key>__<label>`, so a resolver reading the fixed attribute breaks.

Reproduced with this PR's own `MilestoneType.my_bugs_count` (`return root._my_bugs_count`):

```
{ milestoneList { id a: myBugsCount b: myBugsCount } }
main: {'a': 2, 'b': 2}
PR:   AttributeError: 'Milestone' object has no attribute '_my_bugs_count'
```

This is the pattern `docs/guide/optimizer.md` currently teaches, and it breaks the moment a client aliases the field -- silently, from the user's point of view, since nothing in their code changed. `tests/projects/schema.py::my_bugs_count` isn't migrated to `get_hint_value` and no test aliases it.

strawberry_django/optimizer.py
```diff
@@ -1056,7 +1207,15 @@ def _get_hints_from_django_relation(
     if is_inheritance_qs(base_qs):
         base_qs = base_qs.select_subclasses(*subclasses)
     field_qs = field_store.apply(base_qs, info=field_info, config=config)
-    field_prefetch = Prefetch(path, queryset=field_qs)
+    # Don't use to_attr for connection/paginated fields - their resolvers
+    # expect data in _prefetched_objects_cache, not as a plain list attribute.
+    # If to_attr was requested but can't be used, skip optimization entirely
+    # to avoid merging conflicting prefetches for the same path.
+    if to_attr and (
```
bellini666: Claude raised this concern to me:

The guard itself is right, but it returns too late. `cache.setdefault(remote_model, []).append((level, field_store))` (line ~1195), `_get_prefetch_queryset` and `field_store.apply(base_qs, ...)` have all already run by the time we get here. Every other early return in this function bails before the cache append.

Two consequences: the orphaned cache entry is drained by the `level == 0` merge loop and widens the parent's `only()`/`select_related` for a prefetch that was never emitted, and `_get_prefetch_queryset` invokes the field's user `get_queryset` hook -- running it for a result that gets discarded is a side-effect hazard, not just wasted CPU. Moving the check above the recursion at ~line 1146 would avoid building the whole nested subtree first.

Separately: for a connection/paginated field carrying a callable `annotate`, the split at 1521 forces a `to_attr` even for identical aliases, which lands here and emits no prefetch at all -- 0 where main emitted 1, so `2 x N_parents` queries. Code-traced only; the test schema has no connection field with a callable annotate, so I couldn't reproduce it.

One more the guard doesn't cover: it excludes `is_connection`/`is_paginated` but not "not `is_list`". A non-list field over a reverse FK (`Optional[IssueT] = field(field_name="issues", filters=...)`) aliased with different filters would return the raw list, bypassing `get_queryset_hook`'s `.first()` -- schema promises an object, resolver returns a list. Reasoned from the code path, no running repro, so treat that last one as unverified.

bellini666 (commented):
Claude raised this concern to me:

I ran a four-pass self-review over this PR (correctness, regressions, performance, over-engineering) and left the findings inline. Three of them are blockers and all three are reproduced against this branch with a diff against the merge base -- the prefetch `to_attr` name breaking on aliases containing `__` or starting with `_`, callable `prefetch_related` hints raising `ValueError` on two response keys, and dict-form annotate labels breaking the documented `root._label` resolver pattern.

Two of those three collapse into one change: make the `_field_has_callable_hints` split conditional on the coerced arguments actually differing, and disambiguate callable `prefetch_related` the way `annotate` already is.

Three notes that couldn't be anchored inline, since the lines fall just outside the diff hunks:

**`_generate_selection_resolve_info` (~line 842)** -- `info.path` inside hint callables now reports the alias rather than the field name:

```
{ milestoneList { a: graphqlPath b: graphqlPath } }
main: {'a': 'milestoneList,0,graphqlPath', 'b': 'milestoneList,0,graphqlPath'}
PR:   {'a': 'milestoneList,0,a',           'b': 'milestoneList,0,b'}
```

That's the intent of the change, but it's an undocumented break for anyone keying cache or permission logic on `info.path` inside a hint callable or a `get_queryset` reached through `_optimize_prefetch_queryset`. Worth a line in RELEASE.md. I grepped the other `info.path` consumers -- `optimizer.py:1443` walks `path.typename` only, and strawberry core's consumers only ever see the real resolve info -- so nothing in-tree breaks.

**`_get_model_hints` (~lines 1591 and 1602)** -- `to_attr` is threaded as `hint_key` into `_get_hints_from_field` (an annotation label) and as `to_attr` into `_get_hints_from_django_field` (a prefetch attribute). Same string, two subsystems. For a relation field with a single-expression callable annotate, `_scoped_label` returns exactly `hint_key`, so the annotation alias and the `Prefetch(to_attr=...)` would be the same name on the same instance and one clobbers the other. It's unreachable today -- that configuration already fails on the merge base for an unrelated reason -- so latent rather than live, but the two uses probably want distinct names.

Things I checked that came back clean: no `assert_num_queries` expectation was raised anywhere (the test diff is purely additive), no model changes so no migration needed, the snapshot changes are additive test-schema fields only, and aliased connections with equal arguments actually improve from 9 queries to 2. The existing suite passes on this branch (1242 passed).

One structural note in the PR's favour: the argument-coercion change in `_get_field_arguments` is a real fix, not a refactor. `graphql.Node.__eq__` includes `loc`, so the old AST-tuple comparison returned `False` for any two selections carrying arguments, whatever their values -- aliased selections with arguments were never merged on main.

strawberry_django/optimizer.py
```diff
@@ -713,8 +720,102 @@ def _get_selections(
     return get_sub_field_selections(info, parent_type)
 
 
-def _get_field_arguments(node: FieldNode) -> tuple:
-    return tuple(sorted(node.arguments or (), key=lambda a: a.name.value))
+def _get_field_arguments(
```
bellini666, line 745: Claude raised this concern to me:

Both new paths look dead. `get_argument_values` doesn't raise after validation has run, and meta fields never reach a store -- `_get_field_data` returns `None` for them. The fallback also returns a `tuple` where the success path returns a `dict`, so the two can never compare equal; that lands on "different args", which is at least the conservative direction, but it means the suppress can only ever silently change behaviour, never recover.

Worth noting the underlying fix here is real and load-bearing: `graphql.Node.__eq__` includes `loc`, so the old `tuple(sorted(node.arguments))` compared equal only when both nodes had zero arguments. Aliased selections carrying arguments were never merged on main. The coercion is a genuine fix -- it's just the two extra paths around it that don't earn their place.

strawberry_django/optimizer.py
```diff
@@ -713,8 +720,102 @@ def _get_selections(
...
+    Derived from the response key - the alias if the field is aliased,
+    otherwise the field's own name. Response keys are guaranteed to be unique
+    within a selection set, and this returns the same value during optimizer
+    hint collection and at resolve time.
+
+    Use it to name an annotation or a `Prefetch(to_attr=...)` inside an
+    optimizer hint callable, and `get_hint_value` to read the value back
+    inside the resolver.
+    """
+    return f"{ALIAS_PREFIX}{info.path.key}"
```
bellini666: Claude raised this concern to me:

`ALIAS_PREFIX` is concatenated in three places with three different derivations (here, 1530, 1547), two of which duplicate this function. They also disagree: 1530 falls back to the field name for an unaliased member, 1547 falls back to `None`. `_generate_selection_resolve_info` already computes the response key -- deriving it once there and passing it down would remove both f-strings and the disagreement.

strawberry_django/optimizer.py
```diff
@@ -713,8 +720,102 @@ def _get_selections(
...
+    hint_key = optimizer_hint_key(info)
+
+    if default_attr is not None:
+        value = getattr(source, f"{hint_key}{LOOKUP_SEP}{default_attr}", _sentinel)
+        if value is not _sentinel:
+            return value
+
+    value = getattr(source, hint_key, _sentinel)
+    if value is not _sentinel:
+        return value
```
bellini666, line 805: Claude raised this concern to me:

Step 3 of the ladder (bare `default_attr`) has no test coverage. Every test query selects the field at least twice, so `hint_key` is always non-None and this branch never runs with the optimizer on. The step is legitimate -- it's what preserves the existing `root._my_bugs_count` pattern -- but it's exactly the path that Blocker 3 above breaks, so it's worth a single-selection test.

strawberry_django/optimizer.py
```diff
@@ -713,8 +720,102 @@ def _get_selections(
...
+    if default is not _sentinel:
+        return default
+
+    raise AttributeError(
+        f"No optimizer hint value found for {info.path.key!r} on {source!r}. "
+        "Was the query executed without DjangoOptimizerExtension?"
+    )
+
+
+def get_field_arguments(info: Info) -> dict[str, Any]:
```
bellini666, line 821: Claude raised this concern to me:

This looks like it duplicates something strawberry already exposes: `info.selected_fields[0].arguments` (via `convert_arguments` in `strawberry/types/nodes.py`) returns the same GraphQL-named dict with variables resolved, and works in both phases. The only delta is default-value coercion.

Also, half the advertised use case is dead -- inside a resolver the arguments already arrive as Python kwargs, so this only ever matters inside the hint callable. Given it's going into the package's public `__all__` permanently, worth checking whether documenting the one-liner would do instead.

strawberry_django/optimizer.py
```diff
@@ -740,6 +845,27 @@ def _generate_selection_resolve_info(
...
+    schema: Schema,
+    field_name: str,
+) -> StrawberryField | None:
+    for field in object_definition.fields:
+        if schema.config.name_converter.get_graphql_name(field) == field_name:
+            return field
+    return None
+
+
+def _field_has_callable_hints(field: StrawberryField) -> bool:
```
bellini666: Claude raised this concern to me:

One call site (line 1521). Per the project's style this reads better inlined -- it doesn't recurse and the caller isn't long enough to need the split. (`_find_strawberry_field` above has two call sites, so that extraction earns itself.)

strawberry_django/optimizer.py
```diff
@@ -797,6 +922,31 @@ def _get_hints_from_field(
...
+        # their shared label and are annotated only once.
+        def _scoped_label(label: str, value: AnnotateType) -> str:
+            if not callable(value):
+                return label
+            if label == field.name:
+                return hint_key
+            return f"{hint_key}{LOOKUP_SEP}{label}"
+
+        annotate = {_scoped_label(k, v): v for k, v in field_store.annotate.items()}
+        if annotate.keys() != field_store.annotate.keys():
```
bellini666: Claude raised this concern to me:

This reimplements `OptimizerStore.copy()` -- `dataclasses.replace(field_store, annotate=annotate)` does the same thing and won't silently drop a field if the dataclass gains one.

The `if annotate.keys() != field_store.annotate.keys()` guard on the line above is dead: both downstream branches (`with_prefix` / `with_resolved_callables`) return a fresh store regardless, so the micro-optimization buys nothing.

strawberry_django/optimizer.py
```diff
@@ -1347,19 +1508,48 @@ def _get_model_hints(
         name = field_nodes[0].name.value
         field_name_groups.setdefault(name, []).append(field_nodes)
 
-    # Merge aliased selections with same arguments; skip those with different args
-    merged_node_lists: list[list[FieldNode]] = []
-    for groups in field_name_groups.values():
+    # Merge aliased selections with the same arguments; keep aliases with
+    # different args as separate entries by assigning distinct to_attr values.
```
bellini666: Claude raised this concern to me:

These comments narrate the loop rather than explain it -- the `# list of (field_nodes, to_attr)` line restates the type annotation directly below it, and the header restates the body. Same for the two inside the branches at 1534 and 1541.

The block at 1522 is worth keeping down to its one non-obvious clause ("even when the arguments are identical"); the rest walks the code line by line. Ditto the block above `_scoped_label` -- the last sentence (static expressions can't depend on alias arguments, so they keep their shared label) is the real invariant, the rest is narration.

The two I'd definitely keep: the path-key note in `_generate_selection_resolve_info` and the `_prefetched_objects_cache` constraint at ~1210. Both record something the code genuinely can't say for itself.

strawberry_django/optimizer.py
```diff
@@ -1347,19 +1508,48 @@ def _get_model_hints(
...
+            continue
+
+        first_args = _get_field_arguments(groups[0][0], parent_type, info)
+        if all(
+            _get_field_arguments(g[0], parent_type, info) == first_args
+            for g in groups[1:]
+        ):
+            # Same args across all aliases - merge into single entry, no to_attr
+            merged_node_lists.append((
+                [node for group in groups for node in group],
```
bellini666: Claude raised this concern to me:

Minor: this compares every group against `groups[0]`, so if any one group differs, all groups get separate `to_attr` prefetches -- including pairs whose arguments do match. Correct data, extra queries. Grouping by coerced args rather than testing all-equal-to-first would handle the three-way case.

strawberry_django/fields/field.py
```diff
@@ -232,6 +236,13 @@ def get_result(
             # sync_to_async context if the value is already cached, since it will not
             # hit the db anymore
             attname = self.django_name or self.python_name
+
+            # Check for to_attr-based prefetch from optimizer (aliased field with filters)
```
bellini666: Claude raised this concern to me:

Three things here.

This is `get_hint_value(source, info, default=...)` open-coded -- calling the helper would keep the probe order in one place.

It sits in the `else` branch of `get_result`, so it runs for every non-resolver field on every object, scalars included, and it runs even with the optimizer disabled. I measured ~76ns per call (f-string plus failed `hasattr`), so ~0.8ms per 10k field resolutions -- negligible against GraphQL execution, and it can't trigger a deferred-field fetch since `_strawberry_alias_*` is never a `DeferredAttribute`. The one case that isn't free is a model defining `__getattr__`, where every miss now goes through user code. Gating on `is_list` would sidestep all of it.

And the comment restates the three lines under it.

tests/test_optimizer.py
```diff
@@ -1268,6 +1602,403 @@ def test_query_prefetch_with_aliases_different_subselections(
...
+                        },
+                    ],
+                },
+            ],
+        },
+    }
+
+
+@pytest.mark.django_db(transaction=True)
+def test_query_prefetch_aliases_with_different_pagination(
```
bellini666, line 2413: Claude raised this concern to me:

This one can't fail. It asserts `assert_num_queries(3)` against a single project, so prefetched and lazy resolution produce the same count -- I ran it against main and got 3 queries and identical data. Using 2+ projects like the sibling tests would make it a real regression test (main would be 5, this branch 3).

tests/test_optimizer.py
```diff
@@ -1268,6 +1602,403 @@ def test_query_prefetch_with_aliases_different_subselections(
     }
 
 
+@pytest.mark.django_db(transaction=True)
+def test_query_prefetch_aliases_with_nested_selections(
```
bellini666, line 2328: Claude raised this concern to me:

Both aliases here are argument-free, so main already merged them -- I verified it passes unchanged on main (3 queries, same data). Fine as a "still works" test, just noting it isn't guarding the new behaviour.

Most of the others in this batch do fail when the production change is reverted -- `different_filters`, `alias_and_field_with_different_filters`, `same_filters`, `with_ordering`, `filters_and_ordering`, `mixed_filtered_and_unfiltered`, the callable hint tests and `test_get_hint_value_raises_without_hint_or_default` all genuinely guard it.

tests/test_optimizer.py
```diff
@@ -1268,6 +1602,403 @@ def test_query_prefetch_with_aliases_different_subselections(
...
+                "id": to_base64("MilestoneType", milestone_2.pk),
+                "foo": 0,
+                "bar": 1,
+            },
+        ],
+    }
+
+
+@pytest.mark.django_db(transaction=True)
+def test_query_aliased_annotate_callable_with_same_arguments(
```
bellini666, line 2573: Claude raised this concern to me:

Expects 1 query and gets it, but the branch emits two byte-identical `Count(...)` aggregates because `_field_has_callable_hints` splits identical-arg aliases. Since the assertion is on the count and not the SQL, it won't notice if that gets worse. Same root cause as the `prefetch_related` ValueError.

RELEASE.md
```diff
@@ -0,0 +1,37 @@
+---
+release type: minor
+---
+
```
bellini666, line 4: Claude raised this concern to me:

This paragraph describes #882, which is already on `main` (`_get_field_arguments` exists there and the alias-merging behaviour shipped with it) -- worth checking before this goes out, since autopub publishes the file as-is.

The rest is longer than this repo's release notes usually run: three-bullet API tour, a Python example and a GraphQL example, then a recap sentence. One paragraph naming the behaviour change plus a bare list of the three new exports would cover it -- both examples already live in the docs.

docs/guide/optimizer.md
```diff
@@ -287,6 +287,111 @@ The following options are accepted for optimizer hints:
...
+```graphql
+query {
+  orderItems {
+    cheap: totalFiltered(maxPrice: 10)
+    expensive: totalFiltered(maxPrice: 1000)
+  }
+}
+```
+
+Each alias resolves the hint callables with its own `info`, scoped to that
```
bellini666, line 305: Claude raised this concern to me:

The alias/hint-key contract is genuinely non-obvious and deserves this section. This bullet list restates the three docstrings close to word for word, though -- the docstrings are already good, and two copies will drift.

One thing the section doesn't say and should: each alias of a callable `prefetch_related` field costs a full extra query, even when the arguments are identical. That's a real cost someone will hit without warning.

docs/guide/optimizer.md
```diff
@@ -287,6 +287,111 @@ The following options are accepted for optimizer hints:
...
+        self, root: models.Order, info: Info, max_price: int
+    ) -> list[OrderItem]:
+        value = get_hint_value(root, info, default=None)
+        if value is None:
+            # The optimizer is turned off, resolve the value directly
+            value = list(root.items.filter(price__lte=max_price))
+        return value
+```
+
+A field can also annotate multiple values and combine them in the resolver.
```
bellini666, line 360: Claude raised this concern to me:

This third example reteaches what the two above it already showed, with a different aggregate. The annotate example, the prefetch example and the static-annotation note carry the section on their own.

bellini666 (changes requested):
Hey, sorry for taking long to review this

The PR is kinda big, so I asked claude to also take a look. It raised some concerns, so I asked it to post here

docs/guide/optimizer.md
rcybulski1122012: Not sure about it. 

I think a more complex example with usage of the `default_attr` argument  is actually helpful. WDYT?

strawberry_django/fields/field.py
rcybulski1122012: `in_list` on its own will not be sufficent, because it doesn't detect fields produced by `annotated`. I also check there `self.store` is not `None`.

strawberry_django/optimizer.py
rcybulski1122012: Not fully convinced about it, but fixed it either way.

For me it's a natural way of thinking about callables in prefetch_related/annotate, that they are executed per alias.

strawberry_django/optimizer.py
rcybulski1122012: I think it currently has to be that way. It's related to this comment https://github.com/strawberry-graphql/strawberry-django/pull/930#discussion_r3887047378

There is no option to provide a dynamic name based on the info object in the dict-form annotate. We would need to change the annotate so it also accepts `Callable[[Info], dict]` so we can use `info`. Looking back, I don’t really like it either that we change the name of the annotate attribute without user's knowledge :/

Ofc we can detect when the arguments are the same and merge it into a single annotate without changed the name -- it's been already fixed :)

rcybulski1122012 (commented):
I hope i fixed everything. Few things require a discussion, so let me know what you think.

Also, I used AI quite extensively to understand and implement this, but I still don't have a full understanding of how the optimizer works under the hood.

rcybulski1122012:
@bellini666

PR #931, opened 2026-07-04
fix: skip the per-parent COUNT for totalCount of empty prefetched nested connections

rcybulski1122012:
When a nested connection is optimized by prefetching, totalCount is read from the _strawberry_total_count window annotation on the fetched rows. Parents with no matching rows have nothing to read the annotation from, so get_total_count fell back to one real COUNT query per such parent — an N+1 that grows with the number of childless parents.

For a first-page window (no row_number offset filter and a limit that admits at least one row), an empty partition can only mean the filtered set is empty, so totalCount is 0 without querying. Pages past the end (offset > 0) and first: 0 pages still fall back to COUNT, since there the empty page proves nothing.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

## Summary by Sourcery

Avoid unnecessary COUNT queries when resolving totalCount for empty prefetched nested connections.

Bug Fixes:
- Ensure totalCount for nested connections with empty prefetched partitions is correctly resolved as 0 without issuing per-parent COUNT queries.

Enhancements:
- Add logic to detect non-empty first-page window pagination and infer zero totalCount from empty result sets to prevent N+1 COUNT queries in nested connections.

Documentation:
- Add RELEASE notes describing the patch-level fix for nested connection totalCount N+1 issues.

Tests:
- Add regression tests covering totalCount behavior for nested connections with empty partitions, offsets past the end, and first: 0 pages in both Django optimizer and relay cursor pagination flows.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+Release type: patch
+
+Fix an N+1 on `totalCount` of nested connections optimized by prefetching:
+parents whose prefetched first-page partition came back empty issued one
+`COUNT(*)` query each, even though an empty first page already proves the
+total count is 0.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,6 @@
-Release type: patch
+---
+release type: patch
+---
 
 Fix an N+1 on `totalCount` of nested connections optimized by prefetching:
 parents whose prefetched first-page partition came back empty issued one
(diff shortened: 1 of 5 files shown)


botberry:

---

Fix an N+1 on `totalCount` of nested connections optimized by prefetching:
parents whose prefetched first-page partition came back empty issued one
`COUNT(*)` query each, even though an empty first page already proves the
total count is 0.

This release was contributed by @rcybulski1122012 in https://github.com/strawberry-graphql/strawberry-django/pull/931

bellini666 (approved):
Thanks! :)

botberry:
This PR was published as [0.86.6](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.86.6). Thank you for contributing!

PR #932, opened 2026-07-04
perf: resolve prefetch-optimized nested connections without per-node thread hops

rcybulski1122012:
Every sync resolver runs through sync_to_async in async contexts, costing a thread hop (~0.1-0.2 ms) per call. For nested connections the optimizer has already fetched everything with one windowed prefetch query, yet each parent node still paid three hops for pure in-memory work: the connection default resolver (builds the related queryset without evaluating it), the queryset hook (short-circuits for prefetch-optimized querysets) and totalCount (reads the _strawberry_total_count window annotation off a cached row).

- The connection default resolver is marked `is_async_safe` and skipped by the django_resolver wrapping in `safe_resolver`; its only potentially database-touching path (resolve_model_nodes, which may run user-defined get_queryset) is explicitly deferred to a thread in async contexts.
- `get_result` applies the queryset hook inline when the queryset is prefetch-optimized and its results are already cached.
- `total_count` reads the cached window annotation inline via the new `get_cached_total_count` helper and defers to a thread only when a real COUNT query is unavoidable.

Nested cursor connection resolution over prefetched data drops from 9 to 3 thread hops (only the root connection touches the database); nested list connections drop from 7 to 1.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

## Summary by Sourcery

Optimize resolution of prefetch-optimized nested connections to avoid unnecessary sync_to_async thread hops while preserving correct pagination counts.

Enhancements:
- Use cached window annotations via a new helper to serve totalCount for prefetch-optimized querysets without hitting the database.
- Treat library-internal connection default resolvers and queryset hooks as async-safe when operating purely on prefetched data, skipping sync_to_async wrapping in async contexts.
- Refine safe_resolver behavior to respect async-safe resolvers so nested connection resolution over prefetched data runs on the event loop.

Documentation:
- Add a release note describing the optimization of nested connections to resolve prefetched data without per-node thread hops.

Tests:
- Add coverage for get_cached_total_count and its interaction with distinct querysets and missing pagination annotations.
- Add async tests that track SyncToAsync call sites to enforce constant thread hops for prefetched nested connections and per-parent hops when data is not prefetched.

Chores:
- Introduce RELEASE.md with patch-level release metadata for this optimization.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,7 @@
+Release type: patch
+
+Resolve prefetch-optimized nested connections on the event loop instead of
+paying a `sync_to_async` thread hop per parent node: the nested-connection
+default resolver, the queryset hook for prefetch-optimized querysets and
+`totalCount` served from the prefetched window annotation no longer touch
+the database, so they no longer need a worker thread.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,6 @@
-Release type: patch
+---
+release type: patch
+---
 
 Resolve prefetch-optimized nested connections on the event loop instead of
 paying a `sync_to_async` thread hop per parent node: the nested-connection
--- a/strawberry_django/fields/base.py
+++ b/strawberry_django/fields/base.py
@@ -171,7 +171,11 @@ class StrawberryDjangoFieldBase(StrawberryField):
         resolver = self.base_resolver
         assert resolver
 
-        if not resolver.is_async:
+        # Library-internal resolvers marked `is_async_safe` never run a query
+        # themselves, so they can skip the sync_to_async thread hop entirely.
+        if not resolver.is_async and not getattr(
+            resolver.wrapped_func, "is_async_safe", False
+        ):
             resolver = django_resolver(resolver, qs_hook=None)
 
         return resolver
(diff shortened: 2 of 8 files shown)


botberry:

---

Resolve prefetch-optimized nested connections on the event loop instead of
paying a `sync_to_async` thread hop per parent node: the nested-connection
default resolver, the queryset hook for prefetch-optimized querysets and
`totalCount` served from the prefetched window annotation no longer touch
the database, so they no longer need a worker thread.

This release was contributed by @rcybulski1122012 in https://github.com/strawberry-graphql/strawberry-django/pull/932

RELEASE.md
```diff
@@ -0,0 +1,9 @@
+---
+release type: patch
```
bellini666: I would do this as a minor

```suggestion
release type: minor
```

bellini666 (commented):
Sorry for taking long to check this.

Left one comment, rest looks good! Can you address it and also fix the conflict we got when merging your other PR?

bellini666 (approved):

botberry:
This PR was published as [0.87.0](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.87.0). Thank you for contributing!

PR #945, opened 2026-08-20
fix: LOOKUP_SEP boundary when recovering select_related

### LINKED ISSUE #938: Optimizer: select_related preserved from the queryset is dropped from only() by a prefix match without `__` boundary
## Summary

`OptimizerStore._apply_select_related` tries to keep manually-applied `select_related` in the `only()` mask, but the coverage check is a raw string prefix match. When two relations on the same model share a name prefix (`empresa` / `empresa_comercial`), selecting only the longer one makes the optimizer believe the shorter one is covered. It stays `select_related` while being deferred, and Django rejects the whole query:

```
Field Negocio.empresa cannot be both deferred and traversed using select_related at the same time.
```

Version: **strawberry-graphql-django 0.86.5** (current latest), strawberry-graphql 0.322.0, Django 4.2.

## Offending code

`strawberry_django/optimizer.py`, in `_apply_select_related` (0.86.5, around line 468):

```python
for select_related in select_related_set:
    if select_related in only_set:
        continue

    if not any(only.startswith(select_related) for only in only_set):
        extra_only_set.add(select_related)
```

`only.startswith(select_related)` has no `LOOKUP_SEP` boundary, so `"empresa_comercial__id".startswith("empresa")` is `True` and `empresa` is never added to `extra_only_set`.

## Reproduction

```python
class Negocio(models.Model):
    empresa = models.ForeignKey(Empresa, null=True, on_delete=models.CASCADE)
    empresa_comercial = models.ForeignKey(EmpresaComercial, null=True, on_delete=models.CASCADE)
    nome = models.CharField(max_length=200)


@strawberry_django.type(Negocio)
class NegocioType:
    nome: auto
    empresa: Optional[EmpresaType]
    empresa_comercial: Optional[EmpresaComercialType]


@strawberry.type
class Query:
    @strawberry_django.field
    def negocios(self) -> list[NegocioType]:
        # any manual select_related on the shorter-named relation
        return Negocio.objects.select_related("empresa")


schema = strawberry.Schema(query=Query, extensions=[DjangoOptimizerExtension()])
```

Query that fails — it selects `empresaComercial` and **not** `empresa`:

```graphql
{ negocios { nome empresaComercial { id } } }
```

→ `Field Negocio.empresa cannot be both deferred and traversed using select_related at the same time.`

Adding `empresa { id }` to the selection, or removing `empresaComercial`, makes it pass.

The same happens when the colliding sibling is a plain local field rather than a relation: with `select_related("etapa")` and a concrete field named `etapa_atualizada_em`, selecting `etapaAtualizadaEm` alone triggers the identical error for `Negocio.etapa`.

## Expected

A `select_related` path should only count as covered by an `only` entry when the entry is the path itself or a descendant of it — i.e. the match should respect the `__` boundary:

```python
if not any(
    only == select_related or only.startswith(f"{select_related}{LOOKUP_SEP}")
    for only in only_set
):
    extra_only_set.add(select_related)
```

`LOOKUP_SEP` is already imported in that module.

## Impact

This is silent until a query happens to select the longer sibling without the shorter one, and then it fails the *entire* field with `data: null` rather than degrading. It hit us in production on a Kanban board that had been working for weeks — the only change was a new `empresaComercial { pk }` in the client selection set.

Happy to send a PR with the boundary fix plus a regression test if that's welcome.

cjcontreras:
Fixes #938.

<div>
    <a href="https://www.loom.com/share/d689f6e1ed404d5ea102c15b837c9d26">
      <p>strawberry-django PR 945 fix: LOOKUP_SEP boundary when recovering select_related - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/d689f6e1ed404d5ea102c15b837c9d26">
      
    </a>
  </div>

`_apply_select_related` checked whether a manual `select_related` path was covered by an `only()` entry via a raw `only.startswith(select_related)`. Without a `__` boundary, a sibling field sharing the same prefix (e.g. `empresa` vs `empresa_comercial`) was wrongly treated as covering it, so the shorter relation never got re-added to `only()`. Django then raised "cannot be both deferred and traversed using select_related at the same time".

Factored the check into a shared `_is_relation_path_covered` helper, reused at two other call sites with the same shape (one already correct, one with the identical bug).

Correction: the third call site touched by this refactor (`_get_hints_from_django_field`) is unreachable as a live bug. `store.select_related` is always empty at that check, so the guard is vacuous both before and after this change. That change is a no-op refactor for consistency, not a bug fix.

### Tests
- Unit tests pin the boundary semantics directly against `OptimizerStore._apply_select_related` (exact match, descendant, sibling-prefix).
- New end-to-end GraphQL tests cover the bug against real queries: the field-hint `select_related` route, the queryset `select_related` route, a nested-depth collision, a negative control, the `_store_declares_select_related` call site, and a regression guard for FK attname vs relation name in `only()`.
- Full suite: 1247 passed, 6 skipped. `ruff` lint + format clean.

## Summary by Sourcery

Preserve manually selected related fields when recovering `only()` masks by requiring exact or lookup-separator-bounded path matches.

Bug Fixes:
- Prevent relation paths that merely share a string prefix from being treated as covered by `only()`, avoiding invalid deferred and `select_related` combinations.

Enhancements:
- Centralize relation-path coverage checks with lookup-separator boundary semantics and reuse them across optimizer call sites.

Tests:
- Add regression coverage for exact, descendant, and sibling-prefix relation paths.

## Summary by Sourcery

Fix relation-path coverage in the query optimizer so only lookup-separator-bounded matches satisfy manually applied `select_related` paths.

Bug Fixes:
- Preserve manually selected related fields in recovered `only()` masks when field names share a string prefix without a lookup-separator boundary, preventing invalid deferred and `select_related` combinations.

Enhancements:
- Centralize relation-path coverage checks and apply consistent exact-or-lookup-bounded descendant semantics across optimizer call sites.

Tests:
- Add unit and end-to-end regression coverage for exact matches, bounded descendants, sibling-prefix collisions, nested paths, queryset and field-hint routes, and foreign-key attnames.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,7 +4,7 @@ release type: patch
 
 Fix the query optimizer dropping a manually applied `select_related` from the
 `only()` mask when another field on the same model shares its name as a
-prefix without a `__` boundary (e.g. `empresa` vs `empresa_comercial`). This
+prefix without a `__` boundary (e.g. `company` vs `company_branch`). This
 previously made Django raise `FieldError: cannot be both deferred and
 traversed using select_related at the same time` whenever only the longer
 sibling was selected in the query.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,10 @@
+---
+release type: patch
+---
+
+Fix the query optimizer dropping a manually applied `select_related` from the
+`only()` mask when another field on the same model shares its name as a
+prefix without a `__` boundary (e.g. `empresa` vs `empresa_comercial`). This
+previously made Django raise `FieldError: cannot be both deferred and
+traversed using select_related at the same time` whenever only the longer
+sibling was selected in the query.
--- a/strawberry_django/optimizer.py
+++ b/strawberry_django/optimizer.py
@@ -469,7 +469,9 @@ class OptimizerStore:
                 if select_related in only_set:
                     continue
 
-                if not any(only.startswith(select_related) for only in only_set):
+                if not any(
+                    _is_relation_path_covered(only, select_related) for only in only_set
+                ):
                     extra_only_set.add(select_related)
 
         return qs, extra_only_set
@@ -509,6 +511,15 @@ class OptimizerStore:
             to_annotate[k] = v
 
         return qs.annotate(**to_annotate)
+
+
+def _is_relation_path_covered(candidate: str, target: str) -> bool:
+    """Whether `candidate` is `target` itself or a `LOOKUP_SEP`-bounded descendant of it.
+
+    A plain `candidate.startswith(target)` is wrong: e.g. "empresa_comercial"
+    starts with "empresa" even though they're unrelated sibling fields.
+    """
+    return candidate == target or candidate.startswith(f"{target}{LOOKUP_SEP}")
 
 
 def _create_strawberry_info(raw_info: GraphQLResolveInfo) -> Info:
@@ -1078,11 +1089,10 @@ def _store_declares_select_related(
     if not field_store:
         return False
 
-    for sr in field_store.select_related:
-        if sr == relation_name or sr.startswith(f"{relation_name}{LOOKUP_SEP}"):
-            return True
-
-    return False
+    return any(
+        _is_relation_path_covered(sr, relation_name)
+        for sr in field_store.select_related
+    )
 
 
 def _get_hints_from_django_field(
@@ -1223,7 +1233,10 @@ def _get_hints_from_django_field(
         store = OptimizerStore.with_hints(only=[path])
         if relation_prefix:
             relation_path = f"{lookup_prefix}{relation_prefix}"
-            if not any(sr.startswith(relation_path) for sr in store.select_related):
+            if not any(
+                _is_relation_path_covered(sr, relation_path)
+                for sr in store.select_related
+            ):
                 store.select_related.append(relation_path)
 
     return store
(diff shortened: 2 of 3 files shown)


botberry:

---

Fix the query optimizer dropping a manually applied `select_related` from the
`only()` mask when another field on the same model shares its name as a
prefix without a `__` boundary (e.g. `company` vs `company_branch`). This
previously made Django raise `FieldError: cannot be both deferred and
traversed using select_related at the same time` whenever only the longer
sibling was selected in the query.

This release was contributed by @cjcontreras in https://github.com/strawberry-graphql/strawberry-django/pull/945

RELEASE.md
```diff
@@ -0,0 +1,10 @@
+---
+release type: patch
+---
+
+Fix the query optimizer dropping a manually applied `select_related` from the
+`only()` mask when another field on the same model shares its name as a
+prefix without a `__` boundary (e.g. `empresa` vs `empresa_comercial`). This
```
bellini666: As a brazilian myself, I do understand what "empresa"/"empresa_comercial" means :)

But can we use an english example for non-(portuguese/spanish) speakers?

bellini666 (changes requested):

RELEASE.md
cjcontreras: Good deal! I changed it to just be english, let me know if you want both instead. Also fixed playwright issue

bellini666 (approved):

botberry:
This PR was published as [0.87.1](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.87.1). Thank you for contributing!

PR #955, opened 2026-09-07
Require Strawberry 0.322.2 and update GraphQL compatibility

patrick91:
Require Strawberry GraphQL 0.322.2 or newer and update the Relay tests to assert its corrected `first` + `before` pagination behavior. Remove version-dependent expectations, refresh the lockfile to the new minimum, and register schema extensions as classes or factories to match Strawberry's supported API.

Keep project-schema comparisons independent of top-level definition ordering while retaining checks for fields, arguments, descriptions, directive applications, and duplicate definitions. The stored snapshot is unchanged.

Update the GraphQL-core 3.3 CI pin from 3.3.0a12 to 3.3.0rc0, which provides the enum coercion API required by newer Strawberry. Adapt the query optimizer to preserve resolve-info context and pass fragment and variable values in the formats expected by GraphQL-core 3.3. Cover variables inside named fragments, including query counts, and update two assertions for the changed required-argument error wording.

Include a minor release note for the new Strawberry requirement and GraphQL-core compatibility changes. This addresses the stale expectations reported by Strawberry's [downstream compatibility job](https://github.com/strawberry-graphql/strawberry/actions/runs/34154695036/job/101843982903?pr=4618).

Validation on Python 3.14.3 / Django 6.0:
- Strawberry 0.322.2 with GraphQL-core 3.2.7 and 3.3.0rc0: full suite passed on each (1,243 passed, 11 skipped), then the expanded fragment regression passed all 8 cases on each.
- Strawberry 0.327.7 with GraphQL-core 3.3.0rc0: 1,247 passed, 11 skipped, including the expanded regression.
- Full repository Pyright with locked dependencies: 0 errors and 0 warnings.
- Formatting and whitespace checks pass. Lint passes for changed code; the existing ERA001 examples in optimizer.py and RUF076 fixture in tests/conftest.py remain unchanged.

### DIFF
--- a/tests/projects/test_schema.py
+++ b/tests/projects/test_schema.py
@@ -15,9 +15,8 @@ SNAPSHOTS_DIR = pathlib.Path(__file__).parent / "snapshots"
 
 
 def test_schema():
-    # Directive argument types moved into the sorted type map in Strawberry
-    # 0.326.0. Compare definitions without depending on their printed order;
-    # fields, arguments, descriptions and directive applications still match.
+    # Ignore top-level definition order while preserving checks for fields,
+    # arguments, descriptions, and directive applications.
     actual = parse(normalize_sdl(str(schema)), no_location=True)
     expected = parse(
         normalize_sdl((SNAPSHOTS_DIR / "schema.gql").read_text()), no_location=True
--- a/tests/projects/test_schema.py
+++ b/tests/projects/test_schema.py
@@ -1,6 +1,7 @@
 import pathlib
 
 import strawberry
+from graphql import parse
 from pytest_snapshot.plugin import Snapshot
 
 import strawberry_django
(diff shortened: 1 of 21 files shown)


botberry:

---

Require Strawberry GraphQL 0.322.2 or newer, which fixes Relay pagination when combining `first` with `before`. Update compatibility tests for the corrected pagination behavior and make schema comparisons independent of top-level definition order.

Update GraphQL-core 3.3 compatibility for the 3.3.0rc0 execution API, preserving resolve-info context and adapting fragment and variable values used by the query optimizer.

This release was contributed by @patrick91 in https://github.com/strawberry-graphql/strawberry-django/pull/955

Additional contributors: @pre-commit-ci[bot]

tests/relay/test_fields.py
```diff
@@ -1,11 +1,18 @@
...
 from pytest_django import DjangoAssertNumQueries
 from strawberry.relay.utils import to_base64
 
 from .schema import FruitModel, schema
 
 pytestmark = pytest.mark.usefixtures("_fixtures")
 
+# Strawberry 0.322.2 corrected first+before to select from the start of the
+# bounded connection. Keep coverage for the older supported releases as well.
+FIRST_BEFORE_FIXED = Version(version("strawberry-graphql")) >= Version("0.322.2")
```
bellini666: Maybe we should just bump strawberry to require 0.322.2+? Don't see a reason to why we need to support the old version. Wdyt?

tests/projects/test_schema.py
```diff
@@ -13,9 +14,21 @@
 SNAPSHOTS_DIR = pathlib.Path(__file__).parent / "snapshots"
 
 
-def test_schema(snapshot: Snapshot):
-    snapshot.snapshot_dir = SNAPSHOTS_DIR
-    snapshot.assert_match(normalize_sdl(str(schema)), "schema.gql")
+def test_schema():
+    # Directive argument types moved into the sorted type map in Strawberry
+    # 0.326.0. Compare definitions without depending on their printed order;
+    # fields, arguments, descriptions and directive applications still match.
```
bellini666: This comment will look funny in the future 🤣
Maybe just mentioning why and not necessarely when strawberry itself changed

bellini666 (approved):

botberry:
This PR was published as [0.89.0](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.89.0). Thank you for contributing!

PR #956, opened 2026-09-12
feat(relay): add offset argument to relay connection (#339)

### LINKED ISSUE #339: Add offset argument to relay connection
It would be very useful to add an `offset` argument to relay connection. It's a feature that is implemented in graphene. See https://github.com/graphql-python/graphene-django/blob/v3.1.5/graphene_django/fields.py#L139-L145

## Feature Request Type

- [ ] Core functionality
- [x] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description
When using this library's `limit/offset` pagination, `pageInfo` doesn't update `hasPreviousPage` and `hasNextPage`, they are always false. Currently, I am using a custom connection as a workaround:
```python
import strawberry
from strawberry.relay.utils import from_base64, to_base64

import strawberry_django
from strawberry_django import relay

@strawberry.type
class CustomConnection(relay.ListConnectionWithTotalCount[strawberry.relay.NodeType]):
    """
    A strawberry connection to count the number of query results
    """

    # Adding offset argument to custom connection
    @classmethod
    def resolve_connection(
        cls,
        nodes: NodeIterableType[strawberry.relay.NodeType],
        *,
        info: Info,
        before: Optional[str] = None,
        after: Optional[str] = None,
        first: Optional[int] = None,
        last: Optional[int] = None,
        offset: Optional[int] = None,
        **kwargs: Any,
    ) -> AwaitableOrValue[Self]:

        # This implemntation is based on the graphene
        # implementation of first/offset pagination
        if offset:
            if after:
                offset += from_base64(after) + 1
            # input offset starts at 1 while the offset starts at 0
            after = to_base64("arrayconnection", offset - 1)

        conn = super().resolve_connection(
            nodes,
            info=info,
            before=before,
            after=after,
            first=first,
            last=last,
            **kwargs,
        )

        if inspect.isawaitable(conn):

            async def wrapper():
                resolved = await conn
                resolved.nodes = nodes
                return resolved

            return wrapper()

        conn = cast(Self, conn)
        conn.nodes = nodes
        return conn

```

But that means a custom resolver would have to look like this:
```python
    @strawberry.field
    def foos(
        self,
        info: strawberry.types.Info,
        before: Optional[str] = None,
        after: Optional[str] = None,
        first: Optional[int] = None,
        last: Optional[int] = None,
        offset: Optional[int] = None,
    ) -> Optional[CustomConnection[FooType]]:
        # Resolver logic here returning queryset qs
        return CustomConnection[FooType].resolve_connection(info=info, nodes=qs, offset=offset, first=first, last=last, after=after, before=before)
```

mini0n-ai:
### Summary
Add  argument to relay connections, enabling first/offset pagination with proper  ( and ) tracking.

Resolves #339
/claim #339

### Details
- Added  argument to  for relay connections.
- Implemented  helper to adjust/create an  cursor based on integer offset for cursor calculations while preserving Relay pagination semantics.
- Updated  and  to convert offset into cursor and pass through.
- Added offset support to  and  with proper  resolution when offset > 0.
- Updated optimizer prefetching window pagination to account for  parameter.
- Added comprehensive unit tests in  verifying first/offset pagination, cursor bounds, pageInfo state, and negative offset validation.

### Settlement
- Base L2 Sovereign Wallet:

## Summary by Sourcery

Add offset-based pagination to relay connections while preserving cursor behavior and accurate pagination metadata.

New Features:
- Add an optional offset argument to relay connections for first/after and last/before pagination.

Bug Fixes:
- Ensure offset-based pagination reports correct page boundaries and previous/next page state.

Enhancements:
- Support combining offsets with cursors while preserving Relay pagination semantics.
- Apply offset-aware pagination to optimized queryset prefetching.

Tests:
- Add coverage for zero, nonzero, and out-of-range offsets, cursor bounds, page information, and negative offset validation.


botberry:

---

Add `offset` argument to relay connections for first/after and last/before pagination.

This release was contributed by @mini0n-ai in https://github.com/strawberry-graphql/strawberry-django/pull/956

Additional contributors: @pre-commit-ci[bot]

bellini666:
Hi @mini0n-ai ,

I'm trying to understand, what the use case of `offset` is here? Because, in general, `first`/`last`/`before`/`after` should be enough to deal with cases where you would need an offset, no?

mini0n-ai:
Hi @bellini666,

That's a very fair question from a pure Relay specification perspective. 

The rationale comes directly from the context and user need described in #339:

1. Parity with Graphene-Django: In Graphene-Django (which many projects migrate from), offset is natively supported on Relay fields by mapping it directly to an after cursor (graphene_django/fields.py#L138-L145).
2. Page-Number / Jump-to-Page Navigation: Pure cursor pagination (first/after) is designed for sequential scrolling. In UI components like data tables with page pickers (e.g. "Jump to Page 5"), the client does not possess the opaque cursor for item 40 without sequentially fetching pages 1–4 first.
3. Relay Metadata with Offset: When users attempt to use standard limit/offset pagination instead of Relay, pageInfo metadata (hasPreviousPage, hasNextPage) is broken and always evaluates to false.

To respect Relay semantics, this implementation follows Graphene's proven pattern: offset is resolved internally into an equivalent after cursor via resolve_offset_to_after. This keeps edges, cursors, and queryset optimization intact while supporting offset-based navigation for clients that require it.

Br.
Tom

mini0n-ai:
P.S. Also rebased onto latest main (0.89.2) to resolve the merge conflict cleanly.

bellini666:
Even though graphene-django has it, it is not part of the spec, and offset can be easily resolved using `first`/`last`/`before`/`after` combinations.

Note that if you want to use this, you can easily subclass the implementation and add offset to handle it

Sorry about that 😔

